PR #226, opened 2022-08-08
Switch dependabot freq to monthly

smaye81:
Increases dependabot frequency to monthly

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -3,13 +3,13 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "monthly"
       timezone: UTC
       time: "07:00"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yaml
```diff
@@ -3,13 +3,13 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "monthly"
```
timostamm, line 6: Not sure what the default is, probably best to use an explicit `day: "monday"`.

smaye81: Monday is the default, yeah.  Changed to state it explicitly though.

timostamm (approved):
Since we don't have any other runtime dependencies besides protobuf-es, this will reduce the noise without any disadvantages. Thank you!

PR #242, opened 2022-08-26
Updates dependencies for protobuf-es packages

smaye81:
This makes a few updates to depencencies.  Namely:

- Moves `typescript` to the top-level package to stay consistent with `protobuf-es`.
- Pins the TypeScript version to `4.7.4` and all patches (moved from caret to tilde).  This is because there is a breaking change in `lib.d.ts` between 4.7.4 and the new 4.8.1 release:  [`ReadableStreamDefaultReadResult` has been renamed](https://github.com/microsoft/TypeScript/issues/50272).  We can update our code, but we should make the TypeScript update a separate PR.
- Update peer dependencies for all packages to use `protobuf-es@0.1.0`


paul-sachs (approved):

packages/connect-web-bench/README.md
```diff
@@ -10,5 +10,5 @@ it like a web server would usually do.
 
 | code generator | bundle size        | minified               | compressed           |
 |----------------|-------------------:|-----------------------:|---------------------:|
-| connect-web    | 87,178 b | 40,753 b | 11,266 b |
+| connect-web    | 150,308 b | 70,026 b | 12,450 b |
```
timostamm, line 13: This is a massive increase in bundle size, there is something wrong.

timostamm (approved):

PR #256, opened 2022-09-01
Update to TypeScript 4.8.2

smaye81:
Updates TypeScript to v4.8.2.  In addition, this accommodates the breaking change in the type rename (`ReadableStreamDefaultReadResult` was renamed to `ReadableStreamReadResult`).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,7 +17,7 @@
         "eslint": "^8.21.0",
         "eslint-plugin-node": "^11.1.0",
         "prettier": "^2.7.1",
-        "typescript": "~4.7.4"
+        "typescript": "~4.8.2"
       },
       "engines": {
         "node": ">=16",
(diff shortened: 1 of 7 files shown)


timostamm (commented):
> breaking change in the type rename (ReadableStreamDefaultReadResult was renamed to ReadableStreamReadResult).

Will that affect users of previous versions of TypeScript who do not skip lib checks?

smaye81:
Ugh yeah.  Actually users now who upgrade to TypeScript 4.8.2 with `skipLibCheck` false will be affected also.  Maybe this is a good time to use the `typesVersions` field in `package.json`?

timostamm:
> Maybe this is a good time to use the `typesVersions` field in `package.json`?

We could deliver two sets of typings, and use `typesVersions` to differentiate. But how are we going to produce the two sets? We could run the source code through two different versions of the TypeScript compiler to yield two sets, but in this specific case, we would need to change the usage of the type in our sources, because it isn't the compiler output that varies, but the bundled lib.dom.ts. This doesn't seem like a good way forward.

There is another potential alternative, see [here](https://devblogs.microsoft.com/typescript/announcing-typescript-4-5-beta/#supporting-lib-from-node_modules).

The most simple - and IMO most robust option - is to vendor the relevant types in a file lib.dom.streams.ts. I think we should be fine with a .ts file that we import (and not an ambient declaration file). I don't think we need to rename the types. But we definitely need a comment explaining the situation and linking to this PR.

Could you also add a link here to the relevant issues / PR in the TypeScript repository?

PR #257, opened 2022-09-01
Upgrade esbuild to 0.15.6

smaye81:
This upgrades esbuild to `0.15.6` and in doing so, fixes a BrowserStack test that was configured as a callback client.  This was because `esbuild` was unable to transpile down `for...await` loops.  

With the [0.15.6 release](https://esbuild.github.io/content-types/#javascript), that is now possible when the target is less than `es2018`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -88,9 +88,9 @@
       }
     },
     "node_modules/@esbuild/linux-loong64": {
-      "version": "0.14.54",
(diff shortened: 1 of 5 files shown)


timostamm (approved):
Pretty great timing that https://github.com/evanw/esbuild/issues/1930 was fixed right when we needed it :)

PR #258, opened 2022-09-02
Fix protoc option in protoc-gen-connect-web/README.md

ymmt2005:
`--os_opt` was wrong. `--es_opt` is correct.


timostamm (approved):
Thanks for the fix, Yamamoto - much appreciated.

PR #259, opened 2022-09-02
Updates to TypeScript 4.8.2

smaye81:
Update Connect Web to TypeScript 4.8.2 and in so doing, fix the breaking change on `ReadableStreamDefaultReadResult` naming.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "npm": ">=8"
   },
   "devDependencies": {
-    "typescript": "~4.7.4",
+    "typescript": "~4.8.2",
     "@typescript-eslint/eslint-plugin": "^5.36.1",
     "@typescript-eslint/parser": "^5.36.1",
     "eslint": "^8.21.0",
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "npm": ">=8"
   },
   "devDependencies": {
-    "typescript": "~4.8.2",
+    "typescript": "^4.8.2",
     "@typescript-eslint/eslint-plugin": "^5.36.1",
     "@typescript-eslint/parser": "^5.36.1",
     "eslint": "^8.21.0",
--- a/packages/connect-web-test/src/nodeonly/grpc-transport.ts
+++ b/packages/connect-web-test/src/nodeonly/grpc-transport.ts
@@ -17,7 +17,7 @@ import type {
   StreamResponse,
   UnaryRequest,
   UnaryResponse,
-  ReadableStreamDefaultReadResult,
+  ReadableStreamDefaultReadResultLike,
 } from "@bufbuild/connect-web";
 import {
   Code,
@@ -261,7 +261,7 @@ export function createGrpcTransport(
               method,
               header,
               trailer,
-              async read(): Promise<ReadableStreamDefaultReadResult<O>> {
+              async read(): Promise<ReadableStreamDefaultReadResultLike<O>> {
                 const outcome = await new Promise<O | Error | null>(
                   (resolve) => {
                     if (callEnded) {
--- a/packages/connect-web-test/src/nodeonly/grpc-transport.ts
+++ b/packages/connect-web-test/src/nodeonly/grpc-transport.ts
@@ -260,11 +260,7 @@ export function createGrpcTransport(
               method,
               header,
               trailer,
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -14,7 +14,7 @@
     "npm": ">=8"
   },
   "devDependencies": {
-    "typescript": "~4.7.4",
+    "typescript": "~4.8.2",
```
timostamm, line 17: I think we can go back to `^` here?

packages/connect-web/src/lib.dom.streams.ts
```diff
@@ -0,0 +1,32 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// The name for the ReadableStreamDefaultReadResult type changed in TypeScript 4.8.2 to ReadableStreamReadResult.  This
+// makes supporting multiple versions of TypeScript difficult if we use the ambient type declaration.  Instead, this
+// type is meant to mimic the type provided by TypeScript, just defined here so that we can have better control with
+// supporting various versions.  The suffix 'Like' has been added to the type name to differentiate it from the actual
+// type provided by TypeScript libs.
```
timostamm, line 19: Comment is on point :+1: 

Couple of comment style nits though:

Can you leave an empty line before the interface declaration though, so that the comment isn't attached to it?
Alternative, you could attach it to the exported element. In that case I'd recommend to move the exported type to the top, in front of the two interfaces.

Can you wrap the lines at 80 columns?

smaye81: Yep.  Done.  I added the comment to the exported element.

packages/connect-web/src/index.ts
```diff
@@ -51,3 +51,5 @@ export {
 } from "./http-headers.js";
 
 export { createEnvelopeReadableStream } from "./envelope.js";
+
+export { ReadableStreamDefaultReadResultLike } from "./lib.dom.streams.js";
```
timostamm, line 55: We must not export this. The user brings their own type. For example:

```ts
declare var a: StreamResponse<SayRequest>;
const b: ReadableStreamReadResult<SayRequest> = await a.read();
```

`r.read()` returns our `ReadableStreamDefaultReadResultLike`, but it's assignable to `ReadableStreamReadResult` and to `ReadableStreamDefaultReadResult`, which means our typings are compatible with both. As long as the types are assignable, we're good.

packages/connect-web/src/lib.dom.streams.ts
```diff
@@ -0,0 +1,34 @@
...
+// limitations under the License.
+
+// The name for the ReadableStreamDefaultReadResult type changed in TypeScript
+// 4.8.2 to ReadableStreamReadResult.  This makes supporting multiple versions
+// of TypeScript difficult if we use the ambient type declaration.  Instead,
+// this type is meant to mimic the type provided by TypeScript, just defined
+// here so that we can have better control with supporting various versions.
+// The suffix 'Like' has been added to the type name to differentiate it from
+// the actual type provided by TypeScript libs.
+export type ReadableStreamDefaultReadResultLike<T> =
```
timostamm, line 22: Apologies for not noticing before, but wouldn't it be a better fit to name this `ReadableStreamReadResultLike`, since we're at the new TS version now, and the dom lib type is named `ReadableStreamReadResult`?

packages/connect-web-test/src/nodeonly/grpc-transport.ts
```diff
@@ -260,7 +260,11 @@ export function createGrpcTransport(
               method,
               header,
               trailer,
-              async read(): Promise<ReadableStreamDefaultReadResult<O>> {
+              // Note this is using the TypeScript type ReadableStreamReadResult and not the custom type used by
+              // Connect-Web.  This mimics how users will interact with the library (using their own DOM lib) and tests
+              // compatibility with the two types in that the Connect Web custom type should always be assignable to
+              // the type in the user's DOM lib.
+              async read(): Promise<ReadableStreamReadResult<O>> {
```
timostamm, line 267: Just relying on type inference is probably an easy solution here:

```suggestion
              async read() {
```

timostamm (approved):

PR #268, opened 2022-09-15
Fix await res.json() in top-level README

nguyenyou:
Without `await`, we receive a promise instead of an expected value


timostamm (commented):
A perfect example of why you should use connect-web instead of fetch :wink:
Kidding aside, thank you for the fix, @nguyenyou!

Can I ask you to update the [README in packages/connect-web](https://github.com/bufbuild/connect-web/blob/main/packages/connect-web/README.md) as well?

PR #273, opened 2022-09-20
Removes unused release automation artifacts

smaye81:
We are iceboxing release automation for connect-web right now.  The release process is not very onerous and devoting effort to automating it is not the best use of our time.  We will revisit if/when the release process becomes more involved (i.e. after v1.0)


packages/connect-web-bench/README.md
```diff
@@ -10,5 +10,5 @@ it like a web server would usually do.
 
 | code generator | bundle size        | minified               | compressed           |
 |----------------|-------------------:|-----------------------:|---------------------:|
-| connect-web    | 87,712 b | 40,923 b | 11,344 b |
+| connect-web    | 87,899 b | 40,992 b | 11,384 b |
```
smaye81, line 13: Not sure why the bench numbers changed as a result of this.  Maybe this is related to an earlier change?

timostamm: Yes, forgot to do it in https://github.com/bufbuild/connect-web/pull/272, thanks for the update.

timostamm (approved):

PR #274, opened 2022-09-22
Respect binary encoding options in the connect transport

timostamm:
The Connect transport takes `binaryOptions` - rarely used options for the binary encoding - but never passed them to the serialization functions.

This fixes it.


akshayjshah (approved):
Is a unit test covering this case appropriate?

timostamm:
> Is a unit test covering this case appropriate?

Unfortunately, we would have to mock large parts of the fetch API for a unit test. Considering that this only passes on an argument that is already well covered in protobuf-es, it does not seem worth it to add test code several times the LoC of the actual code.

PR #288, opened 2022-10-03
Dependabot batched updates

paul-sachs:
More dependabot updates.


smaye81:
@paul-sachs not sure why the browserstack stuff is failing, but the `karma` updates could be the culprit maybe?

paul-sachs:
> @paul-sachs not sure why the browserstack stuff is failing, but the `karma` updates could be the culprit maybe?

Eh, who knows. I'll dig in.

paul-sachs:
So it seems the culprit is esbuild@0.5.8. Not exactly sure where or how, but it's causing safari 11.1 to hang during testing.

paul-sachs:
So finally just hooked up browserstack and pointed it at the karma runner and it looks like we've got some syntax that Safari doesn't like:

image

No idea why upgrading these deps made it worse but it looks like Safari 11.1 may not have ever worked properly and it was just not getting hit properly before, or there's a flag I need to use to compile this syntax.

packages/connect-web-bench/package.json
```diff
@@ -9,8 +9,8 @@
     "@bufbuild/connect-web": "0.2.1",
     "@bufbuild/protobuf": "0.1.1",
     "@bufbuild/protoc-gen-es": "0.1.1",
-    "esbuild": "^0.15.6",
-    "google-protobuf": "^3.20.1",
-    "grpc-web": "^1.3.1"
+    "esbuild": "^0.15.10",
+    "google-protobuf": "^3.21.0",
+    "grpc-web": "^1.4.1"
```
timostamm, line 14: These runtime dependencies should match the code generator version. 

Can you update packages/connect-web-bench/buf.gen.yaml? 
You will also need to run `make bench` to update to benchmark results.

packages/connect-web-test/karma-fixup-queueMicrotask.js
```diff
@@ -0,0 +1,26 @@
...
+// limitations under the License.
+
+/* eslint-disable */
+
+(function () {
+  if (typeof globalThis.queueMicrotask === "function") {
+    return;
+  }
+
+  globalThis.queueMicrotask = (callback) =>
```
timostamm, line 22: :+1:

packages/connect-web-bench/package.json
paul-sachs: Ah, will do.

paul-sachs: I'll also update the notion doc to remind anyone doing dependabot updates to do this.

paul-sachs: Alright, updated.

timostamm (approved):
Thank you, especially for the karma fix!

PR #318, opened 2022-11-09
updates `service` type to reflect only the necessary properties

dimitropoulos:
I hit a use case in generating a plugin that requires connect-web for transport but doesn't need the generation of services.  In that situation, this was the only blocker I am aware of that forced me to use a service.  Come to find out, then, that the full service isn't actually required.  This PR addresses that by only asking for the required fields.  It's a backwards compatible change because a full service will still structurally match.

### DIFF
--- a/packages/connect-web/src/connect-transport.ts
+++ b/packages/connect-web/src/connect-transport.ts
@@ -107,7 +107,7 @@ export function createConnectTransport(
       I extends Message<I> = AnyMessage,
       O extends Message<O> = AnyMessage
     >(
-      service: ServiceType,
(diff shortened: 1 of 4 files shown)


timostamm:
Thanks for the initiative, Dimitri!

I'm afraid now is not a good time for this change, because we are working on the next release in the connect-web-next package, and there are a couple of things in flight that might be complicated by this change.

Closing this for - let's revisit after we switched @bufbuild/connect-web to use @bufbuild/connect-core. In the meantime, you can simply pass `{typeName: "...", methods: {}}` as a valid `ServiceType`.

PR #320, opened 2022-11-10
Support the construction of errors with details

timostamm:
We want to enable service implementations to raise an error with details. This changes the `details` property of `ConnectError` to hold both incoming _and_ outgoing error details, which need to be represented in slightly different forms.

This is technically a breaking change because the property is public, but we always recommended to use the connectErrorDetails() to obtain details from an error. User-provided transport implementations that parse error details will need to update.

This re-introduces the `details` argument to the ConnectError constructor that was removed in #164 and deprecated in #272. Because we change the order of arguments, the new constructor signature is compatible with the deprecation.

This also adds the functions connectCodeToHttpStatus() and connectErrorToJson().

### DIFF
--- a/packages/connect-core/src/connect-code-to-http-status.ts
+++ b/packages/connect-core/src/connect-code-to-http-status.ts
@@ -1,0 +1,44 @@
+import { Code } from "./code.js";
+
+/**
+ * Returns a HTTP status code for the given Connect code.
+ * See https://connect.build/docs/protocol#error-codes
+ */
+export function connectCodeToHttpStatus(code: Code): number {
+  switch (code) {
+    case Code.Canceled:
+      return 408; // Request Timeout
+    case Code.Unknown:
+      return 500; // Internal Server Error
+    case Code.InvalidArgument:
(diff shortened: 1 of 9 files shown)


smaye81:
Will anything need to change here as a result of https://github.com/bufbuild/protobuf-es/pull/294?  I know we had mentioned that once we implement those changes, we should update connect-web - the implementation of `connectErrorFromJson()` could be more simple.

smaye81 (approved):

timostamm:
> Will anything need to change here as a result of [bufbuild/protobuf-es#294](https://github.com/bufbuild/protobuf-es/pull/294)? I know we had mentioned that once we implement those changes, we should update connect-web - the implementation of `connectErrorFromJson()` could be more simple.

The unpack() method from https://github.com/bufbuild/protobuf-es/pull/288 will simplify the code [here](https://github.com/bufbuild/connect-web/blob/2f1d825cae73648a411b5c9dae843be50a960054/packages/connect-core/src/connect-error.ts#L145).

PR #328, opened 2022-11-16
Add handler options

timostamm:
Update handler options to accept a type registry (along with other JSON or binary format options), or an array of protocols.

This also adds a helper for unimplemented services: Users can fill in a partial service implementation with methods that raise an `unimplemented` error.


packages/connect-node/src/connect-protocol.ts
```diff
@@ -35,7 +35,7 @@ import {
 } from "@bufbuild/protobuf";
 import type { ImplHandler, ImplSpec, Protocol } from "./protocol.js";
 import { createServerMethodSerializers } from "./private/create-server-method-serializers.js";
-import type { HandlerContext } from "./handler.js";
+import type { HandlerContext } from "./implementation";
```
smaye81, line 38: `.js` extension?

packages/connect-node/src/handler.ts
```diff
@@ -58,83 +52,148 @@ export type Handler = NodeHandler & {
...
+    handleProtocol(req, res).catch((reason) => {
+      // TODO need to handle rejections here, but it's unclear how exactly
+      // eslint-disable-next-line no-console
+      console.error("protocol handle failed", reason);
+    });
+  }
+  return Object.assign(handleAny, {
+    service,
+    method,
+    requestPath: createMethodUrl("https://example.com", service, method)
```
smaye81, line 121: Is this `example.com` url intended to be here?

smaye81 (approved):

packages/connect-node/src/connect-protocol.ts
timostamm: Updated

packages/connect-node/src/handler.ts
timostamm: The function requires a valid `baseUrl` argument, even though we only want the path of the resulting URL in the end.
I agree it's irritating to read, and I hope we find a better way to do this.

PR #330, opened 2022-11-17
Add a function to add a gRPC status to a header collection

timostamm:
... and a function to get a gRPC error status from a header collection.

### DIFF
--- a/packages/connect-core/src/grpc-trailer-status.ts
+++ b/packages/connect-core/src/grpc-trailer-status.ts
@@ -22,6 +22,13 @@ const fieldGrpcStatusDetailsBin = "grpc-status-details-bin",
   fieldGrpcStatus = "grpc-status",
   fieldGrpcMessage = "grpc-message";
 
+/**
+ * Sets the fields "grpc-status" and "grpc-message" in the given
+ * Headers object.
+ * If an error is given and contains error details, the function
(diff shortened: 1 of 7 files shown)


packages/connect-core/src/grpc-trailer-status.ts
```diff
@@ -0,0 +1,106 @@
...
+ * Returns an error only if the gRPC status code is > 0.
+ */
+export function grpcFindTrailerError(
+  headerOrTrailer: Headers
+): ConnectError | undefined {
+  // Prefer the protobuf-encoded data to the grpc-status header.
+  const statusBytes = headerOrTrailer.get(fieldGrpcStatusDetailsBin);
+  if (statusBytes != null) {
+    const status = decodeBinaryHeader(statusBytes, Status);
+    if (status.code == 0) {
```
smaye81, line 73: Nit.  Prob should be `===` yeah?

smaye81 (approved):

packages/connect-core/src/grpc-trailer-status.ts
timostamm: We're comparing a guaranteed number against 0, so we're safe from implicit conversion and `==` is equivalent to `===` here.

PR #332, opened 2022-11-18
[TCN-574] - Implement gRPC web transport over node http2

gilwong00:


packages/connect-core/src/grpc-web-create-request-header.ts
```diff
@@ -16,14 +16,16 @@
  * Creates headers for a gRPC-web request.
  */
 export function grpcWebCreateRequestHeader(
+  useBinaryFormat: boolean,
```
timostamm, line 19: :+1: Now that we have added support for JSON with gRPC-web for connect-node, we should add this option to connect-web as well.

packages/connect-node/src/private/io.ts
```diff
@@ -77,7 +77,7 @@ export function read(
 ): Promise<ReadableStreamReadResultLike<Uint8Array>> {
   return new Promise<ReadableStreamReadResultLike<Uint8Array>>(
     (resolve, reject) => {
-      if (stream.readableEnded) {
+      if (stream.readableEnded || size === 0) {
```
timostamm, line 80: Thank you for adding this, I was thinking about that as well. I hope that we can remove this function completely, and replace it with either an object stream, or a [web stream](https://nodejs.org/api/webstreams.html#web-streams-api). This should allow us to clean up the logic for [handlers](https://github.com/bufbuild/connect-web/blob/d72c9237625ae160c2b096efbf6cf5e4ee22cf2a/packages/connect-node/src/connect-protocol.ts), which is unnecessarily (and painfully) verbose right now. But I think we need to look at the gRPC protocol first.

timostamm (approved):
🎉 Thank you for the implementation, Gilbert.

timostamm:
@gilwong00, merging this for you so that it doesn't become stale.

packages/connect-core/src/grpc-web-create-request-header.ts
timostamm: Added in https://github.com/bufbuild/connect-web/pull/334

PR #334, opened 2022-11-22
Allow JSON in the gRPC-web transport of @bufbuild/connect-web

timostamm:
This adds a new feature to @bufbuild/connect-web. With the following call:

```ts
import { createGrpcWebTransport } from "@bufbuild/connect-web-next";

createGrpcWebTransport({
   baseUrl: "https://demo.connect.build",
   useBinaryFormat: false
});
```

Messages are sent in the JSON format. The default behavior format remains binary, because not all server implementations support the JSON format.

### DIFF
--- a/packages/connect-node-test/src/crosstest/ping_pong.spec.ts
+++ b/packages/connect-node-test/src/crosstest/ping_pong.spec.ts
@@ -27,9 +27,18 @@ describe("ping_pong", () => {
     [
       // This test does full duplex, receiving before it closes the input.
       // In this case, your connect fetch transport never issues the request.
-      "connect fetch transport (binary) against Node.js (http)",
-      "connect fetch transport (JSON) against Node.js (http)",
-      "gRPC-web fetch transport (binary) against Node.js (http)",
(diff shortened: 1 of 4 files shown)


packages/connect-node-test/src/crosstest/ping_pong.spec.ts
```diff
@@ -27,9 +27,18 @@ describe("ping_pong", () => {
     [
       // This test does full duplex, receiving before it closes the input.
       // In this case, your connect fetch transport never issues the request.
-      "connect fetch transport (binary) against Node.js (http)",
-      "connect fetch transport (JSON) against Node.js (http)",
-      "gRPC-web fetch transport (binary) against Node.js (http)",
+      "@bufbuild/connect-web (Connect, binary) against connect-go (h1)",
+      "@bufbuild/connect-web (Connect, binary) against connect-go (h2)",
+      "@bufbuild/connect-web (Connect, binary) against @bufbuild/connect-node (h1)",
```
timostamm, line 32: The list of transports to test is growing. Renamed the keys for better readability.

smaye81 (approved):

PR #337, opened 2022-11-23
Upgrade connect-web to v0.3.0 of protobuf-es

smaye81:
This upgrades all dependencies on protobuf-es packages to v0.3.0.  In the process, fixes a parameter type using `PartialMessage` as that has slightly changed.


packages/connect-core/src/callback-client.ts
```diff
@@ -124,7 +124,7 @@ function createUnaryFn<I extends Message<I>, O extends Message<O>>(
  * ServerStreamingFn is the method signature for a server-streaming method of
  * a CallbackClient.
  */
-type ServerStreamingFn<I extends Message, O extends Message> = (
+type ServerStreamingFn<I extends Message<I>, O extends Message<O>> = (
```
smaye81, line 127: @timostamm We should probably mention this in the release notes for protobuf-es, yeah?

timostamm (approved):

PR #340, opened 2022-11-24
add generics for Protocol.createHandler

fubhy:
Adds type coverage for the `impl` function.

### DIFF
--- a/packages/connect-node/src/protocol.ts
+++ b/packages/connect-node/src/protocol.ts
@@ -33,8 +33,12 @@ import type {
  */
 export interface Protocol {
   supportsMediaType(type: string): boolean;
-
-  createHandler(spec: ImplSpec): ImplHandler;
+  createHandler<
+    I extends Message<I> = AnyMessage,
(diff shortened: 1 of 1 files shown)


timostamm:
@fubhy, I've been going back and forth on this. If the return type also had a related type argument, generics could ensure that they match:

```ts
createHandler<T>(spec: ImplSpec<T): ImplHandler<T>
```

But it doesn't, and it's fair to implement a Protocol and use the `ImplSpec` with its default type arguments. We don't because it's a little bit helpful to add none-default type arguments, so that we don't confuse `I` and `O`, but why not leave that up to the implementation?

I can see both sides, but I'm leaning towards the simple version we currently have. Closing this, but thank you for the PR!

fubhy:
Agreed. This only came up while experimenting with other parts (more specifically the testing use-case that I was playoung around with as mentioned on Slack). There, I was trying to work with individual handlers. But I realized (as you also suggested) that it's perfectly fine to just use `unimplementService` for that case so my use-case is solved differently now anyways :-)

PR #342, opened 2022-11-24
Handle invalid input for unary and server streaming methods

fubhy:
This is already handled for bidi and client streaming methods but not for unary & server streaming.

Currently, when passing invalid JSON (or nothing), the server crashes while trying to parse the input as JSON.

### DIFF
--- a/packages/connect-node/src/connect-protocol.ts
+++ b/packages/connect-node/src/connect-protocol.ts
@@ -108,9 +108,9 @@ export function createConnectProtocol(
                 type.binary
               );
 
-            const input = parse(await readToEnd(req));
             let output: O | PartialMessage<O>;
             try {
+              const input = parse(await readToEnd(req));
               output = await spec.impl(input, context);
             } catch (e) {
               return await endWithConnectUnaryError(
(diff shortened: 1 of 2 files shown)


timostamm:
> Currently, when passing invalid JSON (or nothing), the server crashes while trying to parse the input as JSON.

Does the server die, or does the error end up on [this line](https://github.com/bufbuild/connect-web/blob/4c377bfd7b6dc214c1013fe26450f1dd948d0294/packages/connect-node/src/handler.ts#L114)? If it's the former, please let me know!

The current implementation is pretty naive and incomplete. We'll have to implement deadlines and the gRPC protocol first. There will be a round of refactorings after that (I hope that we can drastically reduce the similar code between the protocol handler implementations). I think only then can we look at proper error handling and hardening the handlers. 

Let me close this PR, because I think that it's best to keep it intentionally naive for now. But thank you for poking at it! 😆

fubhy:
@timostamm ok no problem. I just noticed that the input parsing for streaming handlers is happening inside the try/catch... So the behavior is currently different between the two.

So for invalid input (or an empty payload) the server crashes for unary methods but handles it for client and bidi streams.

timostamm:
Ah, good call, I see how that's inconsistent 👍

PR #345, opened 2022-11-25
Implement gRPC handler and http2 client

timostamm:
This adds the gRPC protocol for Node.js handlers, and a transport using the Node.js `http2` package for clients.

The test "timeout_on_sleeping_server" is skipped - we still need to implement deadlines. All other crosstests pass.


Makefile
```diff
@@ -147,8 +147,8 @@ testcore: $(BUILD)/connect-core
 .PHONY: testnode
 testnode: $(BIN)/node16 $(BIN)/node18 $(BUILD)/connect-node-test
 	$(MAKE) crosstestserverrun
-	cd packages/connect-node-test && PATH="$(abspath $(BIN)):$(PATH)" NODE_TLS_REJECT_UNAUTHORIZED=0 node16 ../../node_modules/.bin/jasmine --config=jasmine.json
-	cd packages/connect-node-test && PATH="$(abspath $(BIN)):$(PATH)" NODE_TLS_REJECT_UNAUTHORIZED=0 node18 ../../node_modules/.bin/jasmine --config=jasmine.json
+	cd packages/connect-node-test && PATH="$(abspath $(BIN)):$(PATH)" node16 ../../node_modules/.bin/jasmine --config=jasmine.json
```
timostamm, line 150: We shouldn't ignore certificate errors globally in our tests. We now use a valid one for tests between @bufbuild/connect-node and @bufbuild/connect-node.

I wasn't able to configure the @bufbuild/connnect-node clients to work with the certificate used by connect-go. These client/server combinations now set the option `rejectUnauthorized: false`, so we only ignore errors for the specific combination. Added TODOs to investigate later.

smaye81 (approved):

PR #347, opened 2022-11-28
add method and service info to the handler context

fubhy:
Not sure if you want to have those generics on the HandlerContext interface. They help to narrow the method info type slightly but don't add that much value. Your call.

### DIFF
--- a/packages/connect-node/src/connect-protocol.ts
+++ b/packages/connect-node/src/connect-protocol.ts
@@ -92,7 +92,7 @@ export function createConnectProtocol(
                 "Unsupported Media Type"
               );
             }
-            const context: HandlerContext<I, O> = {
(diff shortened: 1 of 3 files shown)


timostamm:
> Not sure if you want to have those generics on the HandlerContext interface. They help to narrow the method info type slightly but don't add that much value. Your call.

It's not a perfectly clear case, but let's not add the type parameters.

fubhy:
> > Not sure if you want to have those generics on the HandlerContext interface. They help to narrow the method info type slightly but don't add that much value. Your call.
> 
> It's not a perfectly clear case, but let's not add the type parameters.

Agreed. Can always add them later if there's a need.

timostamm (approved):
LGTM!

PR #349, opened 2022-11-28
fix method implementation return types

fubhy:
Especially for the stream handlers, there's a difference between ...

```typescript
AsyncIterable<O> | AsyncIterable<PartialMessage<O>>
```

vs. 

```typescript
AsyncIterable<O | PartialMessage<O>>
```

I believe the latter is actually what we want here (so the iterable can alternate between the two, if that's desirable even).

Either way, typescript seems to struggle also with `Promise<O> | Promise<PartialMessage<O>>` when used with some generic decorators that I was playing around with.


timostamm (approved):
Yep, that's fair.

PR #358, opened 2022-12-02
[TCN-576] Connect Http and Https Transport

gilwong00:
This adds the connect transport using the Node.js http and https package for clients.


packages/connect-node/src/connect-http2-transport.ts
```diff
@@ -99,6 +99,8 @@ export interface ConnectHttp2TransportOptions {
   binaryOptions?: Partial<BinaryReadOptions & BinaryWriteOptions>;
 }
 
+const messageFlag = 0b00000000;
```
timostamm, line 102: TY!

packages/connect-node/src/private/node-http-request.ts
```diff
@@ -0,0 +1,85 @@
...
+   */
+  jsonOptions?: Partial<JsonReadOptions & JsonWriteOptions>;
+
+  /**
+   * Options for the binary wire format.
+   */
+  binaryOptions?: Partial<BinaryReadOptions & BinaryWriteOptions>;
+}
+
+export async function nodeRequest(options: NodeRequestOptions) {
```
timostamm, line 49: I truly sympathize with the idea to split the code into discrete logical units. But it's too early. Can you unroll this into the transport?

packages/connect-node-test/src/helpers/testserver.ts
```diff
@@ -318,8 +317,23 @@ vKy9wyvtUtg=
         },
         useBinaryFormat: false,
       }),
-
-    // gRPC-web
+    "@bufbuild/connect-node (Connect, JSON, http) against connect-go (h1)": (
```
timostamm, line 320: Now that we have the first http1 client, can you also add transports that test against the Node.js http1 server? 


```ts
    "@bufbuild/connect-node (Connect, binary, http1) against @bufbuild/connect-node (h1)":
      (options?: Record<string, unknown>) =>
        createConnectHttp2Transport({
          ...options,
          baseUrl: servers["@bufbuild/connect-node (h1)"].getUrl(),
          useBinaryFormat: true,
        }),
    "@bufbuild/connect-node (Connect, JSON, http1) against @bufbuild/connect-node (h1)":
      (options?: Record<string, unknown>) =>
        createConnectHttp2Transport({
          ...options,
          baseUrl: servers["@bufbuild/connect-node (h1)"].getUrl(),
          useBinaryFormat: false,
        }),
```

If it doesn't work out of the box, let's sync. It's very possible that the http1 _server_ isn't working correctly - this will be the first time we test it.

packages/connect-node/src/connect-http-transport.ts
```diff
@@ -0,0 +1,353 @@
...
+            const response = await nodeRequest({
+              req,
+              useBinaryFormat,
+              jsonOptions: options.jsonOptions,
+              binaryOptions: options.binaryOptions,
+            });
+
+            const responseHeaders = nodeHeaderToWebHeader(response.headers);
+            await validateResponseHeader(
+              response.statusCode as number,
```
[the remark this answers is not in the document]
gilwong00: unfortunately the type for `statusCode` is `IncomingMessage.statusCode?: number | undefined` which is why i casted. I could change it up to check if we have a status code or default to 500?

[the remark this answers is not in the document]
timostamm: I think that is because `http.IncomingMessage` is used both for client responses (where it should always be present, because a HTTP/1 response always has a status), and for incoming server requests (where it cannot be present by definition).

Since we are on the client side here, I think it is fair that we error on a response that cannot be valid. You can use private/assert.ts for this:

```ts
assert(typeof response.statusCode == "number", "http1 client response is missing status code");
```

gilwong00:
@timostamm https server and https + connect-go test have been added

packages/connect-node/src/connect-http-transport.ts
```diff
@@ -0,0 +1,415 @@
...
+  encoding: "utf8" | "binary";
+
+  // Request body
+  payload: Uint8Array;
+}
+
+/**
+ * Create a Transport for the connect protocol on Http.
+ *
+ */
```
timostamm, line 125:
```suggestion
/**
 * Create a Transport for the Connect protocol using the Node.js `http` or 
 * `https` package.
 */
```

timostamm (approved):
Just one JSDoc suggestion. Feel free to merge without re-review.

PR #364, opened 2022-12-06
[TCN - 577] - gRPC Web Http and Https Transports

gilwong00:
This adds the gRPC web transport using the Node.js http and https package for clients.


packages/connect-node/src/grpc-web-http-transport.ts
```diff
@@ -0,0 +1,421 @@
...
+          } catch (e) {
+            throw connectErrorFromNodeReason(e);
+          }
+        }
+      );
+    },
+  };
+}
+
+function makeNodeRequest(options: NodeRequestOptions) {
```
gilwong00, line 368: @timostamm since this is now being used in the connect and gRPC web transport, should we consider making this reusable or continue to leave it at the transport level?

timostamm (approved):
LGTM!

PR #365, opened 2022-12-06
Support compression in the Connect protocol handler, and in Connect http2 clients

timostamm:
This introduces the `Compression` interface, which ties a compression algorithm - for example gzip - to a name.

Handlers will have the following new options:

```ts
  acceptCompression?: Compression[];
  compressMinBytes?: number;
  readMaxBytes?: number;
  sendMaxBytes?: number;
```

By default, handlers accept `gzip` and `brotli` compression.

Clients will have the same new options, but also the following one:

```ts
  sendCompression?: Compression;
```

A client can opt in to request compression by providing one of the exported `compression*` constants (or a custom one). If the client provides `acceptCompression` for a request, the server may choose to use compression in the response.

For the gRPC and the gRCP-web protocol, and for all Connect streaming methods, requests and responses contain a header with the compression name. If the least significant bit of the envelope is set, the data in the envelope is compressed.

For Connect unary methods, the full body is compressed, and the standard `Content-Encoding` header contains the compression name.

Compression is only applied if a serialized message is larger than compressMinBytes.

Note that this PR does not add compression everywhere, only for Connect protocol handlers, and only for Connect http2 clients. That means we still have to implement compression for gRPC and gRPC-web handlers, and for all other Transports.

### DIFF
--- a/packages/connect-node/src/connect-http2-transport.ts
+++ b/packages/connect-node/src/connect-http2-transport.ts
@@ -14,13 +14,13 @@
 
 import {
   appendHeaders,
+  Code,
   connectCodeFromHttpStatus,
-  connectCreateRequestHeader,
   connectEndStreamFlag,
   connectEndStreamFromJson,
   ConnectError,
   connectErrorFromJson,
-  connectExpectContentType,
+  connectParseContentType,
   connectTrailerDemux,
   createClientMethodSerializers,
   createMethodUrl,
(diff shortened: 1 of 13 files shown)


packages/connect-node/src/connect-http2-transport.ts
```diff
@@ -56,7 +58,21 @@ import {
...
+import { compressionBrotli, compressionGzip } from "./compression.js";
+import { validateReadMaxBytesOption } from "./private/validate-read-max-bytes-option.js";
+
+const messageFlag = 0b00000000;
+/**
+ * compressedFlag indicates that the data in a EnvelopedMessage is
+ * compressed. It has the same meaning in the gRPC-Web, gRPC-HTTP2,
+ * and Connect protocols.
+ */
+const compressedFlag = 0b00000001;
```
gilwong00, line 75: since we will probably use this flag in the other transports, does it make sense to export this from the compression lib so we can import it instead of redeclaring it in the other transports?

gilwong00 (approved):

smaye81 (approved):

PR #367, opened 2022-12-08
[TCN-575] gRPC Http and Https Transports

gilwong00:
This adds the gRPC transport using the Node.js http and https package for clients.

### DIFF
--- a/packages/connect-node/src/grpc-http-transport.ts
+++ b/packages/connect-node/src/grpc-http-transport.ts
@@ -24,7 +24,10 @@ import {
   grpcExpectContentType,
   grpcFindTrailerError,
   Interceptor,
+  runStreaming,
   runUnary,
+  StreamingConn,
+  StreamingRequest,
   Transport,
   UnaryRequest,
   UnaryResponse,
@@ -46,7 +49,9 @@ import {
   webHeaderToNodeHeaders,
 } from "./private/web-header-to-node-headers.js";
 import { assert } from "./private/assert.js";
-import { readEnvelope } from "./private/io.js";
+import { end, readEnvelope, write } from "./private/io.js";
+import type { ReadableStreamReadResultLike } from "./lib.dom.streams.js";
+import { defer } from "./private/defer.js";
 
 export interface GrpcHttpTransportOptions {
   /**
@@ -107,7 +112,7 @@ const messageFlag = 0b00000000;
 
 export function createGrpcHttpTransport(
   options: GrpcHttpTransportOptions
-): Partial<Transport> {
+): Transport {
   const useBinaryFormat = options.useBinaryFormat ?? false;
   return {
     async unary<
@@ -164,7 +169,7 @@ export function createGrpcHttpTransport(
             );
 
             const messageResult = await readEnvelope(response);
-            const trailer = mapResponseTrailers(response.trailers);
+            let trailer = mapResponseTrailers(response.trailers);
             validateGrpcStatus(trailer);
 
             if (messageResult.done) {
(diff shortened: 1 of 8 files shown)


packages/connect-node/src/grpc-http-transport.ts
```diff
@@ -0,0 +1,408 @@
...
+
+            const eofResult = await readEnvelope(response);
+            trailer = mapResponseTrailers(response.trailers);
+            if (receivedTrailers(trailer)) {
+              validateGrpcStatus(trailer);
+            }
+
+            if (!eofResult.done) {
+              throw "extraneous data";
+            }
```
timostamm, line 195: The logic of `receivedTrailers` doesn't seem solid to me. Trailers are crucial in gRPC because they contain the status, including the successful `grpc-status: 0`. It is a protocol error if we don't receive a grpc-status, and it appears that we're just ignoring this case.

The documentation states that trailers are "populated at the 'end' event". We can make use of that, a create a promise for the trailer early on, same as we do in the http2 gRPC transport. This allows us to bring the two implementations closer together:

```suggestion
            const responseHeaders = nodeHeaderToWebHeader(response.headers);
            const trailerPromise = readHttp1ResponseTrailer(response);
            assert(
              typeof response.statusCode == "number",
              "http1 client response is missing status code"
            );
            validateResponse(
              useBinaryFormat,
              response.statusCode,
              responseHeaders
            );
            const messageResult = await readEnvelope(response);
            const eofResult = await readEnvelope(response);
            if (!eofResult.done) {
              throw "extraneous data";
            }
            const trailer = await trailerPromise;
            validateGrpcStatus(trailer);
            if (messageResult.done) {
              throw "premature eof";
            }
```

The function readHttp1ResponseTrailer (can put this next to readResponseTrailer in io.ts):

```ts
/**
 * Returns a promise for HTTP/1 response trailers.
 */
export function readHttp1ResponseTrailer(
  response: http.IncomingMessage
): Promise<Headers> {
  return new Promise<Headers>((resolve, reject) => {
    if (response.errored) {
      return reject(response.errored);
    }
    if (response.readableEnded) {
      return resolve(nodeHeaderToWebHeader(response.trailers));
    }
    response.once("error", error);
    response.once("end", end);

    function end() {
      response.off("error", error);
      response.off("end", end);
      resolve(nodeHeaderToWebHeader(response.trailers));
    }

    function error(err: Error) {
      response.off("error", error);
      response.off("end", end);
      reject(err);
    }
  });
}
```

packages/connect-node/src/grpc-http-transport.ts
```diff
@@ -0,0 +1,408 @@
...
+
+            const responsePromise = new Promise<http.IncomingMessage>(
+              (resolve, reject) => {
+                stream.on("response", (res) => {
+                  resolve(res);
+                });
+                stream.on("error", reject);
+              }
+            );
+            const responseTrailer = defer<Headers>();
```
timostamm, line 263: Can you switch this to a promise from `readHttp1ResponseTrailer()` as well? The diff between grpc-http-transport.ts and grpc-http2-transport.ts should be as small as possible. (Don't put extra work on it now, but please make sure we handle trailers exactly the same).

packages/connect-node/src/grpc-http-transport.ts
timostamm: See https://github.com/bufbuild/connect-web/pull/368 for a change to the http2 transport that takes the same code path as suggested above.

packages/connect-node/src/grpc-http-transport.ts
```diff
@@ -0,0 +1,408 @@
...
+                  throw connectErrorFromNodeReason(e);
+                }
+              },
+            };
+            return Promise.resolve(conn);
+          } catch (e) {
+            throw connectErrorFromNodeReason(e);
+          }
+        }
+      );
```
timostamm, line 329: runStreaming() takes three arguments. You're missing the third one here, you need to pass `options.interceptors`.

Let's make the argument optional: https://github.com/bufbuild/connect-web/pull/369

timostamm (approved):

PR #370, opened 2022-12-08
[TCN-840] - Add Connect Protocol Version Header 

gilwong00:
Add connect protocol version header to connect transport


packages/connect-core/src/connect-create-request-header.ts
```diff
@@ -33,5 +35,6 @@ export function connectCreateRequestHeader(
   if (timeoutMs !== undefined) {
     result.set("Connect-Timeout-Ms", `${timeoutMs}`);
   }
+  result.set("connect-protocol-version", protocolVersion);
```
timostamm, line 38: Can you use uppercase starting letters, as in `Connect-Protocol-Version`, like we do for other fields?

packages/connect-web/src/connect-transport.ts
```diff
@@ -381,6 +382,7 @@ function createConnectRequestHeaders(
   if (timeoutMs !== undefined) {
     result.set("Connect-Timeout-Ms", `${timeoutMs}`);
   }
+  result.set("connect-protocol-version", protocolVersion);
```
timostamm, line 385: Same here.

timostamm:
Leaving this comment to make sure we don't release this breaking change to @bufbuild/connect-web without making sure that:
- the public BSR is allowlisting the new header
- we [explain the change](https://github.com/bufbuild/connect-docs/pull/168) in the release notes
- we recommend in the release notes that users make sure their backend is already allowlisting the header
- we give an example in the release notes for an interceptor that strips the header, as a temporary escape hatch

timostamm:
@gilwong00, regarding the example interceptor - could you write one, verify via [the example](https://github.com/bufbuild/connect-web/tree/main/packages/example) that it does remove the header as expected, and paste it here as a comment so that we can later add it to the release notes?

gilwong00:
Hey @timostamm here is a basic interceptor that will strip away the connect protocol version header

```
export function removeConnectProtocolHeader(): Interceptor {
  return (next) => async (req) => {
    if (req.header.has("Connect-Protocol-Version")) {
      req.header.delete("Connect-Protocol-Version");
    }
    const res = await next(req);
    return res;
  };
}
```

To use it, all we need to do is add it to the `interceptors` array when declaring the transport
```
  "connect JSON transport": (options) =>
    createConnectTransport({
      ...options,
      baseUrl,
      useBinaryFormat: false,
      interceptors: [removeConnectProtocolHeader()],
    }),
```

do you want me to push interceptor up?

timostamm:
> Hey @timostamm here is a basic interceptor

TY 🎉 

> do you want me to push interceptor up?

The comment is sufficient :)

timostamm (approved):

gilwong00:
Hey @timostamm just wanted to follow up on this one, are we able to merge this in or are we still waiting?

PR #374, opened 2022-12-15
Add timeout and deadline support to protocols

gilwong00:
This adds timeout and deadline support


packages/connect-node/src/grpc-web-protocol.ts
```diff
@@ -65,9 +68,14 @@ export function createGrpcWebProtocol(
       switch (spec.kind) {
         case MethodKind.Unary:
           return async (req, res) => {
+            const requestHeader = nodeHeaderToWebHeader(req.headers);
             const type = grpcWebParseContentType(
               req.headers["content-type"] ?? null
             );
+            const timeout = parseGrpcTimeoutFromHeader(
```
timostamm, line 75: If this fails to parse, we are raising an error, which is only going to be caught in the fallback [here](https://github.com/bufbuild/connect-web/blob/6ca3d00ca3f59f735df9b014427049559220c83d/packages/connect-node/src/handler.ts#L123). 

Instead, we need to return this error to the client. So for this protocol, you need to call `endWithGrpcWebTrailer()` with the error that occurred during parsing of `grpc-timeout`.

Please parse the timeout _after_ checking whether the content-type is valid. We should not try to parse the `grpc-timeout` header if we don't recognize the `content-type` header.

I think this is passing the tests because `timeout_on_sleeping_server` also goes green for `Code.Unknown`, which is a relic from testing against envoy. Can you remove `Code.Unknown` from timeout_on_sleeping_server.spec.ts:51 to confirm?

packages/connect-node/src/connect-protocol.ts
```diff
@@ -519,6 +570,22 @@ export function createConnectProtocol(
             const type = connectParseContentType(
               requestHeader.get(headerContentType)
             );
+            const timeout = parseInt(
+              requestHeader.get(connectTimeoutHeader) ?? ""
+            );
```
timostamm, line 575: I don't think we can get away with `parseInt` here (it also accepts exponential notation and negative values). Best to add a connectParseTimeoutMs() function to connect-core. 

If parsing of the timeout value failed, you need to call `endWithConnectEndStream()` for all streaming methods, and `endWithConnectUnaryError()` for unary methods, and stop processing. 

Please parse the timeout _after_ checking whether the content-type is valid. We should not try to parse the `Connect-Timeout-Ms` header if we don't recognize the `content-type` header. In case of the Connect protocol, we should also fail on the `Connect-Protocol-Version` header before we fail on an invalid timeout.

packages/connect-node/src/grpc-protocol.ts
```diff
@@ -65,20 +67,35 @@ export function createGrpcProtocol(
       switch (spec.kind) {
         case MethodKind.Unary:
           return async (req, res) => {
+            const requestHeader = nodeHeaderToWebHeader(req.headers);
             const type = grpcParseContentType(
               req.headers["content-type"] ?? null
             );
+            const timeout = parseGrpcTimeoutFromHeader(
+              requestHeader.get(grpcTimeoutHeader) ?? ""
+            );
```
timostamm, line 76: If we fail to parse grpc-timeout, you need to call `endWithGrpcWebTrailer()`. 

Please parse the timeout _after_ checking whether the content-type is valid. We should not try to parse the `grpc-timeout` header if we don't recognize the `content-type` header.

packages/connect-node/src/private/web-header-to-node-headers.ts
```diff
@@ -67,3 +77,47 @@ export function webHeaderToNodeHeaders(
   }
   return o;
 }
+
+export function parseGrpcTimeoutFromHeader(value: string): number | null {
```
timostamm, line 81: Can you move this function to connect-core? Best to prefix the name with "grpc", as in "grpcParseTimeout". 

It will be helpful to accept `string | null` as an argument, similar to `grpcParseContentType()` - this lets us pass a value from a `Headers` object without extra if-statements.

The errors are all correct, but I think it might be easier to work with if we didn't raise an error here, and instead _returned_ an error. We have three possible return values if we accept `string | null` as input:
1. `number` - parsed timeout in milliseconds
2. `undefined` - no input was given
3. `?` - input is invalid

So this could be `function grpcParseTimeout(val: string|null): number | undefined | ConnectError`, used as:

```ts
const timeout = grpcParseTimeout(requestHeader.get("Grpc-Timeout"));
if (typeof timeout == "number") {
  res.setTimeout(...);
}

// validate request content type

if (timeout instanceof ConnectError) {
  return void endWithGrpcWebTrailer(
    res,
    context,
    timeout
  );
}
```

Am open to alternatives here, but I think we should avoid raising errors because it complicates control flow.

packages/connect-node/src/private/web-header-to-node-headers.ts
```diff
@@ -67,3 +77,47 @@ export function webHeaderToNodeHeaders(
...
+  if (value.length === 0) {
+    return null;
+  }
+
+  const result = value.match(/[a-z]+|[^a-z]+/gi); // splits number from character
```
timostamm, line 86: `^(\d{1,8})([HMSmun])$` shifts more work to the reg exp. You can safely call `parseInt` on the first capture group, you're guaranteed to only get valid units, and you don't need to test > 99999999 anymore.

Best to add some test cases for this :)

packages/connect-node/src/grpc-web-protocol.ts
gilwong00: removed the check for `Code.Unknown` and test still pass

packages/connect-node/src/private/web-header-to-node-headers.ts
gilwong00: thats a great idea and definitely in favor of the protocol returning the error instead of raising here in the check

packages/connect-core/src/grpc-parse-timeout.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { ConnectError } from "./connect-error.js";
+import { grpcParseTimeout } from "./grpc-parse-timeout.js";
+
+describe("grpcParseTimeout()", function () {
+  it("should parse proper timeout", () => {
+    expect(grpcParseTimeout("1H")).toEqual(3600000);
+  });
```
timostamm, line 21: Let's make sure we allow 8 digits as per the [spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests).

```suggestion
  it("should parse proper timeout", () => {
    expect(grpcParseTimeout("1H")).toEqual(3600000);
    expect(grpcParseTimeout("12345678m")).toEqual(12345678);
  });
```

packages/connect-core/src/grpc-parse-timeout.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+  it("should should return undefined for null value", () => {
+    expect(grpcParseTimeout(null)).toEqual(undefined);
+  });
+  it("should should return a ConnectError for an incorrect value", () => {
+    const result = grpcParseTimeout("HH");
+    expect(result).toBeInstanceOf(ConnectError);
+  });
```
timostamm, line 28: We can be more aggressive here :)

```suggestion
  it("should should return a ConnectError for an incorrect value", () => {
      expect(grpcParseTimeout("HH")).toBeInstanceOf(ConnectError);
      expect(grpcParseTimeout("123456789m")).toBeInstanceOf(ConnectError);
      expect(grpcParseTimeout("12345678")).toBeInstanceOf(ConnectError);
      expect(grpcParseTimeout("1H1H")).toBeInstanceOf(ConnectError);
      expect(grpcParseTimeout("foo")).toBeInstanceOf(ConnectError);
  });
```

packages/connect-node-test/src/crosstest/timeout_on_sleeping_server.spec.ts
```diff
@@ -48,45 +47,47 @@ xdescribe("timeout_on_sleeping_server", function () {
...
+            request,
+            options
+          )) {
+            fail(
+              `expecting no response from sleeping server, got: ${response.toJsonString()}`
+            );
+          }
+          fail("expected to catch an error");
+        } catch (e) {
+          expectError(e);
```
timostamm, line 70: You can simplify this now, and remove the function expectError completely:

```suggestion
          expect(e).toBeInstanceOf(ConnectError);
          expect(connectErrorFromReason(e).code).toBe(Code.DeadlineExceeded);
```

packages/connect-node/src/connect-protocol.ts
```diff
@@ -149,6 +152,35 @@ export function createConnectProtocol(
               ),
               responseTrailer: new Headers(),
             };
+
+            const timeout = connectParseTimeout(
+              requestHeader.get(connectTimeoutHeader) ?? ""
```
timostamm, line 157: No need to nullish coalesce to empty string:

```suggestion
              requestHeader.get(connectTimeoutHeader)
```

packages/connect-node/src/connect-protocol.ts
```diff
@@ -282,6 +317,34 @@ export function createConnectProtocol(
               );
             }
 
+            const timeout = connectParseTimeout(
+              requestHeader.get(connectTimeoutHeader) ?? ""
```
timostamm, line 321: No need to nullish coalesce to empty string.

packages/connect-node/src/connect-protocol.ts
```diff
@@ -422,6 +487,35 @@ export function createConnectProtocol(
               ),
               responseTrailer: new Headers(),
             };
+
+            const timeout = connectParseTimeout(
+              requestHeader.get(connectTimeoutHeader) ?? ""
```
timostamm, line 492: No need to nullish coalesce to empty string.

packages/connect-node/src/grpc-protocol.ts
```diff
@@ -65,23 +67,43 @@ export function createGrpcProtocol(
...
             const context: HandlerContext = {
               method: spec.method,
               service: spec.service,
-              requestHeader: nodeHeaderToWebHeader(req.headers),
+              requestHeader,
               responseHeader: grpcCreateResponseHeader(type.binary),
               responseTrailer: new Headers(),
             };
+            const timeout = grpcParseTimeout(
+              requestHeader.get(grpcTimeoutHeader) ?? ""
```
timostamm, line 91: No need to nullish coalesce to empty string.

packages/connect-node/src/grpc-protocol.ts
```diff
@@ -117,23 +139,45 @@ export function createGrpcProtocol(
...
               method: spec.method,
               service: spec.service,
-              requestHeader: nodeHeaderToWebHeader(req.headers),
+              requestHeader,
               responseHeader: grpcCreateResponseHeader(type.binary),
               responseTrailer: new Headers(),
             };
+
+            const timeout = grpcParseTimeout(
+              requestHeader.get(grpcTimeoutHeader) ?? ""
```
timostamm, line 164: No need to nullish coalesce to empty string.

packages/connect-node/src/grpc-protocol.ts
```diff
@@ -188,23 +232,43 @@ export function createGrpcProtocol(
...
             const context: HandlerContext = {
               method: spec.method,
               service: spec.service,
-              requestHeader: nodeHeaderToWebHeader(req.headers),
+              requestHeader,
               responseHeader: grpcCreateResponseHeader(type.binary),
               responseTrailer: new Headers(),
             };
+            const timeout = grpcParseTimeout(
+              requestHeader.get(grpcTimeoutHeader) ?? ""
```
timostamm, line 256: No need to nullish coalesce to empty string.

packages/connect-node/src/grpc-protocol.ts
```diff
@@ -254,23 +318,44 @@ export function createGrpcProtocol(
...
             const context: HandlerContext = {
               method: spec.method,
               service: spec.service,
               requestHeader: nodeHeaderToWebHeader(req.headers),
               responseHeader: grpcCreateResponseHeader(type.binary),
               responseTrailer: new Headers(),
             };
+
+            const timeout = grpcParseTimeout(
+              requestHeader.get(grpcTimeoutHeader) ?? ""
```
timostamm, line 343: No need to nullish coalesce to empty string.

packages/connect-node/src/grpc-web-protocol.ts
```diff
@@ -216,13 +262,32 @@ export function createGrpcWebProtocol(
...
+
             const context: HandlerContext = {
               method: spec.method,
               service: spec.service,
               requestHeader: nodeHeaderToWebHeader(req.headers),
               responseHeader: grpcWebCreateResponseHeader(type.binary),
               responseTrailer: new Headers(),
             };
+            const timeout = grpcParseTimeout(
+              requestHeader.get(grpcWebTimeoutHeader) ?? ""
```
timostamm, line 274: No need to nullish coalesce to empty string.

packages/connect-node/src/grpc-web-protocol.ts
```diff
@@ -287,13 +354,32 @@ export function createGrpcWebProtocol(
...
             const context: HandlerContext = {
               method: spec.method,
               service: spec.service,
-              requestHeader: nodeHeaderToWebHeader(req.headers),
+              requestHeader,
               responseHeader: grpcWebCreateResponseHeader(type.binary),
               responseTrailer: new Headers(),
             };
+            const timeout = grpcParseTimeout(
+              requestHeader.get(grpcWebTimeoutHeader) ?? ""
```
timostamm, line 366: No need to nullish coalesce to empty string.

timostamm (commented):
I'm seeing quite a few log statements in the tests:

```
protocol handle failed Error [ERR_STREAM_WRITE_AFTER_END]: write after end
```

Is it possible that the `res.setTimeout` callback is triggered, where we "end" the response, but our code still runs on and tries to write to the response?

gilwong00:
> I'm seeing quite a few log statements in the tests:
> 
> ```
> protocol handle failed Error [ERR_STREAM_WRITE_AFTER_END]: write after end
> ```
> 
> Is it possible that the `res.setTimeout` callback is triggered, where we "end" the response, but our code still runs on and tries to write to the response?

@timostamm I thought so too when I initially saw that but I was getting the same errors when I was setting the timeout on the `request` level too so that made me somewhat confident that we werent. But we should probably investigate this more after https://github.com/bufbuild/connect-web/pull/382 lands to see if that resolves the error or I can make a new issues to continue the triage. Thoughts?

timostamm:
> But we should probably investigate this more after #382 lands to see if that resolves the error or I can make a new issues to continue the triage. Thoughts?

Let's land this as is (after comments from last review are resolved).

timostamm (approved):

PR #380, opened 2022-12-20
Make sure that baseUrl stays flexible

timostamm:
I noticed that some users don't use full baseUrls, for example:

```ts
createConnectTransport({
  baseUrl: "/"
})
```

This is valid and useful in a web application. We switched to `URL` in @bufbuild/connect-core, which requires a scheme and host, so we would break existing use cases. 

This reverts this (unreleased) change - `createMethodUrl()` now returns a string.

### DIFF
--- a/packages/connect-node/src/handler.ts
+++ b/packages/connect-node/src/handler.ts
@@ -126,8 +126,7 @@ export function createHandler<M extends MethodInfo>(
   return Object.assign(handleAny, {
     service,
     method,
-    requestPath: createMethodUrl("https://example.com", service, method)
-      .pathname,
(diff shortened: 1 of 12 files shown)


packages/connect-node/src/handler.ts
```diff
@@ -126,8 +126,7 @@ export function createHandler<M extends MethodInfo>(
   return Object.assign(handleAny, {
     service,
     method,
-    requestPath: createMethodUrl("https://example.com", service, method)
-      .pathname,
+    requestPath: createMethodUrl("/", service, method),
```
timostamm, line 129: 🎉

gilwong00 (approved):
LGTM!

PR #382, opened 2022-12-21
Sensibly handle write errors on closed stream

timostamm:
https://github.com/bufbuild/connect-web/pull/365 has revealed an issue with streams. To illustrate, consider the client-side of a client-streaming call:

```ts
await s.send(msg)
s.close()
await s.receive()
```

The server may error on the request for various reasons, for example because the request was sent with an unsupported encoding. In that case, the server responds with an HTTP error code and may close the connection.

Besides such a protocol error, the service implementation may also raise an error. In this case, the response has a HTTP 200 status code (in all three protocols), but the server sends an error in the trailer and may close the connection. Such a behavior of the service implementation is perfectly valid.

Both cases cause a problem in our client, because send() will call write() on a Node.js stream on a closed connection, which raises an ERR_STREAM_WRITE_AFTER_END error, and not the error intended by the service:

```ts
await s.send(msg) // throws ERR_STREAM_WRITE_AFTER_END
s.close()
await s.receive() // throws the error intended by the service
```

Because the promise rejection on line 1 throws an error, the control flow is interrupted, and the user never sees the error intended by the service.

We have gone through multiple options, and settled and send() and close() resolving to a boolean, and adding a property `sendError` to surface the error. This lets the example above surface the error intended by the service.


packages/connect-node/src/private/node-error.ts
```diff
@@ -25,14 +25,29 @@ import {
...
+    // by client-side errors.
+    // This can occur if the server has written a response with an error
+    // and has ended the connection. This response may already sit in a
+    // buffer on the client, while it is still writing to the request
+    // body.
+    // To avoid this problem, we wrap ERR_STREAM_WRITE_AFTER_END as a
+    // ConnectError with Code.Aborted. The special meaning of this code
+    // in this situation is documented in StreamingConn.send() and in
+    // createServerStreamingFn().
+    code = Code.Aborted;
```
timostamm, line 40: We convert ERR_STREAM_WRITE_AFTER_END to a Code.Aborted.

All transports were updated to wrap errors from using this function when writing to the request body.

Promise clients and callback clients special case Code.Aborted in their implementation for server-streaming methods so that `for await (const res of client.foo({})) {...}` has reasonable behavior.

promise-client.spec.ts of connect-core covers the client behavior. The stub transport introduced for that is now also used to add some test coverage for interceptors in connect-core.

gilwong00 (approved):

PR #383, opened 2022-12-22
Add Compression to protocols and transports

gilwong00:
This add compression support for all http and https transports and the grpc and grpc web protocols


packages/connect-node/src/grpc-http-transport.ts
```diff
@@ -191,12 +232,20 @@ export function createGrpcHttpTransport(
               throw "premature eof";
             }
 
+            let responseBody = messageResult.value.data;
+            if (compression) {
+              responseBody = await compression.decompress(
+                responseBody,
+                readMaxBytes
+              );
+            }
```
timostamm, line 241: You have to check the envelope flags. You must only decompress if the compressed flag is set.

I think it's a misnomer to use `responseBody` here. (Connect unary is different because it is not enveloped). Can you rename to `messageData`?

You can change the order a bit to make this more readable:

```ts
            const messageResult = await readEnvelope(response);
            if (messageResult.done) {
              throw "premature eof";
            }
            let messageData = messageResult.value.data;
            if (compressed flag is set) {
              throw error if grpc-encoding not set
              messageData = decompress(messageData)
            }

            const eofResult = await readEnvelope(response);
            if (!eofResult.done) {
              throw "extraneous data";
            }

            const trailer = await trailerPromise;
            grpcValidateTrailer(trailer);
```

packages/connect-node/src/grpc-http2-transport.ts
```diff
@@ -163,33 +212,44 @@ export function createGrpcHttp2Transport(
...
               throw "premature eof";
             }
 
+            let responseBody = messageResult.value.data;
+            if (compression) {
+              responseBody = await compression.decompress(
+                responseBody,
+                readMaxBytes
+              );
+            }
```
timostamm, line 245: See comments on the grpc http transport. I think all points apply here as well, and these lines should be identical between the two transports.

packages/connect-node/src/grpc-http-transport.ts
gilwong00: hey @timostamm  i tried to follow the ordering you wrote above, however, when we move the `messageResult.done` check below reading of the envelope we fail a lot of test. Would you be ok if i went with this approach?

```ts
            const messageResult = await readEnvelope(response);
            const eofResult = await readEnvelope(response);
            if (!eofResult.done) {
              throw "extraneous data";
            }

            const trailer = await trailerPromise;
            grpcValidateTrailer(trailer);

            if (messageResult.done) {
              throw "premature eof";
            }
            let messageData = messageResult.value.data;
            if (
              (messageResult.value.flags & compressedFlag) ===
              compressedFlag
            ) {
              if (!compression) {
                throw new ConnectError(
                  `received compressed envelope, but no grpc-encoding`,
                  Code.InvalidArgument
                );
              }
              messageData = await compression.decompress(
                messageData,
                readMaxBytes
              );
            }
```

timostamm: Ah, of course we need to check for an error in the trailer first. 
It's not ideal, but yes, let's go with the order you got for now.

packages/connect-node/src/grpc-http-transport.ts
```diff
@@ -148,16 +171,34 @@ export function createGrpcHttpTransport(
...
             message: normalize(message),
             signal: signal ?? new AbortController().signal,
           },
           async (req: UnaryRequest<I>): Promise<UnaryResponse<O>> => {
-            const envelope = encodeEnvelope(
-              messageFlag,
-              serialize(req.message)
-            );
             const encoding = useBinaryFormat ? "binary" : "utf8";
+            let flag = messageFlag;
```
timostamm, line 187: The line above is:

```ts
const encoding = useBinaryFormat ? "binary" : "utf8";
```

We pass the encoding to `makeNodeRequest()` and method `write()` of `http.ClientRequest`.

But with compression, we only ever write Uint8Array. Please remove the variable and code in `makeNodeRequest` completely from all transports.

packages/connect-node/src/grpc-web-http2-transport.ts
```diff
@@ -161,24 +200,43 @@ export function createGrpcWebHttp2Transport(
...
+                messageOrTrailerData,
+                readMaxBytes
+              );
+            }
 
-            if (messageOrTrailerResult.value.flags === grpcWebTrailerFlag) {
+            if (
+              (messageOrTrailerResult.value.flags & grpcWebTrailerFlag) ===
+              grpcWebTrailerFlag
+            ) {
```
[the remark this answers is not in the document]
gilwong00: good catch!

timostamm (approved):
LGTM!

PR #384, opened 2022-12-29
Add subpath export structure

smaye81:
This adds a subpath export structure with a fallback for bundlers that do not support it (such as Parcel).


timostamm (commented):
If this allows consistent use of (for example) `import {connectCodeFromHttpStatus} from "@bufbuild/connect-core/protocol-connect"`, this is great news. The couple of additional .js files seem to be worth it :+1:

I wonder how we can make sure that we don't accidentally break one of those files without noticing.

timostamm (approved):

PR #390, opened 2023-01-03
Add unresolvable host support for Http Transports

gilwong00:
This adjust how to create the node request for http and https transports to reject when calling an unresolvable host and updates the unresolvable host test


packages/connect-node/src/connect-http-transport.ts
```diff
@@ -446,9 +443,25 @@ function makeNodeRequest(options: NodeRequestOptions) {
...
+      path: endpoint.pathname,
+      signal: req.signal,
+      ...httpOptions,
+    });
+    request.on("socket", (socket) => {
+      socket.on("connect", () => {
+        resolve(request);
+      });
+      socket.on("error", reject);
+    });
```
timostamm, line 465: From what I can see from the [documentation](https://nodejs.org/api/http.html#class-httpagent), `http` opens a new socket for every request by default. But it is likely that users may want to provide their own Agent and keep a socket alive across multiple requests. 

We have to support this use case, but we don't have to do it in this PR. Can you take a note?

In this PR, we have to make sure we don't leave too many event listeners attached:
- use `once("socket", ...)`
- for "connect" and "error", `once()` should also be fine, but we have to detach the "connect" listener on "error", and we have to detach the "error" listener on "connect"

This way, there should be any code path where we leave any event listeners attached we don't need.

packages/connect-node/src/grpc-http-transport.ts
```diff
@@ -437,9 +433,25 @@ function makeNodeRequest(options: NodeRequestOptions) {
   });
 }
 
-function nodeRequest(protocol: string) {
-  if (protocol.startsWith("http") || protocol.startsWith("https")) {
-    return protocol.includes("https") ? https.request : http.request;
-  }
-  throw new Error("Unsupported protocol");
+function getNodeRequest(
```
timostamm, line 436: I think it's fair to deduplicate this function now, and move it to `src/private/node-request.ts` for example 🙂

packages/connect-node/src/grpc-web-http2-transport.ts
```diff
@@ -443,7 +443,7 @@ export function createGrpcWebHttp2Transport(
                 }
               },
             };
-            return Promise.resolve(conn);
+            return conn;
```
timostamm, line 446: TY!

packages/connect-node/src/grpc-http-transport.ts
gilwong00: moved all node request functions to `node-request.ts` in private and replaced all duplicated code

timostamm (approved):

PR #399, opened 2023-01-09
Update plugin docs

smaye81:
This updates the plugin documentation to list all possible options available when generating code.


timostamm (commented):
Thanks for the update!

It looks like we have another small mistake [here](https://github.com/bufbuild/protobuf-es/blob/382e30cdd311cac302436fdddb9ac1d5c77fe7e5/packages/protoc-gen-es/README.md?plain=1#L88-L91). Could you update from `_pb` to `_connectweb`?

timostamm (approved):

PR #400, opened 2023-01-09
Switch to curated plugins

smaye81:
Switches the `buf.gen.yaml` in connect-web-bench to use curated plugins instead of the deprecated remote plugin syntax.


packages/connect-web-bench/buf.gen.yaml
```diff
@@ -6,9 +6,9 @@ plugins:
   - name: connect-web
     out: src/gen/connectweb
     opt: ts_nocheck=false,target=ts
-  - remote: buf.build/protocolbuffers/plugins/js:v3.20.1-1
+  - plugin: buf.build/protocolbuffers/js
```
timostamm, line 9: We still need to pin the version though.

smaye81: Ah shoot.  We don't do this in protobuf-es either.  I'll update it there also.

timostamm (approved):

PR #404, opened 2023-01-10
Add interface Serialization

timostamm:
This adds the following interface, similar to [interface Compression](https://github.com/bufbuild/connect-web/blob/1bc26cae5c95e3f41fd63dfa90707f207176b595/packages/connect-node/src/compression.ts#L20-L24):

```ts
interface Serialization<T> {
  serialize: (data: T) => Uint8Array;
  parse: (data: Uint8Array) => T;
}
```

This abstraction will make it a bit more straight-forward to use stream transforms.


packages/connect-core/src/serialization.ts
```diff
@@ -0,0 +1,138 @@
...
+      }
+    },
+  };
+}
+
+/**
+ * Returns functions to normalize and serialize the input message
+ * of an RPC, and to parse the output message of an RPC.
+ */
+export function createClientMethodSerializers<
```
timostamm, line 119: It still makes sense to keep this convenience-function. We need to update [createServerMethodsSerializers](https://github.com/bufbuild/connect-web/blob/1bc26cae5c95e3f41fd63dfa90707f207176b595/packages/connect-node/src/private/create-server-method-serializers.ts) as well.

packages/connect-core/src/serialization.ts
```diff
@@ -0,0 +1,138 @@
...
+): Serialization<T> {
+  const textEncoder = options?.textEncoder ?? new TextEncoder();
+  const textDecoder = options?.textDecoder ?? new TextDecoder();
+  return {
+    parse(data: Uint8Array): T {
+      try {
+        const json = textDecoder.decode(data);
+        return messageType.fromJsonString(json, options);
+      } catch (e) {
+        const m = e instanceof Error ? e.message : String(e);
```
smaye81, line 99: Is it worth having like a `connectErrorFromError` utility function somewhere?  I've noticed this in a few places in `connect-web`.

packages/connect-core/src/serialization.ts
```diff
@@ -0,0 +1,138 @@
...
+  PartialMessage,
+} from "@bufbuild/protobuf";
+import { ConnectError } from "./connect-error.js";
+import { Code } from "./code.js";
+
+/**
+ * Serialization provides methods to serialize or parse data with a certain
+ * format.
+ */
+export interface Serialization<T> {
```
smaye81, line 32: Dumb comment, ignore me if you want, but I feel like `Serializable` makes more sense as the interface name.  Maybe I'm still hung up on Java's Serializable interface though.

smaye81 (approved):

packages/connect-core/src/serialization.ts
timostamm: Good should, we should use `connectErrorFromReason()` here. Updated.

packages/connect-core/src/serialization.ts
timostamm: An object that implements `interface Serializable` is serializable. But this isn't used by objects that are serializable, it's just an object that has two methods `parse()` and `serialize()`, so I don't think `Serializable` fits. 

Not a huge fan of the name though and open to suggestions.

timostamm: Going to merge this because there's some work in the pipeline based on this. Let's discuss offline.

PR #405, opened 2023-01-10
Remove protocol prefixes from code

smaye81:
This removes the superfluous protocol prefixes from filenames and exported symbols.  For example, files named `connect-foo.ts` inside the `protocol-connect` directory have been renamed to simply `foo.ts`.

Likewise, for exported functions such as `connectFoo`, these have been renamed to `foo`.

An effort was also made to consolidate some files.  For example, `connect-code-to-http-status.ts` and `connect-code-from-http-status.ts` simply became `http-status.ts` with the two functions (to/from) put inside.  We could consolidate further but in doing so, I realized that it may not be worth the trouble since users will just be importing from the `protocol-xxx` directory and filenames won't necessarily matter externally.  Open to suggestions though.


gilwong00 (approved):
LGTM

Makefile
```diff
@@ -47,7 +47,7 @@ $(BUILD)/protoc-gen-connect-web: node_modules tsconfig.base.json packages/protoc
 	@mkdir -p $(@D)
 	@touch $(@)
 
-$(BUILD)/connect-core: $(GEN)/connect-core node_modules tsconfig.base.json packages/connect-core/tsconfig.json $(shell find packages/connect-core/src -name '*.ts')
+$(BUILD)/connect-core: $(GEN)/connect-core node_modules tsconfig.base.json packages/connect-core/tsconfig.json $(shell find packages/connect-core/src -name '*.ts') $(shell find packages/connect-core -name 'protocol-*.js')
```
timostamm, line 50: I'm a bit worried this is going to break if we ever add a `protocol-*.ts` file in `packages/connect-core/src` - it will be transpiled to `packages/connect-core/dist`, and then `find` will find it.

What do you think about using `packages/connect-core/*.js` as a prerequisite instead?

smaye81: Works for me.  Changed.

timostamm (approved):
> @timostamm I made the generation of the status.proto pretty specific to the protocol-grpc directory. 

Makes sense to me 👍 

Can you delete this line?

https://github.com/bufbuild/connect-web/blob/20643e24bf0700576f80cf3b686434098544a2c9/packages/connect-core/.npmignore#L6

PR #407, opened 2023-01-11
Add a cause property to ConnectError

smaye81:
This adds a `cause` property to `ConnectError`, which will allow specifying the underlying error cause in cases where the actual error is elided from the message.


packages/connect-core/src/connect-error.ts
```diff
@@ -74,9 +81,23 @@ export class ConnectError extends Error {
...
+    metadata?: HeadersInit,
+    outgoingDetails?: Message[],
+    cause?: unknown
+  );
+  constructor(
+    message: string,
+    code: Code = Code.Unknown,
+    metadata?: HeadersInit,
+    outgoingDetails?: Message[],
+    cause?: unknown
```
timostamm, line 100: Is it necessary to overload the signature? I would have expected we'd just do:

```diff
constructor(
    message: string,
    code: Code = Code.Unknown,
    metadata?: HeadersInit,
    outgoingDetails?: Message[],
+   cause?: unknown
```

packages/connect-core/src/connect-error.ts
```diff
@@ -65,6 +65,13 @@ export class ConnectError extends Error {
 
   override name = "ConnectError";
 
+  /**
+   * The underlying cause of this error.  In cases where the actual cause is
+   * elided with the error message, the cause is specified here so that we don't
+   * leak the underlying error, but instead make it available for logging.
+   */
+  cause: unknown;
```
timostamm, line 73: Can you make this a `unknown | undefined`, so that it is obvious that it is optional?

packages/connect-node/src/compression.ts
```diff
@@ -81,12 +81,22 @@ function wrapZLibErrors<T>(
...
       case "Z_DATA_ERROR":
       case "ERR_PADDING_2":
-        // TODO(TCN-785) we may want to preserve the cause, but not in the connect error message
-        e = new ConnectError("decompression failed", Code.InvalidArgument);
+        e = new ConnectError(
+          "decompression failed",
+          Code.InvalidArgument,
+          undefined,
+          undefined,
+          code
```
timostamm, line 89: We want to maintain a chain of errors, and have to pass the unmodified error as the cause here.

This makes the following piece of code continue to work, whether we wrapped the error or not:

```ts
const isNodeEOF = unwrapNodeErrorChain(err)
  .map(getNodeErrorProps)
  .some((p) => p.code == "ERR_STREAM_WRITE_AFTER_END")
```

packages/connect-node/src/compression.ts
```diff
@@ -81,12 +81,22 @@ function wrapZLibErrors<T>(
...
         break;
       default:
-        // TODO(TCN-785) we may want to preserve the cause, but not in the connect error message
-        e = new ConnectError("decompression failed", Code.Internal);
+        e = new ConnectError(
+          "decompression failed",
+          Code.Internal,
+          undefined,
+          undefined,
+          code
```
timostamm, line 98: Same here, have to pass the unmodified error.

packages/connect-node/src/connect-protocol.ts
```diff
@@ -230,9 +230,13 @@ export function createConnectProtocol(
...
               } else {
-                // TODO(TCN-785) We want to elide the error message for the client, but still
-                //    make it available for logging. We may have to add a "cause" property.
-                ce = new ConnectError("internal error", Code.Internal);
+                ce = new ConnectError(
+                  "internal error",
+                  Code.Internal,
+                  undefined,
+                  undefined,
+                  (e as Error).message
```
timostamm, line 238: Same here.

packages/connect-node/src/private/validate-read-max-bytes-option.ts
```diff
@@ -29,7 +29,10 @@ export function validateReadMaxBytesOption(
     );
   }
   if (readMaxBytes > maxReadMaxBytes) {
-    throw new ConnectError("", Code.Internal);
+    throw new ConnectError(
+      `readMaxBytes ${readMaxBytes} is greater than the maximum allowable size.`,
+      Code.Internal
+    );
```
timostamm, line 35: TY!

packages/connect-core/src/connect-error.ts
smaye81: Do you think it makes more sense to make this a union type of like `Error | undefined | blah`?

timostamm: Unknown is the best choice. In current TypeScript, `e` in `try{}catch(e){}` is `unknown`.

packages/connect-core/src/connect-error.ts
smaye81: Ugh yeah.  Fixed this.  Think my earlier efforts at overloading confused the matter.

timostamm (approved):

gilwong00 (approved):

PR #408, opened 2023-01-11
Add WHATWG transform streams

timostamm:
This adds transform streams for compression, serialization, and enveloping to @bufbuild/connect-core. We will use this feature to streamline the handler and client implementations in @bufbuild/connect-node.

### DIFF
--- a/packages/connect-core/src/transform-stream-story.spec.ts
+++ b/packages/connect-core/src/transform-stream-story.spec.ts
@@ -1,0 +1,324 @@
+// Copyright 2021-2023 Buf Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {
+  createReadableByteStream,
+  createReadableStream,
+  node16WhatwgStreamPolyfill,
+  readAll,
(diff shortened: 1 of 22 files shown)


packages/connect-core/src/transform-stream-story.spec.ts
```diff
@@ -0,0 +1,324 @@
...
+import { ConnectError } from "./connect-error.js";
+import { Code } from "./code.js";
+
+node16WhatwgStreamPolyfill();
+
+// These tests aim to model the usage of transform streams in clients and servers.
+// Note that the tests were written as a proof of concept, and the coverage is
+// incomplete (cancellation, backpressure, error handling, etc.).
+describe("full story", function () {
+  const readMaxBytes = 0xffffffff; // zlib caps maxOutputLength at this value
```
smaye81, line 42: Is it worth reading the max bytes value from here?:  packages/connect-node/src/private/validate-read-max-bytes-option.ts

packages/connect-core/src/transform-stream.ts
```diff
@@ -0,0 +1,320 @@
...
+    transform(chunk, controller) {
+      let data: Uint8Array;
+      let flags = 0;
+      if (chunk.end) {
+        flags = flags | endStreamFlag;
+        try {
+          data = endSerialization.serialize(chunk.value);
+        } catch (e) {
+          return controller.error(
+            new ConnectError("failed to serialize end", Code.Internal)
```
smaye81, line 55: Should we specify `e` as the `cause` here (and a bunch of places below)?

smaye81 (approved):

packages/connect-core/src/transform-stream-story.spec.ts
timostamm: We can't, this is in connect-core.

packages/connect-core/src/transform-stream.ts
timostamm: Yes, good shout. Updated.

gilwong00 (approved):

PR #411, opened 2023-01-11
Move all header names to connect-core

timostamm:
This moves all header names to connect-core, to their respective protocol-X directory.

### DIFF
--- a/packages/connect-node/src/connect-http-transport.ts
+++ b/packages/connect-node/src/connect-http-transport.ts
@@ -69,8 +69,6 @@ import {
 } from "./connect-http2-transport.js";
 import { getNodeRequest, makeNodeRequest } from "./private/node-request.js";
 
-const messageFlag = 0b00000000;
-
 export interface ConnectHttpTransportOptions {
   /**
    * Base URI for all HTTP requests.
@@ -109,7 +107,9 @@ export interface ConnectHttpTransportOptions {
   /**
    * Options for the http request.
    */
-  httpOptions?: http.RequestOptions | https.RequestOptions;
+  httpOptions?:
+    | Omit<http.RequestOptions, "signal">
+    | Omit<https.RequestOptions, "signal">;
 
   // TODO document
   acceptCompression?: Compression[];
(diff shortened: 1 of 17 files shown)


packages/connect-node/src/connect-http-transport.ts
```diff
@@ -109,7 +107,9 @@ export interface ConnectHttpTransportOptions {
   /**
    * Options for the http request.
    */
-  httpOptions?: http.RequestOptions | https.RequestOptions;
+  httpOptions?:
+    | Omit<http.RequestOptions, "signal">
+    | Omit<https.RequestOptions, "signal">;
```
timostamm, line 112: We should not accept an AbortSignal in the request options because we already accept an AbortSignal in CallOptions.

packages/connect-node/src/connect-http-transport.ts
```diff
@@ -306,7 +306,7 @@ export function createConnectHttpTransport(
                     "cannot send, stream is already closed"
                   );
                 }
-                let flags = messageFlag;
+                let flags = 0;
```
timostamm, line 309: I think a straight-forward `0` is the better choice, now that we have moved all other flag constants to a proper place. It doesn't seem worth it to add a class or constructor function for EnvelopedMessage.

packages/connect-node/src/grpc-protocol.ts
```diff
@@ -270,21 +268,6 @@ export function createGrpcProtocol(
...
-                res,
-                context,
-                new ConnectError(
-                  `Unexpected input flags ${inputResult.value.flags.toString(
-                    2
-                  )}`,
-                  Code.Internal
-                )
-              );
-            }
```
timostamm, line 287: I was wondering about this for a while and didn't see the point. I think I introduced it 🙈 😆

packages/connect-node/src/private/node-request.ts
```diff
@@ -47,11 +45,11 @@ export function makeNodeRequest(options: NodeRequestOptions) {
     const endpoint = new URL(options.req.url);
     const nodeRequestFn = nodeRequest(endpoint.protocol);
     const request = nodeRequestFn(options.req.url, {
+      ...options.httpOptions,
```
timostamm, line 48: This is crucial. While we omit "signal" from the type, it is still possible to pass options with a "signal" property to this. So we have to make sure we override it.

gilwong00 (approved):

PR #413, opened 2023-01-11
Move compression header and validate functions into core

gilwong00:
This moves all `createRequestHeaderWithCompression` and `validateResponseWithCompression` into `connect-core`


packages/connect-core/src/protocol-grpc/create-request-header.ts
```diff
@@ -36,3 +38,32 @@ export function createRequestHeader(
...
+/**
+ * Creates headers for a gRPC request with compression.
+ */
+export function createRequestHeaderWithCompression(
+  methodKind: MethodKind,
+  useBinaryFormat: boolean,
+  timeoutMs: number | undefined,
+  userProvidedHeaders: HeadersInit | undefined,
+  acceptCompression: string[],
+  sendCompression: string | undefined
```
timostamm, line 51: Let's use `interface Compression` here.

packages/connect-core/src/protocol-grpc/create-request-header.ts
```diff
@@ -36,3 +38,32 @@ export function createRequestHeader(
...
+  const result = createRequestHeader(
+    useBinaryFormat,
+    timeoutMs,
+    userProvidedHeaders
+  );
+  let acceptEncodingField = "Accept-Encoding";
+  if (methodKind !== MethodKind.Unary) {
+    acceptEncodingField = "GRPC-" + acceptEncodingField;
+    if (sendCompression != undefined) {
+      result.set("Grpc-Encoding", sendCompression);
```
timostamm, line 62: Can you use a const from `./headers.js` here as well? Same for other string literals in this file. If a const does not exist yet, please add it. Let's duplicate definitions in `protocol-grpc/headers.ts` and `protocol-grpc-web/headers.ts`, so that transports and handlers only have to import from one protocol.

packages/connect-core/src/protocol-grpc/validate-response.ts
```diff
@@ -49,3 +50,28 @@ export function validateResponse(
...
+export function validateResponseWithCompression(
+  useBinaryFormat: boolean,
+  acceptCompression: Compression[],
+  status: number,
+  headers: Headers
+): { compression: Compression | undefined } {
+  validateResponse(useBinaryFormat, status, headers);
+
+  let compression: Compression | undefined;
+  const encodingField = "Grpc-Encoding";
```
timostamm, line 63: Same here, let's use the const names.

packages/connect-node/src/grpc-web-http2-transport.ts
```diff
@@ -451,53 +450,27 @@ export function createGrpcWebHttp2Transport(
...
-  headers: Headers
-): { compression: Compression | undefined } {
-  validateResponse(useBinaryFormat, status, headers);
+// export function grpcWebValidateResponseWithCompression(
+//   useBinaryFormat: boolean,
+//   acceptCompression: Compression[],
+//   status: number,
+//   headers: Headers
+// ): { compression: Compression | undefined } {
+//   validateResponse(useBinaryFormat, status, headers);
```
timostamm, line 459: Can you remove this?

timostamm (commented):
Love the clean-up! 

I think we should use `interface Compression` and replace the header name string literals with the const declarations added in https://github.com/bufbuild/connect-web/pull/411.

The functionality of these functions has grown a bit. What do you thing about adding some basic tests?

gilwong00:
> Love the clean-up!
> 
> I think we should use `interface Compression` and replace the header name string literals with the const declarations added in #411.
> 
> The functionality of these functions has grown a bit. What do you thing about adding some basic tests?

Yeah I think some basic test makes sense here. I can add them to this PR

packages/connect-core/src/protocol-connect/create-request-header.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+  });
+});
+
+describe("createRequestHeaderWithCompression", () => {
+  const compressionMock: Compression = {
+    name: "gzip",
+    compress: (bytes: Uint8Array) => Promise.resolve(bytes),
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    decompress: (bytes: Uint8Array, _: number) => Promise.resolve(bytes),
+  };
```
timostamm, line 54: 👍 

In case you want more mock instances (for `acceptCompression: Compression[]`), you could also use 

```ts
const compressionMock = { name: "gzip" } as Compression;
const mocks = [{ name: "alpha" }, { name: "beta" }] as Compression[];
```

packages/connect-core/src/protocol-connect/create-request-header.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+  it("should create request headers", () => {
+    const headers = createRequestHeader(
+      MethodKind.Unary,
+      true,
+      undefined,
+      undefined
+    );
+    expect(headers instanceof Headers).toBe(true);
+    expect(headers.get("Content-Type")).toBe("application/proto");
+    expect(headers.get("Connect-Protocol-Version")).toBe("1");
```
timostamm, line 38: Can you use [listHeaderKeys](https://github.com/bufbuild/connect-web/pull/415#discussion_r1068574063) to verify that we don't create any additional header fields? For example:

```ts
expect(listHeaderKeys(headers)).toBe(["content-type", "connect-protocol-version"]);
```

Please add it to all cases.

timostamm (commented):
The tests are great, TY! Just a couple of comments below.

packages/connect-core/src/protocol-connect/create-request-header.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+
+describe("createRequestHeader", () => {
+  it("should create request headers", () => {
+    const headers = createRequestHeader(
+      MethodKind.Unary,
+      true,
+      undefined,
+      undefined
+    );
+    expect(headers instanceof Headers).toBe(true);
```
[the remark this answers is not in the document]
gilwong00: it does, i was just being explicit but i can remove these

packages/connect-core/src/protocol-connect/create-request-header.spec.ts
gilwong00: sure thing

timostamm (approved):

jlumsden-mts:
This looks great @gilwong00, are there any plans to get this functionality into connect-web as well as connect-node?

timostamm:
@jlumsden-mts, this is available in `@bufbuild/connect-node`. Handlers and clients support gzip and brotli by default. To compress a request, set `sendCompression`:

```ts
import { compressionGzip, createConnectTransport } from "@bufbuild/connect-node";

const transport = createConnectTransport({
  baseUrl: "https://demo.connect.build",
  httpVersion: "2",
  sendCompression: compressionGzip,
});
```

In the web-browser, we are a bit more limited, because `Accept-Encoding` is a [forbidden header name](https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name). I think it is still feasible to implement compression, and it may be beneficial depending on your payload, but we currently do not, because supporting it would increase bundle size. 

Do you have a specific use case in mind for web?

jlumsden-mts:
Thanks for the quick response @timostamm ,

We have a system that uses connect-web on the frontend and gRPC-Web (on ASP.NET Core) for the backend and this works well but some of our response payloads can be relatively large (megabytes) and are quite compressible (2-4x using gzip) from my testing so it would be nice to support this.

It looks like the Grpc-Accept-Encoding header is supported on the backend and this doesn't appear to be forbidden from my reading. In addition, https://developer.mozilla.org/en-US/docs/Web/API/Compression_Streams_API provides a baked in way of doing gzip at least, without increasing bundle size, but sadly not supported by Firefox or Safari yet. Separately, the Brotli.js library does look like it is quite popular: https://www.npmjs.com/package/brotli

jlumsden-mts:
@timostamm I just checked the headers on the existing requests (from connect-web) and noted they already have `Accept-Encoding: gzip, deflate, br` set (at least on Chrome) but not Grpc-Accept-Encoding so is it just a matter of aligning these two headers? Perhaps I am misunderstanding

timostamm:
That's correct, @jlumsden-mts, pretty much every vendor sends a sensible `Accept-Encoding`, and will decompress a response body with `Content-Encoding` transparently (i.e. JS sees the _decompressed_ body only). 

But thinking about this for a minute, I think it might be an issue that we can't turn this request header off in the browser: A middlebox might see the `Accept-Encoding` header, and apply compression to the response body, even though the individual messages it contains may already be compressed via `Grpc-Encoding`. I think we need to be cautious here and do a bit more research to see what's possible.

On the other hand, you can make use of this fact. Have you tried adding [compression middleware](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.builder.responsecompressionbuilderextensions.useresponsecompression?view=aspnetcore-7.0)? I think this should give you exactly what you need for unary responses.

jlumsden-mts:
Thanks @timostamm that put me on the right track, I have been able to configure the ResponseCompression middleware, setting it up with "application/grpc" as one of the MIME types seems to work well. 

It is applying automatically for both unary and streaming responses but it seems to work okay for both given some preliminary testing. There appears to be some flushing logic which ensures the messages come out right away - I was a bit concerned about this for the streaming responses - although I don't imagine the compression will be much help when streaming unless the individual messages are large.

PR #414, opened 2023-01-12
Fix compression for unary methods in the gRPC and the gRPC web client

timostamm:
This fixes bugs in `grpcWebCreateRequestHeaderWithCompression()` and `grpcCreateRequestHeaderWithCompression()`. 

It makes uses of the header name consts added in https://github.com/bufbuild/connect-web/pull/411.


smaye81 (approved):

timostamm (commented):
Forgot to submit my self-review.

PR #418, opened 2023-01-13
Add Express to example

smaye81:
This augments the current `example/` package to use `Connect-Node` and `Express` as the server rather than the Go demo at connect.build.


gilwong00 (approved):
Example looks good to me!

packages/example/README.md
```diff
@@ -1,34 +1,38 @@
...
-It is a web interface for [ELIZA](https://en.wikipedia.org/wiki/ELIZA), a very
-simple natural language processor built in the 1960s to represent a 
-psychotherapist. The ELIZA service is provided by us at [demo.connect.build](https://demo.connect.build).
-You can find the protocol buffer schema [on the BSR](https://buf.build/bufbuild/eliza/tree/main:buf/connect/demo/eliza/v1/eliza.proto).
+This directory contains a simple example for running a Connect-Web client with a Connect-Node server using [Express](https://expressjs.com/).
+
+The application is a web interface for a stripped-down version of [ELIZA](https://en.wikipedia.org/wiki/ELIZA), a very
+simple natural language processor built in the 1960s to represent a psychotherapist. 
 
+You can find the protocol buffer schema [on the BSR](https://buf.build/bufbuild/eliza/tree/main:buf/connect/demo/eliza/v1/eliza.proto).
```
timostamm, line 8: Can we keep the local eliza.proto? Generating from a remote module is super convenient, but I think the ability to modify the proto file and see how the generated code changes overweighs here.

smaye81: Sure.  Added back.

timostamm:
@smaye81 and I discussed merge order: We decided to wait a couple of days, so that we merge it shortly before the first public beta of @bufbuild/connect-node.

package-lock.json
```diff
@@ -631,6 +631,25 @@
       "resolved": "https://registry.npmjs.org/@socket.io/component-emitter/-/component-emitter-3.1.0.tgz",
       "integrity": "sha512-+9jVqKhRSpsc591z5vX+X5Yyw+he/HCB4iQ/RYxw35CEPaY1gnsNE43nf9n9AaYjAQrTiI/mOwKUKdUs9vf7Xg=="
     },
+    "node_modules/@types/body-parser": {
```
bufdev, line 634: Does this affect dependencies that people take on if they depend on connect-web? If our examples are affecting the size of our libraries, we should move examples to another repo. I legitimately do not know how this ecosystem works, so wanted to ask

timostamm: Absolutely not. 

This lock file includes packages required for tests and benchmarks of the entire monorepo, but it's not used when you install Connect-Web. Connect-Web only needs @bufbuild/protobuf. This [package.json](https://github.com/bufbuild/connect-web/blob/main/packages/connect-web/package.json#L26)) is the authority for that.

bufdev: OK great, just making sure

timostamm:
@smaye81, pushed up https://github.com/bufbuild/connect-web/pull/418/commits/67ccd8bb0d583bb2f52aff2232d6956bf9817c41. 
- fixes a lint issue (our config needs all sources in a `src` directory...)
- serves the web client and the API from the same server, so you don't need two terminal windows
- adds a terminal client (so you do need a second terminal window after all 😆)

It's a bit unfortunate that we have to use H1 for Express here, and use the gRPC protocol over H1, which isn't widely supported. But IMO we should merge this, and revisit once we have a better idea.

packages/example/src/client.ts
```diff
@@ -0,0 +1,44 @@
...
+  createPromiseClient,
+} from "@bufbuild/connect-node";
+import { ElizaService } from "./gen/eliza_connectweb.js";
+import { stdin, stdout } from "process";
+import * as readline from "node:readline/promises";
+
+const rl = readline.createInterface(stdin, stdout);
+
+void (async () => {
+  const transport = createGrpcTransport({
```
smaye81, line 26: @timostamm Did you mean to use a `GrpcTransport` here?

timostamm: Yes, to show that the same handler servers multiple protocols. I'm not super happy about it because of H1 - feel free to change it.

smaye81: I moved it back to Connect just for consistency and because of the usage of HTTP/1.1.  Kind of felt confusing being there.

packages/example/README.md
```diff
@@ -1,49 +1,75 @@
...
+The file `src/client.ts` implements a CLI client that you can run in Node.js. 
 
-Of course, you can use `protoc` as well:
+```shell
+$ npm run client
+```
+
+```
+What is your name?
+> John
```
timostamm, line 44: The screenshot above uses "Donald" 😆

timostamm (approved):

packages/example/README.md
smaye81: Ugh.  Fixed.  Also made some fixes to the Makefile also that I noticed during build.

Makefile
```diff
@@ -115,8 +115,8 @@ $(GEN)/connect-web-bench: node_modules/.bin/protoc-gen-es $(BUILD)/protoc-gen-co
 	@touch $(@)
 
 $(GEN)/example: node_modules/.bin/protoc-gen-es $(BUILD)/protoc-gen-connect-web packages/example/buf.gen.yaml $(shell find packages/example -name '*.proto')
-	rm -rf packages/example/src/*_pb.ts packages/example/src/*_connectweb.ts
-	npm run -w packages/example generate
+	rm -rf packages/example/src/gen/*_pb.ts packages/example/src/gen/*_connectweb.ts
```
timostamm, line 118:
```suggestion
	rm -rf packages/example/src/gen/*
```

timostamm (approved):

PR #424, opened 2023-01-19
Add fetch instance option

### LINKED ISSUE #403: Allow customizing fetch
I'd like to make it easier to integrate connect-web into sveltekit but the way sveltekit provides a custom [fetch](https://kit.svelte.dev/docs/load#universal-vs-server-input) is not compatible with connect-web. 

I think it would be sufficient to add an optional param when creating a client to take a custom fetch and this fetch will be used instead of the native fetch API. Or possibly on each request, which is probably more ergonomic.

There may be a way to extract the headers used and just build a custom interceptor but I couldn't find anything in the docs. Perhaps just accessing cookies would be enough but also not very ergonomic and it still wouldn't solve the hydration problem for SSR.

chanced:
This fixes #403  by adding `fetch` as an option to `GrpcWebTransportOptions` so that users of SvelteKit gain various benefits from providing a fetch instance ([docs](https://kit.svelte.dev/docs/load#making-fetch-requests)).

My prettier config mangled the formatting. Is there anyway you could please share your prettier config so that I can align it?

_edit_:

This was the easiest path forward. Arguably, it would be slightly better performance wise to add fetch as an optional parameter to the returned functions. However, doing so would require a bit more changes, including the generators.

I'm willing to make those changes if that's the path you'd prefer.


chanced:
I'll add that this was the easiest path forward. Arguably, it would be slightly better performance wise to add fetch as an optional parameter to the returned functions. However, doing so would require a bit more changes, including the generator.

I'm willing to make those changes if that's the path you'd prefer.

chanced:
Since the formatting is making diff tracking difficult, this is what was done:

```ts
// added
const FETCH = fetch;
export interface ConnectTransportOptions {
    // ...other options

    // added:
    /**
     * The fetch instance to use. Defaults to the global fetch.
     */
    fetch?: typeof fetch;
}

export function createConnectTransport(
    options: ConnectTransportOptions
): Transport {
    assertFetchApi();
    const fetch = options.fetch ?? FETCH;
   // ... remaining code
}
```

chanced:
One issue that I just realized is that the check for fetch (`assertFetchApi`) isn't run until the `createConnectTransport` is called.

chanced:
Updated by adding checks:

```ts
const FETCH = typeof fetch !== "undefined" ? fetch : undefined;

export function createConnectTransport(
    options: ConnectTransportOptions
): Transport {
    assertFetchApi();
    const fetch = options.fetch ?? FETCH;
    if (!fetch) {
        throw new Error("fetch is not available");
    }
    
   // ... remaining code
}

**Comment by timostamm** [comment]
Apologies for the late reply, @chanced. It looks like an option to provide `fetch` in the transport is the best way forward. 
Having to create a new transport instance for every server side render is not ideal, but I don't think it matters much in the real world.

The pattern could be to define a function for creating clients with the desired transport at a single place in the app:

```ts
export function createClient<T extends ServiceType>(service: T, fetch?: FetchFn): PromiseClient<T> {
  const t = createConnectTransport({baseUrl: "/", fetch});
  return createPromiseClient(service, t);
}
```

And than use it in the SvelteKit `load` functions:

```ts
export async function load({ fetch, params }) {
  const c = createClient(SomeService, fetch);
  return await c.someMethod({id: params.id});
}
```

Some initial testing by @paul-sachs looked promising, but there are issues with binary data and we did not look into server-streaming yet. Supporting SSR via a custom fetch sounds great, but we need more investigation and also integration tests to make sure it works properly.

timostamm:
Superseded by https://github.com/bufbuild/connect-es/pull/618. 

Thank you for the contribution, @chanced!

PR #425, opened 2023-01-19
Add reader and writer for unit testing

smaye81:
This adds a reader and writer implementation for unit testing the AsyncIterable stream approach from a client standpoint.


packages/connect-core/src/transform-iterable-story.spec.ts
```diff
@@ -33,6 +33,64 @@ import {
...
+        for await (const item of this.it) {
+          return { value: item };
+        }
+        return { value: "", done: true };
+      },
+    };
+  }
+}
+
+class Writer<T> {
```
timostamm, line 54: The Write is the difficult part for sure, but this looks pretty neat!

packages/connect-core/src/transform-iterable-story.spec.ts
```diff
@@ -131,102 +189,39 @@ describe("full story", function () {
...
-      const end = await streamingConn.read();
-      expect(end).toEqual({
-        done: true,
-        value: undefined,
-      });
+      writer.send({ value: "alpha", end: false });
+      writer.send({ value: "beta", end: false });
+      writer.send({ value: "gamma", end: false });
+      writer.send({ value: "delta", end: false });
+      writer.close();
```
timostamm, line 213: One thing we have to take care of here: writer.send() should return a promise, that only resolves when the value has been consumed by the other side, the reader. 

Imagine a client that sends gigabytes of data. It needs to wait for the data to actually leave the local buffers.

packages/connect-core/src/transform-iterable-story.spec.ts
```diff
@@ -33,6 +33,78 @@ import {
...
+
+  async process(item: IteratorResult<T, undefined>) {
+    // If there is an iterator resolver then a consumer of the async iterator is waiting on a value.  So resolve that
+    // promise with the new value being sent and return a promise that is immediately resolved
+    if (this.queueResolver) {
+      this.queueResolver(item);
+      this.queueResolver = undefined;
+      return new Promise<void>((resolve) => {
+        resolve();
+      });
```
timostamm, line 71: Could this simply be `return Promise.resolve();` ?

packages/connect-core/src/transform-iterable-story.spec.ts
```diff
@@ -33,6 +33,78 @@ import {
...
+    }
+    // Otherwise no one is waiting on a value yet so add it to the queue and return a promise that will be resolved
+    // when someone reads this value
+    this.queue.push(item);
+    return new Promise<void>((resolve) => {
+      this.sendResolver = resolve;
+    });
```
timostamm, line 78: Consider the case where the user calls send() multiple times, without waiting for the returned promise to resolve:

```ts
function f(w: Writer<string>): Promise {
  return Promise.all(
    w.send("a"),
    w.send("b"),
    w.send("c"),
  );
}
```

In this case, the promises for "a" and "b" are never resolved, because `sendResolver` is overwritten by sending "c".

I think we have two options here: 1. Prevent concurrent send by raising an error. 2. Push not just the value to be sent onto the queue, but also its `sendResolver`. I think option 2 is more forgiving.

I think we should add a test for this situation.

packages/connect-core/src/transform-iterable-story.spec.ts
```diff
@@ -33,6 +33,78 @@ import {
...
+  }
+  async send(item: T) {
```
timostamm, line 80: One thing we don't account for yet is errors. Consider a client doing this:

```ts
async function f(w: Writer) {
  const dbCon = ...
  try {
    for (;;) {
      const row = dbConn.getNextRow();
      if (!row) break;
      await w.write( row );
    }
  } finally {
    dbConn.close();
  }
}
```

We need some way to tell the calling code to stop in case of an error. Our `CallOptions` also accept a `signal?: AbortSignal` that needs to stop the calling code when triggered (AbortError).

One option is to implement `throw` in the async iterator. The the code handling this `Writer` needs to make use of it. Another option is to have a facility like `abort(reason)` in `Writer`. 

There is actually a similar situation for handlers, where we want to abort the asynchronous iterable of a server-streaming RPC implementation in case of an error (request aborted, error when serializing the response, or the client just hung up). We do not have a solution for this case in handlers yet either.

smaye81: @timostamm I pushed up some changes that took a stab at trying to handle these errors using a `throw` implementation.  It _seems_ ok, although I'm not confident it's correct and probably needs work.  Please take a look and let me know.

One thing I'm stuck on is that I'd like the consumer to be able to call `throw` on the `Reader` and have that bubble up to the `Writer` also by calling its `throw`.  But the `pipe` call returns a generator function, which behaves differently when `throw` is called (it throws the error given).

timostamm: Using AsyncIterator.throw sounds great. From the tests that https://github.com/bufbuild/connect-web/pull/431, it seems like a decent approach.

Do we need the capability to throw on the reader? Clients and handlers will operate on the iterable, and we only have [StreamingConn.read](https://github.com/bufbuild/connect-web/blob/3b8d028a88c509a92f841fd3c262273bcbbaa4f6/packages/connect-core/src/interceptor.ts#L273) reading from an asynchronous iterable. I think it should be fine to have users use an AbortSignal if they want to abort a request because they can't process response messages?

timostamm (commented):
Ok, let's get this merged, but:

- Our `StreamingConn` needs to know whether it's already closed, and throw an error on send() or close() if it already is. We should move this functionality into the writer.
- process() should not be exposed
- For composition in our clients, it should be possible to assign a method to a variable, and call it as a function. We cannot use a class for this.
- The type should explicitly implement AsyncIterable<T>

I suggest the following interface:

```ts
interface WritableIterable<T> extends AsyncIterable<T>{
  write: (payload: T) => Promise<void>;
  close: () => Promise<void>;
  isClosed: () => boolean;
}
```

And a `function createWritableIterable()`, both in connect-core/async-iterable.ts, and both exported from the package.

I've got the gRPC-web tests running with the interface above and this piece of code:

```ts
export function createWritableIterable<T>(): WritableIterable<T> {
  const w = new Writer<T>();
  let closed = false;
  return {
    isClosed() {
      return closed;
    },
    async write(payload) {
      if (closed) {
        throw new ConnectError("cannot write, already closed");
      }
      return w.send(payload);
    },
    async close() {
      if (closed) {
        throw new ConnectError("cannot close, already closed");
      }
      closed = true;
      return w.close();
    },
    [Symbol.asyncIterator]() {
      return w[Symbol.asyncIterator]();
    }
  };
}
```

Could you adapt the code?

For later (let's get this merged before we do this): The "client integration" tests should not be concerned with the details of the writer. It should have dedicated tests for its behavior, and "client integration" should use an object that's very close to our `StreamingConn`.

packages/connect-core/src/async-iterable.ts
```diff
@@ -1126,3 +1126,136 @@ export function makeIterableAbortable<T>(
...
+          // drained and rejected with the given error
+          for (const item of queue) {
+            if (item.reject) {
+              item.reject(e);
+            }
+          }
+          queue = [];
+          return new Promise<IteratorResult<T>>((resolve) => {
+            resolve({ value: String(e), done: true });
+          });
```
timostamm, line 1256: Hm, why return the error as a string here?

Wouldn't a simple `undefined` be more fitting?

```suggestion
          return new Promise<IteratorResult<T>>((resolve) => {
            resolve({ value: undefined, done: true });
          });
```

packages/connect-core/src/async-iterable.ts
```diff
@@ -1126,3 +1126,136 @@ export function makeIterableAbortable<T>(
...
+    queue.push(elem);
+
+    return prom;
+  };
+  const send = async (payload: T) => {
+    return process({ value: payload, done: false });
+  };
+  const close = async () => {
+    return process({ value: undefined, done: true });
+  };
```
timostamm, line 1203: These two functions are only called in the corresponding `write` and `close` methods - can't we simply call `process` there?

This works the same, but doesn't need `const close ...`:

```ts
    async close() {
      if (closed) {
        throw new ConnectError("cannot close, already closed");
      }
      closed = true;
      return process({ value: undefined, done: true });
    },
```

packages/connect-core/src/async-iterable.ts
smaye81: I don't know, I feel like it's nice to have the error information here returned.  That way you know that the reason the writer is returning `done` is because `throw` was invoked with an error.  

I don't feel super strongly about it though.  Just feels more intuitive than an empty string.

timostamm: You are basically doing this:

```ts
async function* gen() {
  try {
    yield 1;
  } catch (e) {
    return String(e)
  }
}
```

I'm not sure there's a good answer here, but this doesn't seem right. Let's just return `undefined`.

smaye81: Sounds good.  Updated.

timostamm (approved):

PR #427, opened 2023-01-19
Update asynchronous iterable transforms

timostamm:
This updates the code for asynchronous iterables add in https://github.com/bufbuild/connect-web/pull/422:
- adds 3 new transforms
- drops some try-catch blocks (let's keep it simply, we can rely on interface contracts)
- adds the concept of a "sink" (a function that consumes an iterable)
- adds the function pipeTo() that ends in a sink

Other changes:
- Moves compression constants and validation of compression options to connect-core
- Adds constants for content types to connect-core
- Adds more constants for header field names to connect-core


packages/connect-core/src/transform-iterable.ts
```diff
@@ -40,166 +40,368 @@ import { compressedFlag, Compression } from "./compression.js";
...
+  options?: PipeOptions
+): Promise<T11>;
+// prettier-ignore
+/**
+ * Takes an asynchronous iterable as an input, applies transformations, and
+ * passes it to a sink.
+ */
+export function pipeTo<T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12>(
+  iterable: AsyncIterable<T1>,
+  transform1: AsyncIterableTransform<T1, T2>,
```
smaye81, line 203: Would be cool to encapsulate these somehow into a class or entity like a WHATWG pipe chain called `AsyncIterableTransformChain` or something that had all these overloads inside, which would allow these signatures to be cleaned up a bit to something like:

```ts
export function pipeTo(
  iterable: AsyncIterable<I>,
  transformChain: AsyncIterableTransformChain<I, O>,
  sink: AsyncIterableSink<O, T>,
  options?: PipeOptions
): Promise<T>;
```
No clue at all if that would work, but just spitballing.

smaye81 (approved):

PR #428, opened 2023-01-20
Reuse HTTP2 Session

gilwong00:
This adds the ability to make the Http2 session reusable with multiple request instead of creating a new instance with each request


gilwong00:
> This looks nice and simple 👍
> 
> In our tests, we don't wait for servers to shut down because it takes too long, and I suspect that this is because of open http2 sessions. Can you try removing the following lines and see if that noticeably slows down the tests?
> 
> * https://github.com/bufbuild/connect-web/blob/fbb089ed2de52412f4ae76126fe7fcdd8f6dc525/packages/connect-node-test/src/helpers/testserver.ts#L164
> * https://github.com/bufbuild/connect-web/blob/fbb089ed2de52412f4ae76126fe7fcdd8f6dc525/packages/connect-node-test/src/helpers/testserver.ts#L194
> * https://github.com/bufbuild/connect-web/blob/fbb089ed2de52412f4ae76126fe7fcdd8f6dc525/packages/connect-node-test/src/helpers/testserver.ts#L255

hey @timostamm I gave it a shot and unfortunately it does significantly slow down our test a lot. Should I investigate why this is the case or its ok for now?

timostamm:
> hey @timostamm I gave it a shot and unfortunately it does significantly slow down our test a lot. Should I investigate why this is the case or its ok for now?

Ah, well. Please just update the comments on those lines:

```ts
resolve(); // the server.close() callback above slows down our tests
```

timostamm (approved):
LGTM, but please add a PR description.

PR #430, opened 2023-01-20
Switch handlers to use asynchronous iterables

timostamm:


packages/connect-node-test/src/badweather/broken-input.spec.ts
```diff
@@ -20,130 +20,141 @@ import {
...
-        "POST",
-        url,
-        { "content-type": type },
-        body,
-        opt
-      );
-    }
+  // TODO(TCN-785) add @bufbuild/connect-node, cover gRPC and gRPC-web, cover invalid protobuf binary input
+  servers.describeServers(
+    ["connect-go (h2)", "@bufbuild/connect-node (h2c)"],
```
timostamm, line 30: This enables the tests added in https://github.com/bufbuild/connect-web/pull/350 to run against @bufbuild/connect-node. 

The diff for those files is larger than the actual change, because indentation changed, and the test helper `http2Request()` was updated to take an object argument for readability.

The only actual change here was in the error message expectations, and adding a Content-Type to requests with an unsupported HTTP verb because connect-go and connect-node slightly differ in implementation in this case ( connect-node always returns 415 on Content-Type mismatch, regardless of verb).

packages/connect-node-test/src/handler.spec.ts
```diff
@@ -54,7 +53,9 @@ describe("createHandler()", () => {
       TestService.methods.emptyCall,
       (req) => req,
       {
-        protocols: [createConnectProtocol({})],
+        connect: true,
+        grpcWeb: false,
+        grpc: false,
```
timostamm, line 58: This PR changes the public API for creating handlers. The functions `createXProtocol()` are no longer available. 
All protocols are still enabled by default. You can disable individual protocols with a boolean property.

packages/connect-node/src/private/validate-read-max-bytes-option.ts
```diff
@@ -16,6 +16,9 @@ import { Code, ConnectError } from "@bufbuild/connect-core";
 
 const maxReadMaxBytes = 0xffffffff; // zlib caps maxOutputLength at this value
 
+/**
+ * @deprecated use compressionValidateOptions from @bufbuild/connect-core
+ */
```
timostamm, line 21: Deprecated a couple of functions that are still used in transports. They will be replaced when we switch the Node transports to use asynchronous iterables.

packages/connect-node/src/private/node-universal.ts
```diff
@@ -0,0 +1,173 @@
...
+  request: NodeRequest,
+  response: NodeResponse
+) => void;
+type NodeRequest = http.IncomingMessage | http2.Http2ServerRequest;
+type NodeResponse = http.ServerResponse | http2.Http2ServerResponse;
+
+/**
+ * Convert a universal handler to a Node.js handler function.
+ */
+export function universalHandlerToNodeHandler(
```
timostamm, line 38: For handlers, this is now the single bridge between Node.js and our implementation.

packages/connect-node/src/grpc-web-handler.ts
```diff
@@ -0,0 +1,182 @@
...
+      }),
+      transformCompressEnvelope(compression.response, opt.compressMinBytes),
+      transformJoinEnvelopes()
+    );
+    return {
+      ...uResponseOk,
+      body: outputIt,
+      header: context.responseHeader,
+    };
+  };
```
timostamm, line 181: This is the implementation of gRPC-web handlers and shows the meat of this PR. Granted, we are shifting a lot of the complexity into connect-core, but the result seems more manageable than what we previously had.

smaye81 (approved):

PR #431, opened 2023-01-25
Propagate downstream errors in asynchronous iterables

timostamm:
This adds the option `propagateDownStreamError?: boolean` to `pipe()` and `pipeTo()`.

For context: 

If iterators are chained, any error raised by the source or any transform travels down the stream. But if an error happens downstream, the source and transformations are left dangling:

```ts
async function source*() {
  const conn = await dbConn();
  yield await conn.query("SELECT 1"); // consumed downstream
  yield await conn.query("SELECT 2"); // never consumed
  conn.close(); // never runs
}
for await (const element of source()) {
  // let's say we try to write the element to disk, but the disk is full
  throw "err";
}
```

If this option is set to true, an error raised by the sink function given to pipeTo() will raise the same error in the source iterable.

```ts
async function source*() {
  const conn = await dbConn();
  try {
    yield await conn.query("SELECT 1"); // consumed downstream
    yield await conn.query("SELECT 2"); // never consumed
  } finally {
    conn.close(); // runs!
  }
}
await pipeTo(source(), async iterable => {
  for await (const element of source()) {
    // let's say we try to write the element to disk, but the disk is full
    throw "err";
  }
}, { propagateDownStreamError: true });
```

If this option is set to true with pipe(), the downstream consumer of the iterable returned by pipe() can abort the source iterable by calling throw() on the iterator.

For more context, see [this tc39 discussion](https://github.com/tc39/proposal-async-iteration/issues/126). This PR may not be the perfect solution, but it seems like a reasonable approach to get the abort behavior in streams that we need.

### DIFF
--- a/packages/connect-core/src/async-iterable.spec.ts
+++ b/packages/connect-core/src/async-iterable.spec.ts
@@ -13,7 +13,9 @@
 // limitations under the License.
 
 import {
+  makeIterableAbortable,
   pipe,
+  pipeTo,
   transformCatch,
   transformCompressEnvelope,
   transformDecompressEnvelope,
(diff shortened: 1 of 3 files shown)


packages/connect-core/src/async-iterable.spec.ts
```diff
@@ -35,6 +37,389 @@ import { Code } from "./code.js";
...
+        consumerLog.push("received " + chunk);
+        throw "CONSUMER_ERROR";
+      }
+    }
+
+    try {
+      await consume(source());
+      fail("expected error");
+    } catch (e) {
+      expect(e).toBe("CONSUMER_ERROR");
```
smaye81, line 108: I'm not sure if this is worth it or not, but you could maybe call `source.throw(e)` here and test that the `finally` was written?

smaye81 (approved):

packages/connect-core/src/async-iterable.spec.ts
timostamm: This is just the baseline test, proving that a plain for-await loop leaves the iterator dangling, and that `finally` is never reached.

In `pipe()` / `with error-raising consumer`, we actually call `AsyncIterator.throw`, and we assert that this
- `should not propagate error to source by default, and leave it dangling`
- `should propagate error to source with propagateDownStreamError: true`

So I think we're covered pretty well?

PR #433, opened 2023-01-25
Start migrating clients to asynchronous iterables

timostamm:
This adds a function signature for universal clients, similar to `UniversalHandlerFn` for handlers:

```ts
type UniversalClientFn = (request: UniversalClientRequest) => Promise<UniversalClientResponse>;
```

There are two new functions `createNodeHttp1Client()` and `createNodeHttp2Client()`, which return such a universal client function. 

Basic functionality is tested over http/1.1 and http/2 against connect-go and connect-node, with an ad-hoc implementation of gRPC-web.

I am sure we will change the actual implementations of the Node.js constructor functions going forward, but this should be a good start for migrating all Node.js transports to asynchronous iterables.

### DIFF
--- a/packages/connect-node/src/private/node-universal-client.ts
+++ b/packages/connect-node/src/private/node-universal-client.ts
@@ -1,0 +1,288 @@
+// Copyright 2021-2023 Buf Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import * as http2 from "http2";
+import * as http from "http";
+import * as https from "https";
+import {
+  AsyncIterableSink,
+  Code,
+  ConnectError,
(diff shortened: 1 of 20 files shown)


packages/connect-node/src/private/node-universal-client.ts
```diff
@@ -0,0 +1,290 @@
...
+          function connect() {
+            socket.off("error", rejectError);
+            resolve(request);
+          }
+
+          function rejectError(err: Error) {
+            socket.off("connect", connect);
+            error(err);
+          }
+        });
```
timostamm, line 77: @gilwong00, this ports https://github.com/bufbuild/connect-web/pull/390 - do you spot any obvious issues? (I know we are not converting errors yet)

packages/connect-node/src/private/node-universal-client.ts
```diff
@@ -0,0 +1,290 @@
...
+          origin,
+          {
+            ...http2Options,
+          },
+          (s) => resolve(s)
+        );
+        s.on("error", (err) => reject(err));
+      }
+    ));
+  }
```
timostamm, line 157: @gilwong00, this ports https://github.com/bufbuild/connect-web/pull/428 - do you spot any obvious issues?

packages/connect-node/src/private/node-universal-client.ts
gilwong00: nope, this looks good!

packages/connect-node/src/private/node-universal-client.ts
gilwong00: same here, looks good!

gilwong00 (approved):

PR #435, opened 2023-01-27
Migrate Node.js clients to asynchronous iterables

timostamm:
This migrates all `Transport` implementations for Node.js to asynchronous iterables.

Notable changes:

Instead of `createConnectHttp2Transport()` and `createConnectHttpTransport()`, there is now a single `function createConnectTransport()` that requires a `httpVersion: "2" | "1.1"` option. Same for the gRPC and the gRPC-web transports.

We're removing about 1750 lines of code with this, but unfortunately, we are not propagating errors correctly yet. This disables the "unresolvable_host" tests for streaming until we have a better mechanism in place.


packages/connect-core/src/protocol-grpc-web/validate-response.ts
```diff
@@ -13,32 +13,42 @@
...
 /**
  * Validates response status and header for the gRPC-web protocol.
+ *
  * Throws a ConnectError if the header contains an error status,
  * the HTTP status indicates an error, or if the content type is
  * unexpected.
+ *
+ * Returns an object that indicates whether a gRPC status was found
+ * in the response header. In this case, clients can not expect a
+ * trailer.
```
timostamm, line 37: An unrelated addition to handle trailers-only responses more leniently.

smaye81 (approved):

PR #436, opened 2023-01-31
Migrate transports to asynchronous iterables

timostamm:
We switched handlers to use asynchronous iterables in https://github.com/bufbuild/connect-web/pull/430, and clients in https://github.com/bufbuild/connect-web/pull/435.

This PR changes the `interface Transport` to use simple `AsyncIterable` for input and output streams, and it also changes clients with streaming input to accept an `AsyncIterable` instead of using a writer abstraction. 

This PR re-enables the following tests:
- "unresolvable_host", disabled in https://github.com/bufbuild/connect-web/pull/435
- "unimplemented_server_streaming_service", disabled in https://github.com/bufbuild/connect-web/pull/374


packages/connect-core/src/interceptor.spec.ts
```diff
@@ -14,44 +14,39 @@
...
+  async function* logStream(res: StreamResponse) {
+    for await (const m of res.message) {
+      log.push(`${name} response stream received message`);
+      yield m;
+    }
+    yield* res.message;
+    log.push(
+      `${name} response stream done with trailers: ${headerKeys(res.trailer)}`
+    );
+  }
```
timostamm, line 48: Intercepting a streaming response is a bit more straight-forward with this.

packages/connect-core/src/interceptor.ts
```diff
@@ -54,8 +51,8 @@ export type Interceptor = (next: AnyFn) => AnyFn;
  * response to inspect and log.
  */
 type AnyFn = (
-  req: UnaryRequest | StreamingRequest
-) => Promise<UnaryResponse | StreamingConn>;
+  req: UnaryRequest | StreamRequest
+) => Promise<UnaryResponse | StreamResponse>;
```
timostamm, line 55: A streaming function takes a `StreamRequest` and returns the promise for a `StreamResponse`. It's actually the same as a unary function, just that instead of a single message `T`, it's a `AsyncIterable<T>` for streams.

packages/connect-core/src/promise-client.ts
```diff
@@ -39,99 +39,11 @@ export type PromiseClient<T extends ServiceType> = {
   [P in keyof T["methods"]]:
     T["methods"][P] extends MethodInfoUnary<infer I, infer O>           ? (request: PartialMessage<I>, options?: CallOptions) => Promise<O>
   : T["methods"][P] extends MethodInfoServerStreaming<infer I, infer O> ? (request: PartialMessage<I>, options?: CallOptions) => AsyncIterable<O>
-  : T["methods"][P] extends MethodInfoClientStreaming<infer I, infer O> ? (options?: CallOptions) => Promise<ClientStreamCall<PartialMessage<I>, O>>
-  : T["methods"][P] extends MethodInfoBiDiStreaming<infer I, infer O>   ? (options?: CallOptions) => Promise<ClientBiDiCall<PartialMessage<I>, O>>
+  : T["methods"][P] extends MethodInfoClientStreaming<infer I, infer O> ? (request: AsyncIterable<PartialMessage<I>>, options?: CallOptions) => Promise<O>
+  : T["methods"][P] extends MethodInfoBiDiStreaming<infer I, infer O>   ? (request: AsyncIterable<PartialMessage<I>>, options?: CallOptions) => AsyncIterable<O>
```
timostamm, line 43: This is the relevant change for streaming in the promise-client. Note that since connect-web does not support streaming input, this change is not breaking.

smaye81 (approved):

PR #456, opened 2023-02-12
Extend universal handlers to support pre-parsed JSON requests

timostamm:
Fastify, express, and other servers parse `application/json` automatically. To support this behavior, this changes the "body" property of universal server requests to accept either a byte stream, or a JSON value.

The Connect protocol handler - the only one that needs to handle this behavior - was updated to support JSON request bodies.

Note that the `function createHandler()` exported by `@bufbuild/connect-node` does not make use of the new capability yet - this will be implemented in a follow-up.

Other changes:
- removes superseded `parseContentType` functions
- moves logic for protocol negotiation from the Node.js world to `createUniversalHandler()`
- adds `ContentTypeMatcher` to cache RegExp invocations
- makes I/O limits available outside of streams, for Connect unary


packages/connect-node/src/protocol-handler.ts
```diff
@@ -85,8 +94,94 @@ export interface UniversalHandler extends UniversalHandlerFn {
   allowedMethods: string[];
 
   /**
-   * A regular expression that matches all Content-Type header values that this
-   * procedure supports.
+   * A matcher for Content-Type header values that this procedure supports.
    */
-  supportedContentType: RegExp;
+  supportedContentType: ContentTypeMatcher;
```
timostamm, line 99: `ContentTypeMatcher` allows us to use the same type for this property in individual handlers for a protocol, and in merged handlers that handle multiple protocols.

packages/connect-node/src/grpc-web-transport.ts
```diff
@@ -166,8 +167,7 @@ export function createGrpcWebTransport(
             body: pipe(
               createAsyncIterable([req.message]),
               transformSerializeEnvelope(
-                serialization.getI(opt.useBinaryFormat),
-                opt.writeMaxBytes
```
timostamm, line 170: The I/O limit is now added to `Serialization` via `limitSerialization()`. That means we can apply limits to Connect unary without going through boilerplate.

packages/connect-node/src/private/universal.ts
```diff
@@ -54,7 +56,13 @@ export interface UniversalServerRequest {
...
   method: string;
   header: Headers;
-  body: AsyncIterable<Uint8Array>;
+  /**
+   * Many server frameworks parse request bodies with the mime type
+   * application/json automatically. We accept a JSON value as an
+   * alternative to a byte stream here so that this situation can be
+   * handled efficiently.
+   */
+  body: AsyncIterable<Uint8Array> | JsonValue;
```
timostamm, line 65: Adding `JsonValue` for JSON requests that are already parsed by a framework.

smaye81 (approved):

PR #459, opened 2023-02-13
Test more versions of Node.js

timostamm:
This runs tests in two additional versions of Node.js:
- v17.9.1
- v16.0.0

It also bumps the v16 and v18 versions we already run to their latest release.


timostamm:
Closing in favor of https://github.com/bufbuild/connect-web/pull/462

PR #460, opened 2023-02-14
Create protoc-gen-connect-es plugin

smaye81:
This adds the new plugin `protoc-gen-connect-es` alongside the current `protoc-gen-connect-web`.  Eventually, we would like to move users over to use this plugin.

The previous `protoc-gen-connect-web` plugin is a bit of a misnomer because it is capable of generating code that works for Node.js and the web.  This is mostly an exact replica of that plugin except that it generates files with a `_connect` suffix instead of `_connectweb`.


Makefile
```diff
@@ -95,26 +101,26 @@ $(GEN)/connect-core: node_modules/.bin/protoc-gen-es packages/connect-core/buf.g
 	@mkdir -p $(@D)
 	@touch $(@)
 
-$(GEN)/connect-web-test: node_modules/.bin/protoc-gen-es $(BUILD)/protoc-gen-connect-web packages/connect-web-test/buf.gen.yaml Makefile
+$(GEN)/connect-web-test: node_modules/.bin/protoc-gen-es $(BUILD)/protoc-gen-connect-web $(BUILD)/protoc-gen-connect-es packages/connect-web-test/buf.gen.yaml Makefile
```
smaye81, line 104: @timostamm I added this alongside the connect-web plugin because my thought was we can just go and remove it at some point without having to remove and then add.  Lmk if you'd rather it be a different configuration.

packages/protoc-gen-connect-es/README.md
```diff
@@ -0,0 +1,204 @@
+# @bufbuild/protoc-gen-connect-es
+
+The code generator for connect-es, a simple library to work with servers and clients
```
timostamm, line 3:
```suggestion
The code generator for Connect, a simple library to work with servers and clients
```

This is shown on npmjs.com, so calling it just "Connect" doesn't seem totally out of place. Gives us a bit more wiggle room for the transition. WDYT?

Makefile
timostamm: No buf.gen.yaml uses `protoc-gen-connect-web` anymore, right? 

We still want to be able to make bugfixes, so we should keep it in `build` and in `release` for now.

But can't we remove it here and at other places?

packages/protoc-gen-connect-es/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "@bufbuild/protoc-gen-connect-es",
+  "version": "0.7.0",
+  "description": "Code generator for connect-es",
```
timostamm, line 4:
```suggestion
  "description": "Code generator for Connect",
```

?

timostamm (commented):
Considering we are using `protoc-gen-connect-es` in the examples now, I think we should publish this. We can do this without bumping the version.

The top-level README.md points to `protoc-gen-connect-web`. Can you replace it?

packages/protoc-gen-connect-es/README.md
smaye81: sgtm.  updated.

Makefile
smaye81: Yeah, good call.  Now that it's removed from those `buf.gen.yaml` files, it's not needed in the individual package pre-reqs any more.  Removed.

smaye81:
> Considering we are using `protoc-gen-connect-es` in the examples now, I think we should publish this. We can do this without bumping the version.
> 
> The top-level README.md points to `protoc-gen-connect-web`. Can you replace it?

Done.

packages/protoc-gen-connect-web/src/javascript.ts
```diff
@@ -26,7 +26,7 @@ export function generateJs(schema: Schema) {
     if (protoFile.services.length == 0) {
       continue;
     }
-    const file = schema.generateFile(protoFile.name + "_connectweb.js");
+    const file = schema.generateFile(protoFile.name + "_connect.js");
```
timostamm, line 29: Should not change the existing `protoc-gen-connect-web`.

packages/protoc-gen-connect-web/README.md
```diff
@@ -104,7 +104,7 @@ This option controls whether the plugin generates JavaScript, TypeScript,
 or TypeScript declaration files.
 
 Possible values:
-- `target=js` - generates a `_connectweb.js` file for every `.proto` input file.
+- `target=js` - generates a `_connect.js` file for every `.proto` input file.
```
timostamm, line 107: Should not change the existing `protoc-gen-connect-web`.

packages/protoc-gen-connect-web/src/javascript.ts
smaye81: Ugh.  Sorry, was a find/replace mishap.

timostamm (approved):

bufdev:
We're in beta - there's no need to maintain compatibility, this is what bets is for. We can just delete the old plugin including in remote plugins.

timostamm:
I think you're right @bufdev, we can just remove packages/protoc-gen-connect-web from here, merge this PR and publish the new protoc-gen-connect-es. We also need to get the remote plugin for this out ASAP.

timostamm (approved):

PR #462, opened 2023-02-15
Polyfill headers in connect-node

smaye81:
This a few things regarding connect-node and the versions supported by the framework:

- Moves the Headers polyfill logic to `connect-node` instead of the test package.  This will automatically polyfill connect-node for users on versions less than 18.
- Adds an `engine` definition to our package.json for Node versions.
- Adds additional tests for Node 17


timostamm (commented):
Nice! Does this replace https://github.com/bufbuild/connect-web/pull/459 ?

smaye81:
> Nice! Does this replace #459 ?

Mostly yeah.  I can merge in the versions you're using in 459 though if you'd like (including the use of a MIN_VERSION)

timostamm (approved):
> > Nice! Does this replace https://github.com/bufbuild/connect-web/pull/459 ?
> 
> Mostly yeah. I can merge in the versions you're using in 459 though if you'd like (including the use of a MIN_VERSION)

Don't have a strong opinion on the exact versions, as long as we test the lowest one we have in the engines field.

PR #475, opened 2023-02-21
Fix Fastify README.md

timostamm:
A headline was referring to `connectNodeAdapter` instead of `fastifyConnectPlugin`.

Also fixes two imports in the example.

### DIFF
--- a/packages/connect-fastify/README.md
+++ b/packages/connect-fastify/README.md
@@ -48,7 +48,7 @@ With that server running, you can make requests with any gRPC, gRPC-Web, or Conn
 `buf curl` with the gRPC protocol:
 
 ```bash
-buf curl --schema /Users/ts/buf/connect-web-main/packages/example/eliza.proto \
+buf curl --schema buf.build/bufbuild/eliza \
   --protocol grpc --http2-prior-knowledge \
   -d '{"sentence": "I feel happy."}' \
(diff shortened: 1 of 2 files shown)


packages/connect-fastify/README.md
```diff
@@ -43,7 +43,7 @@ await server.listen({
 });
 ```
 
-With that server running, you can make requests with any gRPC or Connect client.
+With that server running, you can make requests with any gRPC (-web) or Connect client.
```
smaye81, line 46: I feel like `gRPC, gRPC-Web, or Connect` is clearer.  Not a huge deal though.

smaye81 (approved):

smaye81:
One thing I noticed in the `buf curl` command.  Should the `schema` path be changed?  Right now it's:

`buf curl --schema /Users/ts/buf/connect-web-main/packages/example/eliza.proto`

packages/connect-fastify/README.md
timostamm: 👍

timostamm:
> buf curl --schema /Users/ts/buf/connect-web-main/packages/example/eliza.proto

Thanks for catching 🙈 

Changed to `buf curl --schema buf.build/bufbuild/eliza`

smaye81 (approved):

PR #477, opened 2023-02-21
Improve example

timostamm:
This adds a locally-trusted development certificates with `mkcert` to the example, shows how to use `curl` and `buf curl` against the server, and adds explanatory comments in the example code.


packages/example/README.md
```diff
@@ -21,14 +21,31 @@ cd connect-es-main/packages/example
...
+```
+
+If you don't use macOS or `brew`, see the [mkcert docs](https://github.com/FiloSottile/mkcert#installation) 
+for instructions. You can copy the last line to your `~/.zprofile` or `~/.profile`, so that 
+the environment variable for Node.js is set every time to open a terminal.
+
+If you already use `mkcert`, just run `mkcert localhost 127.0.0.1 ::1` to issue a certificate 
+for our example server.
+
+Once we have that out of the way, let's start the Connect server:
```
timostamm, line 41: It's really unfortunate that no browser supports H2C (plain-text HTTP/2). If we want to show both gRPC (many clients require H2) and a web client, we can't use H2C. Previously, this was using a self-signed certificate, and you had to manually accept the connect in the web browser.

`mkcert` installs a CA that is picked up by browsers and other tools like `buf curl` and it's typically a one-time setup, so I figure it is worth using it here.

packages/example/README.md
```diff
@@ -51,6 +68,25 @@ How are you feeling today?
...
+  --data '{"sentence": "I feel happy."}' \
+   https://localhost:8443/buf.connect.demo.eliza.v1.ElizaService/Say
+```
+
+We can also use any gRPC client like `buf curl`:
+
+```shell
+npx buf curl --protocol grpc --schema . -d '{"name": "John"}' \
+   https://localhost:8443/buf.connect.demo.eliza.v1.ElizaService/Introduce
+```
```
timostamm, line 88: We might as well show other clients too!

packages/example/src/client.ts
```diff
@@ -15,20 +15,29 @@
...
-    },
-  });
+let rejectUnauthorized = true;
+// prettier-ignore
+if (env.NODE_EXTRA_CA_CERTS == undefined) {
+  rl.write("! Looks like you did not let Node.js know about your certificate authority for \n");
+  rl.write("! local development yet. This is fine, we'll ignore TLS errors in this example \n");
+  rl.write("! client. \n\n");
+  rejectUnauthorized = false;
+}
```
timostamm, line 30: Of course Node.js does not pick up OS CAs - that would be too easy 😆 With this if-block, you can still run the Node.js server and client without using `mkcert` at all.

smaye81 (approved):

PR #478, opened 2023-02-21
Move polyfill dep to connect-node

smaye81:
This moves the `header-polyfill` dependency to connect-node where it is actually used.


timostamm (commented):
Makes sense.

I wonder about the `"sideEffects": false` we have in the package.json. Is it really free of side effects if I just use `import {createGrpcTransport} from "@bufbuild/connect-node"`? Will a bundler shake the polyfill call away?

timostamm (approved):

PR #479, opened 2023-02-21
Add support for express

timostamm:


packages/connect-fastify/src/fastify-connect-plugin.ts
```diff
@@ -60,30 +60,21 @@ export function fastifyConnectPlugin(
     done();
     return;
   }
+  if (opts.acceptCompression === undefined) {
+    opts.acceptCompression = [compressionGzip, compressionBrotli];
+  }
```
timostamm, line 65: Defaults for acceptCompression moved a couple of lines up. Otherwise, createConnectRouter overrides with an empty array as a default, and these defaults never apply.

packages/connect-fastify/src/fastify-connect-plugin.ts
```diff
@@ -60,30 +60,21 @@ export function fastifyConnectPlugin(
...
-    protoGrpc.contentTypeRegExp,
-  ];
-  for (const type of types) {
-    instance.addContentTypeParser(type, (_req, _payload, done) => {
-      done(null, undefined);
-    });
-  }
+  // we can override all content type parsers (including application/json) in
+  // this plugin without affecting outer scope
+  addNoopContentTypeParsers(instance);
```
timostamm, line 77: We are adding all known content-types as string _and_ as regex now. From the [fastify docs](https://www.fastify.io/docs/latest/Reference/ContentTypeParser/):

> Fastify first tries to match a content-type parser with a string value before trying to find a matching RegExp.

packages/connect-node/src/connect-transport.ts
```diff
@@ -206,10 +206,11 @@ export function createConnectTransport(
               opt.readMaxBytes
             );
           }
-          if (isConnectUnaryError) {
+          if (isUnaryError) {
             throw errorFromJsonBytes(
               responseBody,
-              appendHeaders(header, trailer)
+              appendHeaders(header, trailer),
+              unaryError
```
timostamm, line 213: We ran into https://github.com/bufbuild/connect-es/issues/469 with fastify, which replies with an application/json content type for a HTTP 404 when application/json was sent.

The logic was changed to raise a ConnectError for the HTTP status code when parsing fails to fix the issue. This change is applied to the transports from @bufbuild/connect-web and @bufbuild/connect-node.

smaye81 (approved):

PR #485, opened 2023-02-23
Declare type exports

fubhy:
Type exports for the connect package are currently not explicitly specified.

- Adds explicit type exports for every entrypoint of the `connect` package.
- Puts the type exports for all packages first (in accordance with the documentation: https://www.typescriptlang.org/docs/handbook/esm-node.html)
- Removes `typesVersions` which shouldn't be necessary afaik (please correct me if I'm missing sth.)

### DIFF
--- a/packages/connect-express/package.json
+++ b/packages/connect-express/package.json
@@ -18,9 +18,9 @@
   "type": "module",
   "types": "./dist/types/index.d.ts",
   "exports": {
+    "types": "./dist/types/index.d.ts",
     "import": "./dist/esm/index.js",
-    "require": "./dist/cjs/index.js",
-    "types": "./dist/types/index.d.ts"
+    "require": "./dist/cjs/index.js"
   },
   "engines": {
     "node": ">=16.0.0 <19"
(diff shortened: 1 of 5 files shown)


timostamm:
Thanks for the contrib, @fubhy ❤️ 

We do need the "typesVersion", unfortunately. 

The repo uses "Node" module resolution, and if you run `make build` without "typesVersion", you'll see that failing with:

```
src/connect-transport.ts:47:8 - error TS7016: Could not find a declaration file for module '@bufbuild/connect/protocol'. '/Users/ts/buf/connect-es/packages/connect/protocol.js' implicitly has an 'any' type.
  If the '@bufbuild/connect' package actually exposes this module, try adding a new declaration (.d.ts) file containing `declare module '@bufbuild/connect/protocol';`

47 } from "@bufbuild/connect/protocol";
          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Many projects are stuck with the node module resolution algorithm. It's still used when you run create-react-app. We need to support it.

Can you revert this one? Otherwise, this looks good to me!

fubhy:
Ah right. Good catch, thanks! Will revert tha tone in a moment.

timostamm (approved):
I just ran this in a project using NodeNext module resolution - works like a charm.

LGTM!

PR #488, opened 2023-02-23
Update badges

smaye81:
This updates the Connect-ES badges, but also removes `private` from connect-web so that it can be published.


README.md
```diff
@@ -2,7 +2,7 @@
...
+[![License](https://img.shields.io/github/license/bufbuild/connect-es?color=blue)](./LICENSE) [![NPM Version](https://img.shields.io/npm/v/@bufbuild/connect/latest?color=green&label=%40bufbuild%2Fconnect)](https://www.npmjs.com/package/@bufbuild/connect) [![NPM Version](https://img.shields.io/npm/v/@bufbuild/protoc-gen-connect-es/latest?color=green&label=%40bufbuild%2Fprotoc-gen-connect-es)](https://www.npmjs.com/package/@bufbuild/protoc-gen-connect-es) [![NPM Version](https://img.shields.io/npm/v/@bufbuild/connect-node/latest?color=green&label=%40bufbuild%2Fconnect-node)](https://www.npmjs.com/package/@bufbuild/connect-node) [![NPM Version](https://img.shields.io/npm/v/@bufbuild/connect-fastify/latest?color=green&label=%40bufbuild%2Fconnect-fastify)](https://www.npmjs.com/package/@bufbuild/connect-fastify) [![NPM Version](https://img.shields.io/npm/v/@bufbuild/connect-express/latest?color=green&label=%40bufbuild%2Fconnect-express)](https://www.npmjs.com/package/@bufbuild/connect-express) [![NPM Version](https://img.shields.io/npm/v/@bufbuild/connect-web/latest?color=green&label=%40bufbuild%2Fconnect-web)](https://www.npmjs.com/package/@bufbuild/connect-web)
```
timostamm, line 5: I removed the other badges earlier because I think they are becoming distracting. They will always have matching versions anyways (unless we have private: true somewhere 😆). 

If you think showing all packages on the main page, I'd prefer if we added them to the "Packages" section, where we can also add a description and a link to the source.

WDYT?

smaye81: That's fine.  I can remove them.  I just it as a way to see all packages this repo publishes very quickly.

timostamm: It's possible that I went a bit too far removing the package list, but we'll have three rows of badges soon:

Screenshot 2023-02-23 at 21 06 52

So I think adding them back in the packages section would be less overwhelming.

timostamm (approved):
Whatever we do with badges, thanks for fixing the private package 🙈

PR #506, opened 2023-03-02
Updates connect logo to use transparent background

dimitropoulos:
The previous image asset didn't have a transparent background which resulted in it looking funky in dark mode.

| BEFORE | AFTER |
| - | - |
| Screenshot_20230302_115201 | Screenshot_20230302_115227 |


dimitropoulos:
@akshayjshah yes, but we'll forever pay the cost in our git tree: :gravestone: :rip:

PR #511, opened 2023-03-06
Update "Getting started" section in the @bufbuild/connect README.md

timostamm:
This replaces the "Getting started" section in the README.md of @bufbuild/connect with the sections "Get started on the web" and "Get started on Node.js" we added in https://github.com/bufbuild/connect-es/pull/495.

It also fixes some deprecation messages by escaping at-signs - thanks to @dimitropoulos for the shout.

### DIFF
--- a/packages/connect-web/src/index.ts
+++ b/packages/connect-web/src/index.ts
@@ -42,71 +42,71 @@ import {
 
 export {
   /**
-   * @deprecated please import from @bufbuild/connect instead
+   * @deprecated please import from \@bufbuild/connect instead
    */
   createPromiseClient,
   /**
-   * @deprecated please import from @bufbuild/connect instead
+   * @deprecated please import from \@bufbuild/connect instead
    */
   createCallbackClient,
   /**
-   * @deprecated please import from @bufbuild/connect instead
+   * @deprecated please import from \@bufbuild/connect instead
    */
   CallbackClient,
   /**
-   * @deprecated please import from @bufbuild/connect instead
+   * @deprecated please import from \@bufbuild/connect instead
    */
   PromiseClient,
   /**
-   * @deprecated please import from @bufbuild/connect instead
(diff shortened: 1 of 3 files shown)


packages/connect/README.md
```diff
@@ -41,11 +41,25 @@ optimized for the web. This gives you unparalleled interoperability with
...
+and we have an expansion pack for [TanStack Query](https://github.com/bufbuild/connect-query).
+We support all modern web browsers that implement the widely available
+[fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API)
+and the [Encoding API](https://developer.mozilla.org/en-US/docs/Web/API/Encoding_API).
 
+
+## Get started on Node.js
+
+Follow our [10 minute tutorial](https://connect.build/docs/node/getting-started)
+to spin up service in Node.js, and call it from the web, and from a gRPC client
```
smaye81, line 60: nit:  spin up a service

smaye81 (approved):

packages/connect/README.md
timostamm: TY

packages/connect-web/src/index.ts
```diff
@@ -42,71 +42,71 @@ import {
 
 export {
   /**
-   * @deprecated please import from @bufbuild/connect instead
+   * @deprecated please import from \@bufbuild/connect instead
```
dimitropoulos, line 45: just to double check: this definitely is great as a fix for the escaping, but you know it's still the case that users will not be able to see this message, right?  are you planning to tackle that in a separate PR?

timostamm: I'm not sure there is a good solution to this. See https://github.com/microsoft/TypeScript/issues/42684 and https://github.com/microsoft/TypeScript/pull/47293

PR #516, opened 2023-03-07
Add missing JSDoc for new handler and transport options

timostamm:
This adds documentation for several new options for client transports and servers that were added in the last weeks.

### DIFF
--- a/packages/connect/src/implementation.ts
+++ b/packages/connect/src/implementation.ts
@@ -56,16 +56,44 @@ interface MI<
   readonly idempotency?: MethodIdempotency;
 }
 
-// TODO document
+/**
+ * Context for an RPC on the server. Every RPC implementation can accept a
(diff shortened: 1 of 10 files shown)


packages/connect/src/implementation.ts
```diff
@@ -56,16 +56,43 @@ interface MI<
   readonly idempotency?: MethodIdempotency;
 }
 
-// TODO document
+/**
+ * Context for an RPC on the server. Every RPC implementation can accept a
+ * HandlerContext as an argument, and
```
smaye81, line 61: Should there be additional text here after `, and...`?

smaye81 (approved):

packages/connect/src/implementation.ts
timostamm: TY

PR #520, opened 2023-03-07
Add support for Connect-Protocol-Version header

timostamm:
To allow for proxies to easily distinguish Connect RPC requests from other HTTP requests with the `application/json` content-type, we have introduced the `Connect-Protocol-Version` header.

Clients from this repository send this header along with every request since [v0.6.0](https://github.com/bufbuild/connect-es/releases/tag/v0.6.0). `connect-go` supports the header since [v1.4.0](https://github.com/bufbuild/connect-go/releases/tag/v1.4.0).

With this PR, servers can opt in to require this header with an option:

```ts
  /**
   * Require requests using the Connect protocol to include the header
   * Connect-Protocol-Version. This ensures that HTTP proxies and other
   * code inspecting traffic can easily identify Connect RPC requests,
   * even if they use a common Content-Type like application/json.
   *
   * If a Connect request does not include the Connect-Protocol-Version
   * header, an error with code invalid_argument (HTTP 400) is returned.
   * This option has no effect if the client uses the gRPC or the gRPC-web
   * protocol.
   */
  requireConnectProtocolHeader: boolean;
```


packages/connect/src/protocol/universal-handler.ts
```diff
@@ -189,12 +209,33 @@ export function createUniversalMethodHandler(
...
+ * handle exactly the same RPC.
  */
-function negotiateProtocol(
-  service: ServiceType,
-  method: MethodInfo,
+export function negotiateProtocol(
   protocolHandlers: UniversalHandler[]
 ): UniversalHandler {
+  if (protocolHandlers.length == 0) {
+    throw new ConnectError("require at least one protocol", Code.Internal);
```
smaye81, line 220: maybe `at least one protocol is required`?

smaye81 (approved):

packages/connect/src/protocol-connect/handler-factory.ts
```diff
@@ -153,6 +154,7 @@ function createUnaryHandler<I extends Message<I>, O extends Message<O>>(
     let status = uResponseOk.status;
     let body: Uint8Array;
     try {
+      if (opt.requireConnectProtocolHeader) requireProtocolVersion(req.header);
```
smaye81, line 157: probably just my OCD but I like curly braces or at least this to be on the same line.  I sorta wish we enforced https://eslint.org/docs/latest/rules/nonblock-statement-body-position

smaye81 (approved):

packages/connect/src/protocol-connect/handler-factory.ts
```diff
@@ -278,6 +280,8 @@ function createStreamHandler<I extends Message<I>, O extends Message<O>>(
     const outputIt = pipe(
       req.body,
       transformPrepend<Uint8Array>(() => {
+        if (opt.requireConnectProtocolHeader)
```
smaye81, line 283: probably just my OCD but I like curly braces or at least this to be on the same line.  I sorta wish we enforced https://eslint.org/docs/latest/rules/nonblock-statement-body-position

bufdev:
@akshayjshah just making sure we're on the same page - is this a new part of the connect protocol?

akshayjshah:
Yes. It lets middle boxes like Envoy reliably identify unary Connect requests. It's optional (for back-compat), but servers may reject requests without it and generated clients are encouraged to set it. From the protocol spec:

> The Connect-Protocol-Version header distinguishes unary Connect RPC traffic from other requests that may use the same Content-Type. (In the future, it may also be used to support revisions to this protocol.) Clients, especially generated clients, should send this header. Servers and proxies may reject traffic without this header with an HTTP status code of 400.

bufdev:
Not so much for this PR, but will this be required? I assume it can't be since we hit v1.0

bufdev (approved):

packages/connect/src/protocol/universal-handler.ts
timostamm: TY, updated

packages/connect/src/protocol-connect/handler-factory.ts
timostamm: :hear_no_evil: updated

PR #521, opened 2023-03-08
Fix npm v9 packlist

timostamm:
npm v9 changed how it parses its include/ignore lists, see https://github.com/npm/cli/issues/5853#issuecomment-1317760279. 

Where it would previously include all files within the `dist` directory with the following setting:

```json
  "files": [
    "dist/**/"
  ]
```

In v9.5.0, this pattern matches nothing.

This led to some packages of release v0.8.2 to be published empty, which we addressed by publishing v0.8.2-1 of the affected packages (`@bufbuild/connect-node`, `@bufbuild/connect-fastify`, `@bufbuild/connect-express`, and `@bufbuild/connect-web`).

This PR updates the pattern to work with npm v9, verified by running `npm pack` with and without the change.

### DIFF
--- a/packages/connect-express/package.json
+++ b/packages/connect-express/package.json
@@ -34,6 +34,6 @@
     "@bufbuild/protobuf": "^1.0.0"
   },
   "files": [
-    "dist/**/"
+    "dist/**"
   ]
 }
--- a/packages/connect-fastify/package.json
+++ b/packages/connect-fastify/package.json
@@ -34,6 +34,6 @@
     "@bufbuild/protobuf": "^1.0.0"
   },
   "files": [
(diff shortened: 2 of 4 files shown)


smaye81:
Explains why `connect` and the plugins worked since they don't specify `files`.  But related question:  should they?

smaye81 (approved):

timostamm:
> Explains why `connect` and the plugins worked since they don't specify `files`. But related question: should they?

No. See [.npmignore](https://docs.npmjs.com/cli/v8/using-npm/developers#keeping-files-out-of-your-package) and how it relates to .gitignore and "files".

PR #525, opened 2023-03-12
Fix the gRPC User-Agent header name

timostamm:
The gRPC-web protocol uses `X-User-Agent`, but gRCP uses the standard header `User-Agent`.

The gRCP transports mistakenly set the `X-User-Agent` header, like the gRCP-web transport. This PR corrects that mistake to use `User-Agent` instead.


smaye81 (approved):
These changes seem to already be in the CORS export PR right?

timostamm:
> These changes seem to already be in the CORS export PR right?

Yes, did this in one go, but splitting it up so we have a comprehensible changelog.

PR #526, opened 2023-03-12
Export CORS constants for convenience

timostamm:
This adds the export `cors` to @bufbuild/connect. This object provides helpful constants to configure CORS middleware for cross-domain requests with the protocols supported by Connect.

### DIFF
--- a/packages/connect-node-test/src/helpers/testserver.ts
+++ b/packages/connect-node-test/src/helpers/testserver.ts
@@ -18,6 +18,7 @@ import * as https from "https";
 import * as fs from "fs";
 import * as path from "path";
 import type { Transport } from "@bufbuild/connect";
+import { cors } from "@bufbuild/connect";
 import {
   compressionGzip,
   connectNodeAdapter,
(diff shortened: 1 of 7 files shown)


fubhy:
Timo, wouldn't an approach around or similiar to interceptors on the ConnectRouter be a good place for handling CORS too? There are use-cases for CORS where you want to explicitly enable it for some service methods but not for others. Hence, tying it to the router seems reasonable instead of lifting it to the server / adapter level.

timostamm:
> Timo, wouldn't an approach around or similiar to interceptors on the ConnectRouter be a good place for handling CORS too? 

I think Connect routes should receive only requests in protocol. This export will make it really easy to configure @fastify/cors and the like properly. It's just a small incremental improvement.

smaye81 (approved):

fubhy:
> > Timo, wouldn't an approach around or similiar to interceptors on the ConnectRouter be a good place for handling CORS too?
> 
> I think Connect routes should receive only requests in protocol. This export will make it really easy to configure @fastify/cors and the like properly. It's just a small incremental improvement.

OK. The only limitation there would be that you wouldn't have the gRPC service method context outside of the router. But I can see wour point.

PR #531, opened 2023-03-14
Update HandlerContext signal

johynpapin:
This PR renames the AbortSignal property in the HandlerContext from `deadline` to a more generic `signal`. 

It also adds a signal to universal server requests and links it to the handler context signal. 

We expect server plugins to trigger the AbortSignal of the request when a request is finished or aborted prematurely. 

As a result, service implementations can now rely on the signal of the HandlerContext to clean up, whether the RPC timed out (via a deadline), whether the client canceled the request, or whether the request finished normally.

### DIFF
--- a/packages/connect-node/src/node-universal-handler.ts
+++ b/packages/connect-node/src/node-universal-handler.ts
@@ -89,12 +89,15 @@ export function universalRequestFromNodeRequest(
     parsedJsonBody !== undefined
       ? parsedJsonBody
       : asyncIterableFromNodeServerRequest(nodeRequest);
+  const abortController = new AbortController();
+  nodeRequest.on("close", () => abortController.abort());
   return {
     httpVersion: nodeRequest.httpVersion,
     method: nodeRequest.method ?? "",
     url: new URL(pathname, `${protocol}://${authority}`),
     header: nodeHeaderToWebHeader(nodeRequest.headers),
     body,
+    signal: abortController.signal,
   };
 }
 
--- a/packages/connect-node/src/node-universal-handler.ts
+++ b/packages/connect-node/src/node-universal-handler.ts
@@ -25,7 +25,10 @@ import {
   nodeHeaderToWebHeader,
   webHeaderToNodeHeaders,
 } from "./node-universal-header.js";
-import { connectErrorFromNodeReason } from "./node-error.js";
+import {
+  connectErrorFromH2ResetCode,
+  connectErrorFromNodeReason,
+} from "./node-error.js";
 
 /**
  * NodeHandlerFn is compatible with http.RequestListener and its equivalent
@@ -90,7 +93,20 @@ export function universalRequestFromNodeRequest(
       ? parsedJsonBody
       : asyncIterableFromNodeServerRequest(nodeRequest);
   const abortController = new AbortController();
-  nodeRequest.on("close", () => abortController.abort());
+  if ("stream" in nodeRequest) {
+    // HTTP/2 has error codes we want to honor
+    nodeRequest.on("close", () => {
+      const err = connectErrorFromH2ResetCode(nodeRequest.stream.rstCode);
+      if (err !== undefined) {
+        abortController.abort(err);
+      } else {
+        abortController.abort();
+      }
+    });
+  } else {
+    // HTTP/1.1 does not
+    nodeRequest.on("close", () => abortController.abort());
+  }
   return {
     httpVersion: nodeRequest.httpVersion,
     method: nodeRequest.method ?? "",
--- a/packages/connect/src/implementation.ts
+++ b/packages/connect/src/implementation.ts
@@ -73,7 +73,7 @@ export interface HandlerContext {
   readonly service: ServiceType;
 
   /**
-   * An AbortSignal to know when the connection with the client is closed.
+   * An AbortSignal that is aborted when the connection with the client is closed or when the deadline is reached.
    */
   readonly signal: AbortSignal;
 
--- a/packages/connect/src/implementation.ts
+++ b/packages/connect/src/implementation.ts
@@ -108,12 +108,10 @@ export function createHandlerContext(
   responseHeader: HeadersInit,
   responseTrailer: HeadersInit
 ): HandlerContext {
-  const linkedAbortcontroller = createLinkedAbortController(deadline, signal);
-
   return {
     method: spec.method,
     service: spec.service,
-    signal: linkedAbortcontroller.signal,
+    signal: createLinkedAbortController(deadline, signal).signal,
     requestHeader: new Headers(requestHeader),
     responseHeader: new Headers(responseHeader),
     responseTrailer: new Headers(responseTrailer),
--- a/packages/connect/src/implementation.ts
+++ b/packages/connect/src/implementation.ts
@@ -75,6 +75,11 @@ export interface HandlerContext {
   readonly deadline?: AbortSignal;
 
   /**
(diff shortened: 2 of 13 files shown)


timostamm:
Thank you for the PR, Johyn! We'll take a look soon.

johynpapin:
After reading #591, I wonder if we could just abort the `deadline` signal when the request is closed instead of adding another signal.

packages/connect-node/src/node-universal-handler.ts
```diff
@@ -89,12 +89,15 @@ export function universalRequestFromNodeRequest(
     parsedJsonBody !== undefined
       ? parsedJsonBody
       : asyncIterableFromNodeServerRequest(nodeRequest);
+  const abortController = new AbortController();
+  nodeRequest.on("close", () => abortController.abort());
```
timostamm, line 93: We have to honor HTTP/2 error codes here. We can use `connectErrorFromH2ResetCode()` added in https://github.com/bufbuild/connect-es/pull/619.

For example:

```ts
  if ("stream" in nodeRequest) {
    // HTTP/2 has error codes we want to honor
    nodeRequest.on("close", () => {
      const err = connectErrorFromH2ResetCode(nodeRequest.stream.rstCode);
      if (err) {
        ac.abort(err);
      } else {
        ac.abort();
      }
    });
  } else {
    // HTTP/1.1 does not
    nodeRequest.on("close", () => ac.abort());
  }
```

Yes, I expect the "close" event to always fire.

Aborting with a reason is not supported in Node.js v16, but it will be EOL soon. For the interim, it will be fine if v16 does not bubble up the code correctly.

We should update `universalServerRequestFromFetch()` as well, and use the signal from the fetch Request.

packages/connect/src/implementation.ts
```diff
@@ -95,13 +100,15 @@ export interface HandlerContext {
  */
 export function createHandlerContext(
   spec: { service: ServiceType; method: MethodInfo },
+  signal: AbortSignal,
```
timostamm, line 103: I am afraid I made this a bit more difficult with https://github.com/bufbuild/connect-es/pull/591 🙈 

Yes, it would be nice if we just had a single AbortSignal, but you need an Abort_Controller_ to trigger an AbortSignal, and I think `universalRequestFromNodeRequest()` should neither accept nor return an AbortController. That means we cannot avoid linking AbortSignals.

But it's not too bad: https://github.com/bufbuild/connect-es/pull/623 added `createLinkedAbortController()` for this purpose. We can merge the signal from the deadline / shutdown with the one from the request before passing it to the HandlerContext.

timostamm (commented):
Hey @johynpapin, took me a minute to get back to this :blush:

Adding the signal to the UniversalServerRequest, aborting it in node-universal-handler.ts, and making it available in the HandlerContext is a good call. 

There is one small but important bit though: HTTP/2 has a set of error codes ([spec](https://httpwg.org/specs/rfc7540.html#ErrorCodes)), and gRPC defines a mapping between these codes, and gRPC error codes ([spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#errors)). If a HTTP/2 client cancels a request, it should close the stream with the HTTP/2 code CANCEL (implemented in https://github.com/bufbuild/connect-es/pull/552). The server should pick this code up, and convert it to a Connect/gRPC code "canceled".

If you can spare the time to update the PR, we'd be super happy to have it 🙂

johynpapin:
Hey @timostamm! :wave:

Thank you very much for this detailed answer. I suspected that there was a problem with the error handling, it's clearer to me now.

I will take the time to understand and make these changes! :smile:

packages/connect/src/implementation.ts
johynpapin: If I understand correctly it is necessary to wait until #623 is merged to implement this change? :open_mouth:

timostamm: Yes, https://github.com/bufbuild/connect-es/pull/623 should be merged soon 🙂

packages/connect/src/protocol/universal-fetch.ts
```diff
@@ -91,6 +91,7 @@ function universalServerRequestFromFetch(
     url,
     header: req.headers,
     body: iterableFromReadableStream(req.body),
+    signal: req.signal,
```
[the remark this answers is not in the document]
timostamm: I believe it is the best we can do. For context, Deno has been starting to make use of the property on servers in https://github.com/denoland/deno/pull/17822.

packages/connect/src/protocol/universal-handler-client.ts
```diff
@@ -44,6 +44,7 @@ export function createUniversalHandlerClient(
       method: uClientReq.method,
       url: reqUrl,
       header: uClientReq.header,
+      signal: uClientReq.signal ?? new AbortController().signal,
```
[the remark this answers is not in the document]
timostamm: Yes, I believe this should work. We will add more test coverage for this function soon, and can fix up if necessary.

packages/connect/src/protocol/universal.ts
```diff
@@ -64,6 +64,7 @@ export interface UniversalServerRequest {
    * handled efficiently.
    */
   body: AsyncIterable<Uint8Array> | JsonValue;
+  signal: AbortSignal;
```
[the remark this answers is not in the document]
timostamm: I think it's fair to use just `signal`, like the fetch API Request 👍

packages/connect/src/implementation.ts
```diff
@@ -99,14 +102,17 @@ export interface HandlerContext {
...
   responseHeader: HeadersInit,
   responseTrailer: HeadersInit
 ): HandlerContext {
+  const linkedAbortcontroller = createLinkedAbortController(deadline, signal);
+
   return {
     method: spec.method,
     service: spec.service,
-    deadline,
+    signal: linkedAbortcontroller.signal,
```
timostamm, line 115: Good idea to link the signals here 👍 
Can you simply put `createLinkedAbortController(deadline, signal)` on this line?

packages/connect/src/implementation.ts
```diff
@@ -71,8 +72,10 @@ export interface HandlerContext {
    */
   readonly service: ServiceType;
 
-  // TODO
-  readonly deadline?: AbortSignal;
+  /**
+   * An AbortSignal that is aborted when the connection with the client is closed or when the deadline is reached.
+   */
```
timostamm, line 77: We wrap JSDoc at 80 columns:

```suggestion
  /**
   * An AbortSignal that is aborted when the connection with the client is closed 
   * or when the deadline is reached.
   */
```

timostamm (commented):
Looks great! 

Do not worry about the CI failure. The fork does not have access to the secret that is required for browserstack. As long as `make` passes (as it did), everything is fine, since we are not touching code that directly affects @bufbuild/connect-web.

I'm going to think about test coverage tomorrow morning and get back to you.

packages/connect/src/implementation.ts
johynpapin: Yes!

timostamm (approved):
Thank for for the contribution, Johyn, LGTM! 

I'm going to update the PR description, and also update the description of https://github.com/bufbuild/connect-es/pull/591 for the renamed HandlerContext property.

timostamm (approved):
Thank for for the contribution, Johyn, LGTM! 

I'm going to update the PR description, and also update the description of https://github.com/bufbuild/connect-es/pull/591 for the renamed HandlerContext property.

PR #532, opened 2023-03-14
Fix socket connect issue in Node 19

smaye81:
Node 19 turns on `keepAlive` by default when using a global `Agent` for HTTP/1.1.  This in turn causes our sockets to be pooled and reused.  

We issue the request in the socket's `connect` event, but, since the sockets are being reused, the `connect` event doesn't fire.  So, the request is never sent and the socket times out.

This checks the socket's `readyState` to see if it is `open` and if so, it simply issues the request. Otherwise, it binds the even listener to the socket's `connect` event.

Checking `readyState` should be the best way to verify the socket's connection. See code [here](https://github.com/nodejs/node/blob/4ae7c7a00258cd0ac2bb344afcaa57a2b2d2ed3f/lib/net.js#LL652C26-L652C26).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,7 @@ GEN   = .tmp/gen
 CROSSTEST_VERSION := 2c228a09c52e3a95263034c1fb79119d33ab3258
 LICENSE_HEADER_YEAR_RANGE := 2021-2023
 LICENSE_IGNORE := -e .tmp\/ -e node_modules\/ -e packages\/.*\/src\/gen\/ -e packages\/.*\/dist\/ -e scripts\/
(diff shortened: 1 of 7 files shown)


timostamm (commented):
That means we can lift the engines field constraint `<19` now, right (package.json)? We should also run our tests on v19 now (Makefile).

smaye81:
> That means we can lift the engines field constraint `<19` now, right (package.json)? We should also run our tests on v19 now (Makefile).

Good call.  We can yeah.  Added those changes to this PR.

timostamm (approved):

PR #543, opened 2023-03-16
Enable exactOptionalPropertyTypes

dimitropoulos:
To quote from the TypeScript team lead:

> If we were to start TypeScript over again, we believe the behavior of strictOptionalProperties would be on by default
> https://github.com/microsoft/TypeScript/issues/44421

This [configuration](https://www.typescriptlang.org/tsconfig#exactOptionalPropertyTypes) brings a pretty critical element to the type-soundness (critical because it drastically changes the behavior of operations that check for keys). [SameValueZero](https://262.ecma-international.org/7.0/#sec-samevaluezero) checks remain unchanged, anyway, so there's basically no downside to having this rule enabled. The https://github.com/microsoft/TypeScript/pull/43947 from Anders provides some more context, as well.

Thankfully, this codebase was already pretty close to the mark, with only a few little minor things to update to increase strictness. Note a similar PR: https://github.com/bufbuild/protobuf-es/pull/431

### DIFF
--- a/packages/connect/src/protocol-connect/end-stream.ts
+++ b/packages/connect/src/protocol-connect/end-stream.ts
@@ -34,7 +34,7 @@ export const endStreamFlag = 0b00000010;
  */
 export interface EndStreamResponse {
   metadata: Headers;
-  error?: ConnectError;
+  error?: ConnectError | undefined;
 }
 
 /**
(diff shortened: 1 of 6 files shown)


timostamm:
Let's enable exactOptionalPropertyTypes when it becomes the default with TSC.

PR #545, opened 2023-03-24
Document how to use multiple plugin options

timostamm:
It is not obvious how to give multiple plugin options from our readme. This updates the example buf.gen.yaml to use a yaml list, and adds a comment.


srikrsna-buf (approved):
Wring account!

PR #548, opened 2023-03-28
Add a `ConnectRouter` based `Transport`

srikrsna-buf:
Add a `UniversalHandler` based http client. Also add `createRoutesTransport`. 

This can be used to create mock/stub handlers for testing client and also to test backends without needed to run a server.

### DIFF
--- a/packages/connect/src/index.ts
+++ b/packages/connect/src/index.ts
@@ -60,5 +60,4 @@ export {
   createMethodImplSpec,
 } from "./implementation.js";
 export type { ServiceImplSpec, MethodImplSpec } from "./implementation.js";
-
-export { createRouterHttpClient } from "./router-http-client.js";
+export { createTransport } from "./protocol-connect/transport.js";
--- a/packages/connect/src/index.ts
+++ b/packages/connect/src/index.ts
@@ -60,3 +60,5 @@ export {
   createMethodImplSpec,
 } from "./implementation.js";
 export type { ServiceImplSpec, MethodImplSpec } from "./implementation.js";
+
+export { createRouterHttpClient } from "./router-http-client.js";
--- a/packages/connect/src/index.ts
+++ b/packages/connect/src/index.ts
@@ -60,4 +60,4 @@ export {
   createMethodImplSpec,
 } from "./implementation.js";
 export type { ServiceImplSpec, MethodImplSpec } from "./implementation.js";
-export { createTransport } from "./protocol-connect/transport.js";
+export { createRouterTransport } from "./router-transport.js";
--- a/packages/connect/src/protocol/universal-handler-client.ts
+++ b/packages/connect/src/protocol/universal-handler-client.ts
@@ -12,25 +12,21 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import type { ConnectRouter } from "./router.js";
-import type { UniversalClientFn } from "./protocol/universal.js";
-import type { UniversalHandler } from "./protocol/universal-handler.js";
-import { ConnectError } from "./connect-error.js";
-import { createAsyncIterable } from "./protocol/async-iterable.js";
-import { Code } from "./code.js";
+import { Code } from "../code.js";
+import { ConnectError } from "../connect-error.js";
+import { createAsyncIterable } from "./async-iterable.js";
+import type { UniversalHandler } from "./universal-handler.js";
+import type { UniversalClientFn } from "./universal.js";
 
 /**
- * An in-memory {@link UniversalClientFn | http client} that can be used to route requests to a {@link ConnectRouter}
- * by passing network calls. Useful for testing and calling in-process services.
- *
- * @param router The {@link ConnectRouter} to use.
- * @returns The {@link UniversalClientFn} that can be passed to `createTransport` functions.
+ * An in-memory UniversalClientFn that can be used to route requests to a ConnectRouter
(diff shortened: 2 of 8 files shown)


packages/connect/src/router-http-client.ts
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * An in-memory {@link UniversalClientFn | http client} that can be used to route requests to a {@link ConnectRouter}
+ * by passing network calls. Useful for testing and calling in-process services.
+ *
+ * @param router The {@link ConnectRouter} to use.
+ * @returns The {@link UniversalClientFn} that can be passed to `createTransport` functions.
+ */
+export function createRouterHttpClient(
+  router: ConnectRouter
+): UniversalClientFn {
```
timostamm, line 31: This is a nice function - doesn't do too much, but enables cool features 👍 

But I think we should still have a function on top that returns a `Transport`. This will be easier to explain,  gives us the leeway to add features (e.g. skip serialization, like we discussed), and doesn't require the user to import from @bufbuild/connect/protocol-* (which I would like to discourage).

WDYT about accepting `handlers: UniversalHandler[]` instead of `router: ConnectRouter` and moving it to `src/protocol/router-client.ts`? That way, we keep the exports from @bufbuild/connect to a minimum.

packages/connect/src/router-http-client.ts
```diff
@@ -0,0 +1,69 @@
...
+    const handler = handlerMap.get(reqUrl.pathname);
+    if (!handler) {
+      throw new ConnectError(
+        `RouterHttpClient: no handler registered for ${reqUrl.pathname}`,
+        Code.Unimplemented
+      );
+    }
+    const uServerRes = await handler({
+      body: uClientReq.body,
+      httpVersion: "1.1",
```
timostamm, line 47: For bidi-streaming, all clients will run into this code path on the handler: https://github.com/bufbuild/connect-es/blob/main/packages/connect/src/protocol/universal-handler.ts#L244-L247

Let's pass `2.0` here. We have to keep in mind that we might want to have this configurable at some point.

packages/connect/src/router-http-client.ts
```diff
@@ -0,0 +1,69 @@
...
+import { ConnectError } from "./connect-error.js";
+import { createAsyncIterable } from "./protocol/async-iterable.js";
+import { Code } from "./code.js";
+
+/**
+ * An in-memory {@link UniversalClientFn | http client} that can be used to route requests to a {@link ConnectRouter}
+ * by passing network calls. Useful for testing and calling in-process services.
+ *
+ * @param router The {@link ConnectRouter} to use.
+ * @returns The {@link UniversalClientFn} that can be passed to `createTransport` functions.
```
timostamm, line 27: The entire codebase skips JSDoc annotations. Unless you want to add them everywhere (which I'm not opposed to), it's best to remove them for consistency.

timostamm (commented):
Some test coverage would be nice. Also see comments below:

packages/connect/src/router-http-client.ts
srikrsna-buf: I'll move it to protocol, I had the router in mind so I thought it was not possible but just using handlers makes sense to me.

I also wanted a Transport but was unsure where it will go as it will be tied to the protocol in its current form. Should we make an entirely new transport for this one or just use connect?

timostamm: I agree that it is odd to pick one of the protocols, but picking Connect seems reasonable. We have to create the ConnectRouter in the transport constructor function though for this to work, and cannot accept an existing ConnectRouter because createConnectRouter() takes options to enable/disable protocols.

I'm fine with adding a function that creates the Transport in this PR, or in a follow-up.

packages/connect/src/index.ts
```diff
@@ -60,3 +60,4 @@ export {
   createMethodImplSpec,
 } from "./implementation.js";
 export type { ServiceImplSpec, MethodImplSpec } from "./implementation.js";
+export { createTransport } from "./protocol-connect/transport.js";
```
timostamm, line 63: Shouldn't this export `createRoutesTransport` instead?

packages/connect/src/protocol/universal-handler-client.ts
```diff
@@ -0,0 +1,65 @@
...
+      method: uClientReq.method,
+      url: reqUrl,
+      header: uClientReq.header,
+    });
+    if (uServerRes.body === undefined) {
+      throw new ConnectError(
+        `RouterHttpClient: missing body in handler response for ${reqUrl.pathname}`,
+        Code.Internal
+      );
+    }
```
timostamm, line 53: We will hit this code path if the user sends a request with an unsupported compression algorithm. I think we're better off by falling back to an empty byte array here.

srikrsna-buf: Shouldn't that result in a `ConnectError`?

timostamm: Yes, but one that's created by the client from an HTTP status. 

I think I lied, it doesn't happen with compression, but with gRPC(-web) trailers-only responses which are actually headers-only responses. Either way, we shouldn't error on an empty body.

timostamm (approved):
Nice!

dimitropoulos (approved):
really ran this code through its paces with https://github.com/bufbuild/connect-query/commits/dimitri/mock-transport and it's working amazingly!

PR #553, opened 2023-03-30
Fix package definitions for Connect

### LINKED ISSUE #544: error codes are not preserved
**Describe the bug**

I am emitting an error from `connect-go` like so:
```go
connectgo.NewError(connectgo.CodeFailedPrecondition, errors.New("Service is disabled."))
```

The server is emitting the error correctly but when the [transport handler calls connectFromReason](https://github.com/bufbuild/connect-es/blob/8a8fdbd5970094397c2f74d5aad54ea67e47b4df/packages/connect-web/src/connect-transport.ts#L309), the `instanceof` check is failing and a new error is being created with the default code `Internal`.

See debugger output below. The error should have been returned on line 98 but instead the `instanceof` check fails and the code continues to create a new error.

Screenshot from 2023-03-16 12-35-57

Here are more details about the error:
Screenshot from 2023-03-16 12-41-55

I did some research and from what I can tell this should ensure that we can correctly determine what the class is but it seems not to be working as expected: https://github.com/bufbuild/connect-es/blob/8a8fdbd5970094397c2f74d5aad54ea67e47b4df/packages/connect/src/connect-error.ts#L92

npm list
```npm list
├── @bufbuild/connect-web@0.8.3
├── @bufbuild/protobuf@1.1.0
├── @bufbuild/protoc-gen-connect-web@0.8.3
├── @bufbuild/protoc-gen-es@1.1.0
├── @parcel/config-webextension@2.8.2
├── @primer/octicons-react@17.12.0
├── @primer/octicons@17.12.0
├── @types/react-dom@18.0.10
├── @types/react@18.0.26
├── @types/webextension-polyfill@0.9.2
├── @typescript-eslint/eslint-plugin@5.48.0
├── @typescript-eslint/parser@5.48.0
├── assert@2.0.0
├── axios@1.2.2
├── babel-plugin-macros@3.1.0
├── buffer@6.0.3
├── delay@5.0.0
├── dom-loaded@3.0.0
├── element-ready@6.2.1
├── eslint-config-airbnb@19.0.4
├── eslint-config-preact@1.3.0
├── eslint-config-prettier@8.6.0
├── eslint-plugin-prettier@4.2.1
├── eslint@8.31.0
├── github-url-detection@6.1.0
├── jwt-decode@3.1.2
├── lightningcss@1.19.0
├── os-browserify@0.3.0
├── parcel@2.8.2
├── path-browserify@1.0.1
├── prettier@2.8.1
├── process@0.11.10
├── react-dom@18.2.0
├── react-timeago@7.1.0
├── react-tooltip@5.8.3
├── react@18.2.0
├── select-dom@7.1.1
├── styletron-engine-atomic@1.5.0
├── styletron-engine-monolithic@1.0.0
├── styletron-react@6.1.0
├── styletron@3.0.4
├── typescript@4.9.4
├── util@0.12.5
├── web-ext@7.4.0
├── webext-base-css@1.4.3
├── webext-options-sync@4.0.0
└── webextension-polyfill@0.10.0
```

- Node.js version: v18.14.2
- Browser and version: Google Chrome 111.0.5563.64

smaye81:
Fixes #544

This fixes the workaround for bundlers/module loaders that do not support subpath exports by changing the Connect package to be a CJS module and updating the workaround to use CJS.

Previously, for bundlers/module loaders that did not support subpath exports, they would fallback to our workarounds, which were using ESM. However, for any client using CJS, this was leading to the [dual package hazard](https://nodejs.org/api/packages.html#dual-package-hazard) where the client loaded an entrypoint of CJS for Connect-Web and the fallback to ESM for Connect caused both CJS and ESM to run together. This led to issues where the `instanceof ConnectError` calls were returning false, which in turn caused Connect to wrap Connect errors in another Connect error, which was very confusing to users.

Note that we are now adding a dummy package.json to the `connect/dist/esm` directory with a `type` of `module` so that consumers can import from that directory as ESM modules. However, adding this file caused our bundle to inflate. This is because esbuild checks for a `sideEffects` map in the nearest enclosing directory with a package.json file ([source](https://github.com/evanw/esbuild/blob/main/internal/resolver/resolver.go#L661-L671)). Our root directory already had this present, but once we started adding this new package.json, it's lack of a sideEffects map overrode the root file, which caused additional files to be added to the bundle. So, the code was updated to add this dummy package.json with an additional `sideEffects: false` property.

Tested on all projects in connect-es-integration to verify it works appropriately for different bundlers and different module systems.


packages/connect/protocol-connect.js
```diff
@@ -12,5 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// workaround for bundlers that do not support subpath exports.
-export * from "./dist/esm/protocol-connect/index.js";
+/* eslint-disable */
```
timostamm, line 15: Instead of disabling eslint completely, please use `eslint-disable-line` or `eslint-disable-next-line`.

packages/connect/protocol-connect.js
```diff
@@ -12,5 +12,7 @@
...
+
+// Workaround for bundlers that do not support subpath exports.
+module.exports = require("./dist/cjs/protocol-connect/index.js");
```
timostamm, line 18: Why is this changed, but the other files are not?

packages/connect/protocol-connect.js
smaye81: @timostamm there are multiple files in `connect-web-test` that also use this directive (`packages/connect-web-test/karma-fixup-queueMicrotask.js` for example). should I change those also?

packages/connect/protocol-connect.js
smaye81: tbh, it's very confusing.  I think eslint is seeing these files as ESM and the `module.exports` is causing it to fail. If I change these files to have a `.cjs` extension, they pass. I borrowed the `.js` + `eslint-disable` from the Karma files I mentioned in the above comment.

packages/connect/protocol-connect.js
timostamm: Unless it turns out to need a linter disable on every second line or so, yes.

packages/connect/protocol-connect.js
timostamm: > I think eslint is seeing these files as ESM and the module.exports is causing it to fail. If I change these files to have a .cjs extension, they pass. 

The reason is that we run eslint from the root of the repository, where a package.json with type:module sits. This makes eslint believe all `.js` files should be ESM.

We can not change the file extension to `.cjs`. The correct fix would be to add an override for `.js` in this path. Maybe @dimitropoulos can help set this up? 

If that isn't possible, please update the linter comment to explain the situation.

smaye81: If you'd rather it be an override (instead of my use of `/* global` above) that works too. Totally cool if @dimitropoulos wants to set that up.

smaye81:
@timostamm I believe the bundle issue is fixed. I added a summary in the PR description.

timostamm (commented):
Looks like a reasonable solution to me 👍 

Do we have confidence that this still fixes https://github.com/bufbuild/connect-es/issues/544? IIRC we switched protocol.js to CommonJS late in the game, which might change behaviour.

smaye81:
> Looks like a reasonable solution to me 👍
> 
> Do we have confidence that this still fixes #544? IIRC we switched protocol.js to CommonJS late in the game, which might change behaviour.

Yep, I tested this with a few different bundlers in connect-es-integration (Parcel, Webpack4, etc.)

packages/connect/protocol-connect.js
dimitropoulos: ok: I updated it in dca59a36ebe1a286c2337d3878e6f606093070e2

For whatever it's worth, I personally would prefer to leave it as is because there's no feedbacking check that if these files are ever removed that this line from the eslint config will also be removed.  I don't mind that it's duplicated on a per-file basis because at least then the override can't be accidentally applied to other files that match the same pattern and also you can clearly see why it works the way it does just be seeing the one source file (instead of having to know there's a specific override just for these files in the root config).

But, it's all the same if that's what you guys think is best.  no problemo.

timostamm: I thought we would be able to set `sourceType: "commonjs"` for the offending files. I can't imagine that every CommonJS projects needs to set globals manually.  Does that not work?

dimitropoulos: it does not work, yeah.  You still get `no-undef` triggering if you write the config like this:

```json
    {
      files: ["packages/connect/*.js"],
      parserOptions: {
        sourceType: "commonjs",
      },
    },
```

timostamm: Thanks for looking into it, @dimitropoulos.

dimitropoulos: happy to help.. although... haha... I'm suddenly now quite curious why `sourceType` _doesn't_ work.  Feels like it should.  Oh well.  We have dozens of bigger fish to fry so I'll just hope to have the spontaneous "ah-ha" moment on that one in the future.

PR #588, opened 2023-04-19
Fix UniversalHandlers resolving the response before headers are complete

timostamm:
https://github.com/bufbuild/connect-es/pull/582 and https://github.com/bufbuild/connect-es/pull/575 and surfaced an issue with the `UniversalHandler` implementations for the Connect, gRPC, and gRCP-web protocols: The function returns a promise for a `UniversalServerResponse` that resolves immediately, before the RPC implementation is invoked.

This causes some server implementations (fetch API handlers, and `createRouterTransport`, but _not_ `@bufbuild/connect-node`) to write the response headers before the RPC implementation had a chance to add its own response headers.

The fix is to wait until the first response body byte is produced before resolving the response.

### DIFF
--- a/packages/connect/src/protocol-connect/handler-factory.spec.ts
+++ b/packages/connect/src/protocol-connect/handler-factory.spec.ts
@@ -12,28 +12,30 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import {
-  Int32Value,
-  Message,
-  MethodKind,
-  StringValue,
-} from "@bufbuild/protobuf";
-import type { MethodInfo, ServiceType } from "@bufbuild/protobuf";
+import type { MethodInfo } from "@bufbuild/protobuf";
+import { Int32Value, MethodKind, StringValue } from "@bufbuild/protobuf";
 import { createHandlerFactory } from "./handler-factory.js";
+import type { MethodImpl } from "../implementation.js";
 import { createMethodImplSpec } from "../implementation.js";
-import type { HandlerContext, MethodImpl } from "../implementation.js";
 import type { UniversalHandlerOptions } from "../protocol/index.js";
-import { errorFromJsonBytes } from "./error-json.js";
+import {
+  createAsyncIterable,
+  createUniversalHandlerClient,
+  pipeTo,
+  sinkAll,
+} from "../protocol/index.js";
 import { ConnectError } from "../connect-error.js";
-import { Code } from "../code.js";
-import { endStreamFromJson } from "./end-stream.js";
 import {
   createAsyncIterableBytes,
   readAllBytes,
 } from "../protocol/async-iterable-helper.spec.js";
+import { Code } from "../code.js";
+import { errorFromJsonBytes } from "./error-json.js";
+import { endStreamFromJson } from "./end-stream.js";
+import { createTransport } from "./transport.js";
 
 describe("createHandlerFactory()", function () {
-  const testService: ServiceType = {
+  const testService = {
     typeName: "TestService",
     methods: {
       foo: {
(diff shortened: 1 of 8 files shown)


packages/connect/src/protocol-connect/handler-factory.spec.ts
```diff
@@ -51,51 +53,89 @@ describe("createHandlerFactory()", function () {
...
+      );
+      const r = await transport.stream(
+        service,
+        method,
+        undefined,
+        undefined,
+        undefined,
+        createAsyncIterable([new Int32Value({ value: 123 })])
+      );
+      expect(r.header.get("implementation-called")).toBe("yes");
```
timostamm, line 123: We are testing that response headers set by the implementation are present in the response received by the client, in all three protocols, for unary and for server-streaming.

packages/connect/src/protocol-connect/handler-factory.ts
```diff
@@ -327,7 +328,7 @@ function createStreamHandler<I extends Message<I>, O extends Message<O>>(
     );
     return {
       ...uResponseOk,
-      body: outputIt,
+      body: await untilFirst(outputIt),
```
timostamm, line 331: Not pretty, but I'm not sure how we could improve this.

packages/connect/src/protocol-connect/handler-factory.spec.ts
```diff
@@ -51,51 +53,89 @@ describe("createHandlerFactory()", function () {
     },
   } as const;
```
dimitropoulos, line 54: suggestion: if we want the best of both worlds (having the type contract and also a constant), we can do

```suggestion
  } satisfies ServiceType;
```

packages/connect/src/protocol-connect/handler-factory.spec.ts
```diff
@@ -51,51 +53,89 @@ describe("createHandlerFactory()", function () {
...
 
-  function stub<M extends MethodInfo>(
-    opt: {
-      service?: ServiceType;
-      method?: M;
-      impl?: MethodImpl<M>;
-    } & Partial<UniversalHandlerOptions>
+  function setupTestHandler<M extends MethodInfo>(
```
dimitropoulos, line 56: question: I'm wondering if you think there's anything here that's worth folding into `creatRouterTransport` instead?  seems fine like it is, but just would like to hear your thoughts on whether this is really a one-off that one we (in our codebases) would care about and not something a typical user would ever need.

packages/connect/src/protocol-connect/handler-factory.spec.ts
```diff
@@ -51,51 +53,89 @@ describe("createHandlerFactory()", function () {
...
   }
 
+  describe("returned handler", function () {
+    it("should surface headers for unary", async function () {
+      const { transport, service, method } = setupTestHandler(
+        testService.methods.foo,
+        {},
+        (req, ctx) => {
+          ctx.responseHeader.set("implementation-called", "yes");
+          return { value: req.value.toString(10) };
```
dimitropoulos, line 90: question: just for my own learning, I'm curious what value it brings to do `toString(10)` when the default is already 10.  My first reaction is that it's worth it to be explicit because there are other situations where we use other radices, but I searched the codebase and didn't see any others.   just curious.

packages/connect/src/protocol-connect/handler-factory.spec.ts
dimitropoulos: looks great.  I like that we're also checking that the header value is correct at the same time.  The current approach used in `interceptor.spec.ts` should probably be improved to do the same thing because right now the "`log` vs `wantLog`" implementation just checks header presence without inspecting the values.

packages/connect/src/protocol-connect/handler-factory.spec.ts
```diff
@@ -12,28 +12,30 @@
...
+import { Code } from "../code.js";
+import { errorFromJsonBytes } from "./error-json.js";
+import { endStreamFromJson } from "./end-stream.js";
+import { createTransport } from "./transport.js";
 
 describe("createHandlerFactory()", function () {
-  const testService: ServiceType = {
+  const testService = {
     typeName: "TestService",
     methods: {
```
dimitropoulos, line 40: thought: would this test suite be an appropriate place to also have a bidi test (right now it's just unary and server-streaming)?

packages/connect/src/protocol/async-iterable.ts
```diff
@@ -1148,6 +1149,42 @@ function parseLengthHint(
   return [true, n];
 }
 
+/**
+ * Wait for the first element of an iterable without modifying the iterable.
+ * This consumes the first element, but pushes it back on the stack.
+ *
+ * @private Internal code, does not follow semantic versioning.
+ */
+export async function untilFirst<T>(
```
dimitropoulos, line 1158: suggestion: of course, we're exercising this indirectly by the means of the other tests in this PR, but it'd be super great to have some coverage for this function because it's a bit esoteric what it actually does and if we ever have to come back here and adjust it it'd be pretty hard.

If you're interested: I'd love to help with writing that test if you're interested in pairing on it!

packages/connect/src/protocol-grpc-web/handler-factory.ts
```diff
@@ -175,7 +175,7 @@ function createHandler<I extends Message<I>, O extends Message<O>>(
     );
     return {
       ...uResponseOk,
-      body: outputIt,
```
dimitropoulos, line 178: suggestion: I'm not sure if this suggestion would be strange for this codebase but it'd be nice to leave a comment here (or somewhere?  maybe the untilFirst JSDoc itself?) explaining why this is necessary and maybe some detail about how node's behavior differs from something like cloudflare workers.

packages/connect/src/protocol-grpc-web/handler-factory.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+    return {
+      service: testService,
+      method: method,
+      handler: h,
+      transport: t,
+    };
+  }
+
+  describe("returned handler", function () {
+    it("should surface headers for unary", async function () {
```
dimitropoulos, line 75: I really like that after reading and understanding one test, you've read and understood them all!

packages/connect/src/protocol-connect/handler-factory.spec.ts
dimitropoulos: (after reading the rest of the PR, I see it's copied 3 times: if we don't upgrade createRouterTransport in some way to support the same use-cases, maybe it'd be good to extract to somewhere so the implementations don't drift?)

dimitropoulos (approved):
https://github.com/bufbuild/connect-es/pull/590 passing CI confirms that this works!  None of my feedback is super blocking.

packages/connect/src/protocol-connect/handler-factory.ts
dimitropoulos: suggestion: I'm not sure if this suggestion would be strange for this codebase but it'd be nice to leave a comment here (or somewhere? maybe the untilFirst JSDoc itself?) explaining why this is necessary and maybe some detail about how node's behavior differs from something like cloudflare workers.

dimitropoulos (commented):
update: inspired by https://github.com/bufbuild/connect-es/pull/588/files#r1171944922 I went ahead and checked the header values the same way for the interceptor tests in 0526f9ad8719a42b4c15a0fa76dafd49679d5df8 and it doesn't pass, but I'm not sure why?  Could there be a sneaky bug somewhere still, or did I just do something wrong?

packages/connect/src/protocol-connect/handler-factory.spec.ts
timostamm: Yep, forgot this one, the other two have it already. We have to be cautious with `satisfies`, we promise compat with TypeScript 4.1. We'll have to add ts-compat checks like in protobuf-es if we start marking more widespread use.

packages/connect/src/protocol-connect/handler-factory.spec.ts
timostamm: Same code, but not identical, because it uses different imports. I only put the headers tests here because it was the least effort to get some coverage without updating our test infra. I think we're better off by testing the router transport the same way we test other transports (crosstests).

packages/connect/src/protocol-connect/handler-factory.spec.ts
timostamm: This would become a very hard to maintain x-thousand LoC test file if it tried to get decent coverage, and we would still need coverage for running on an actual network with Node.js etc. See comment above about crosstests. They don't cover this case for bidi yet though.

packages/connect/src/protocol-connect/handler-factory.spec.ts
timostamm: Wouldn't hurt, yes.

packages/connect/src/protocol-connect/handler-factory.spec.ts
timostamm: Just to be explicit, but I wouldn't say it's the most important thing in the world. There should be a couple of other places where it's used.

packages/connect/src/protocol-connect/handler-factory.ts
timostamm: Added explanatory comments in https://github.com/bufbuild/connect-es/pull/588/commits/6c6dce8fd0dea0acdc633175900224bbea49c711.

packages/connect/src/protocol/async-iterable.ts
timostamm: Won't say no to a couple of concise tests for this, but we'll get better mileage from improving the crosstests. Not saying that they are a replacement unit tests for this function, but that there is more pressing need.

timostamm:
> I went ahead and checked the header values the same way for the interceptor tests in [0526f9a](https://github.com/bufbuild/connect-es/commit/0526f9ad8719a42b4c15a0fa76dafd49679d5df8) and it doesn't pass, but I'm not sure why? 

Looks like you set "stream-response-header", but expect "unary-response-header".

dimitropoulos:
> Looks like you set "stream-response-header", but expect "unary-response-header".

oh phew!  thanks!!

PR #589, opened 2023-04-19
Add tests for the Vercel edge runtime

timostamm:
This runs all crosstests against connect-es handlers running in the `edge-runtime` package ([on npmjs.com](https://www.npmjs.com/package/edge-runtime), documented [here](https://edge-runtime.vercel.app/packages/runtime)). These are similar, but not identical conditions to running on the Vercel platform.

This is not ready to be merged yet, because we are missing an important part: Outgoing requests. We have to find a way to run the clients in the environment, not just handlers.


timostamm:
This has become stale. We've started adding conformance tests for the cloudflare. If we continue work on this, we should pick it up from there.

PR #598, opened 2023-04-29
Always generate files when no services exist

### LINKED ISSUE #597: protoc-gen-connect-es: keep_empty_files=true doesn't emit some empty files
**Describe the bug**

Using `protoc-gen-conect-es` with `--connect-es_opt=keep_empty_files=true` doesn't produce an empty output file when a service-less input file is provided. This unfortunately makes Bazel integration really difficult.

**To Reproduce**

Neither `protoc` nor the `protoc-gen-connect-es` produce an error message — this presents as a silent lack of an output file.

Using this reduced version of `eliza.proto` (`no-service.proto` in this example):

```proto
// no-service.proto
syntax = "proto3";

message SayRequest {
  string sentence = 1;
}

message SayResponse {
  string sentence = 1;
}

// Note: no 'service ElizaService' exists here
```

and running:

```sh
mkdir -p out
protoc \
  --plugin=protoc-gen-connect-es=./bin/protoc-gen-connect-es \
  --connect-es_out=target=ts,keep_empty_files=true:./out \
  ./no-service.proto
```

results in a still-empty `out/` directory, though I'd expect an empty file to exist at `out/no-service_connect.ts`.

The same appears to repro regardless of the value of the `target` option, for what it's worth.

- `@bufbuild/protoc-gen-connect-es` version: Current tip of `master` (0.8.6 @ f6bafee (Simplify options of the Node.js HTTP client (#594), 2023-04-27))
- `protoc` version: libprotoc 3.14.0
- Node.js version: v16.20.0

**Additional context**
I tracked the problem down to the early-continue before calling `schema.generateFile()` in `typescript.ts` (and the equivalents in `declaration.ts` and `javascript.ts`): https://github.com/bufbuild/connect-es/blob/f6bafee56893eb35a3c60a4d6ed00696bf88af02/packages/protoc-gen-connect-es/src/typescript.ts#L26-L29

Moving the `schema.generateFile()` before the `if (profoFiles.services.length == 0)` check seems to fix it. ~I'm happy to put up a PR, just waiting for my legal team to clear me signing the CLA 😃~ Legal cleared me: https://github.com/bufbuild/connect-es/pull/598

sjbarag:
Previously, protoc-gen-connect-es would not emit a file when an input .proto file had no services declared within it. This unfortunately broke the keep_empty_files option, as there was never an empty file to keep. Ensure a file is always generated for each input file, allowing keep_empty_files to determine if it should remain before exiting.

fixes #597

### DIFF
--- a/packages/protoc-gen-connect-es/src/declaration.ts
+++ b/packages/protoc-gen-connect-es/src/declaration.ts
@@ -23,9 +23,6 @@ import {
 
 export function generateDts(schema: Schema) {
   for (const protoFile of schema.files) {
-    if (protoFile.services.length == 0) {
-      continue;
-    }
     const file = schema.generateFile(protoFile.name + "_connect.d.ts");
     file.preamble(protoFile);
     for (const service of protoFile.services) {
--- a/packages/protoc-gen-connect-es/src/javascript.ts
+++ b/packages/protoc-gen-connect-es/src/javascript.ts
@@ -23,9 +23,6 @@ import {
 
 export function generateJs(schema: Schema) {
   for (const protoFile of schema.files) {
-    if (protoFile.services.length == 0) {
-      continue;
-    }
     const file = schema.generateFile(protoFile.name + "_connect.js");
     file.preamble(protoFile);
     for (const service of protoFile.services) {
--- a/packages/protoc-gen-connect-es/src/typescript.ts
+++ b/packages/protoc-gen-connect-es/src/typescript.ts
@@ -23,9 +23,6 @@ import {
 
 export function generateTs(schema: Schema) {
   for (const protoFile of schema.files) {
-    if (protoFile.services.length == 0) {
-      continue;
-    }
     const file = schema.generateFile(protoFile.name + "_connect.ts");
     file.preamble(protoFile);
     for (const service of protoFile.services) {


packages/protoc-gen-connect-es/src/declaration.ts
```diff
@@ -23,10 +23,12 @@ import {
 
 export function generateDts(schema: Schema) {
   for (const protoFile of schema.files) {
+    // Generate a file before checking for services to properly support
+    // keep_empty_files=true.
+    const file = schema.generateFile(protoFile.name + "_connect.d.ts");
     if (protoFile.services.length == 0) {
       continue;
     }
```
timostamm, line 31:
```suggestion
    const file = schema.generateFile(protoFile.name + "_connect.d.ts");
```

timostamm (commented):
Thank for the update, Sean. We actually don't need the if statement at all - updating the docs in https://github.com/bufbuild/protobuf-es/pull/480. Do you mind removing the if statement completely as suggested in the three generator functions?

packages/protoc-gen-connect-es/src/declaration.ts
sjbarag: Ooh thanks for the docs update to `preamble()` — this is much simpler 🙂

timostamm (approved):
Thank you for for the contribution 🙂

PR #614, opened 2023-05-03
Update package.json engines field for node 16

dimitropoulos:
I noticed that protoc-gen-connect-es is [using](https://github.com/bufbuild/connect-es/blob/29dc336bff2246044a9ac331ee70fedcfdcfa46c/packages/protoc-gen-connect-es/package.json#L15) node 14, which is now [completely EOL](https://endoflife.date/nodejs).  Especially with the pending 1.0 release for some things, it's a good time to get on node 18, the present "active LTS".

closes TCN-1694

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,8 +17,8 @@
   "type": "module",
   "engineStrict": true,
   "engines": {
-    "node": ">=16",
-    "npm": ">=8"
+    "node": ">=18",
+    "npm": ">=9"
   },
   "devDependencies": {
     "typescript": "^5.0.3",
--- a/package.json
+++ b/package.json
@@ -17,8 +17,8 @@
   "type": "module",
   "engineStrict": true,
   "engines": {
-    "node": ">=18",
-    "npm": ">=9"
+    "node": ">=16",
+    "npm": ">=8"
   },
   "devDependencies": {
     "typescript": "^5.0.3",
--- a/Makefile
+++ b/Makefile
@@ -10,13 +10,13 @@ TMP   = .tmp
 BIN   = .tmp/bin
 BUILD = .tmp/build
 GEN   = .tmp/gen
-CROSSTEST_VERSION := 35f013f2f543646f4a40579993e7bb68cfb03133
+CROSSTEST_VERSION := 538426871c217f16bf6171dccab0a58e11b39daa
 LICENSE_HEADER_YEAR_RANGE := 2021-2023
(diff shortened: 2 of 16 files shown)


package.json
```diff
@@ -17,8 +17,8 @@
   "type": "module",
   "engineStrict": true,
   "engines": {
-    "node": ">=16",
-    "npm": ">=8"
+    "node": ">=18",
+    "npm": ">=9"
```
timostamm, line 21: What is the benefit of constraining the npm version?

timostamm (changes requested):
Thanks for keeping this up to date. Node v14 is no longer maintained, good call to remove it. Node v16 on the other hand is maintained and we support it.

package.json
dimitropoulos: I only did it for consistency since I saw it in some places but not others.  Should I remove it everywhere?

package.json
```diff
@@ -17,8 +17,8 @@
   "type": "module",
   "engineStrict": true,
```
dimitropoulos, line 18: hey, would you like to have `engineStrict` everywhere? seems like it can't hurt and is related to this PR.

dimitropoulos:
@timostamm [`a4dfe8d` (#614)](https://github.com/bufbuild/connect-es/pull/614/commits/a4dfe8d3e06ba925a97934037370d37a4da72ae2) goes with node 16

package.json
timostamm: There is _some_ use to having them in the top-level package.json, so that you at least get a warning if you try to build the project with an unsupported version. In other places, yes, let's remove them 🙂

dimitropoulos: sure! I'm happy to put it in whichever ones you think is best :)

Which ones are the top-level packages?  E.g. `connect` didn't previously have an npm designation.  Should it now?

timostamm: [This](https://github.com/bufbuild/connect-es/blob/29dc336bff2246044a9ac331ee70fedcfdcfa46c/package.json#L19-L22) is the top-level package.json. It's only relevant for contributors. 

I do not think the @bufbuild/connect package.json should have an engines field. If you are aware of a good benefit for constraining the npm version there, I'm all ears. If there isn't a really good one, we can save some time by not setting it, and then never ever having to update it.

dimitropoulos: ahhh gotcha gotcha.  I misunderstood because you said "them", so I thought you were talking about more than one.

please just tell me plainly which one's you'd like to have/not have and I'd be happy to update.  It sounds like you're saying to remove all engines fields except the repo-root package.json.

PR #618, opened 2023-05-04
Allow a fetch override for connect-web

### LINKED ISSUE #403: Allow customizing fetch

dimitropoulos:
closes #403

This PR allows for passing in a custom fetch implementation, such as that used by sveltekit (and Angular zones, so I have it).

To really test this, you need to actually use it with one of these libraries, so I have done so as a proof of concept in https://github.com/bufbuild/connect-es-integration/commit/6ddf8e39b87a752d87de61a538714202b483ff3a.


timostamm (commented):
> This PR allows for passing in a custom fetch implementation, such as that used by sveltekit (and Angular zones, so I have it).

Sweet!

> To really test this, you need to actually use it with one of these libraries, so I have done so as a proof of concept in [bufbuild/connect-es-integration@6ddf8e3](https://github.com/bufbuild/connect-es-integration/commit/6ddf8e39b87a752d87de61a538714202b483ff3a).

It looks like we are providing the Svelte fetch function to a client that makes both a unary and a server-streaming call, with the Connect protocol with the JSON format. Do both calls happen server side with this call, and no RPC request happens after page load?

Does this work with gRPC-web as well?

dimitropoulos:
>Do both calls happen server side with this call, and no RPC request happens after page load?

[yes](https://github.com/bufbuild/connect-es-integration/commit/b84f0b475cb24a17c0e8ef670a0ac09c8d1a7c0d)

> Does this work with gRPC-web as well?

[yes](https://github.com/bufbuild/connect-es-integration/commit/ec6e067fb132882d22e7440cf141d461305bd528)

timostamm (approved):

paul-sachs (approved):

Issue #576, opened 2023-04-05
DEADLINE_EXCEEDED not throwing an error and hangs indefinitely


RichieAHB:
**Describe the bug**

I'm calling a gRPC service written in Java, and when passing an invalid argument to the service, the service will hang until the deadline. On the backend for the client I'm using `connect-node`. The channel is configured like so:

```ts
const DYNAMO_CLIENT = createPromiseClient(
  DynamoService,
  createGrpcTransport({
    baseUrl: process.env.BACKEND_URL,
    httpVersion: '2',
    useBinaryFormat: true,
  }),
);
```

I'm then calling a unary method on the service like so:

```ts
try {
    const response = await DYNAMO_CLIENT.listTargetServices(
      {
        targetServiceDiscoveryName: decodeURIComponent(target as string),
      },
      { timeoutMs: 10000 },
    );
    console.log(response);
  } catch (e) {
    console.error('Error listing target services', e);
  }
```

With this setup the call above just hangs indefinitely. I can see the server is responding with a `DEADLINE_EXCEEDED` as when calling this same method with `grpcurl` like so:

```bash
grpcurl \           
  --plaintext \
  -d '{ "target_service_discovery_name": "shasdfasdf" }' \
  -max-time 10 \
  -H 'service: dynamo' \
  localhost:5990 \
  richieahb.dynamo.v1.DynamoService/ListTargetServices
```

I get a response after 10 seconds:

```
ERROR:
  Code: DeadlineExceeded
  Message: context deadline exceeded
```

I've also added breakpoints into the Java and tested locally and can see it's calling `onError` in the `StreamObserver` used for sending the responses to the client.

**To Reproduce**

If you encountered an error message, please copy and paste it verbatim.
If the bug is specific to an RPC or payload, please provide a reduced
example.

**Environment (please complete the following information):**
- ~@bufbuild/connect-web version: (for example, `0.1.0`)~
- @bufbuild/connect-node version: (for example, `0.8.5`)
- Frontend framework and version: (for example, `next@12.1.0`) (although this call is from the server)
- Node.js version: (for example, `18.15.0`)
- Browser and version: (for example, `Google Chrome 111.0.5563.146`)

If your problem is specific to bundling, please also provide the following information: 

- ~Bundler and version: (for example, `webpack@5.74.0`)~
- ~Bundler plugins and version: (for example `compression-webpack-plugin@10.0.0`)~
- ~Bundler configuration file: ~
```js

```

**Additional context**
N/A

[...]

PR #619, opened 2023-05-05
Improve tests for Node.js clients and fix hanging requests

### LINKED ISSUE #463: Stream connection hanging forever
**Describe the bug**

I have a stream service in a server behind a proxy that has a timeout of 10 min and then closes the connection, when this happens I expect the async iterator to throw an exception (so I can reconnect again) but instead it just hangs forever.

I found this [related issue in the grpc-js library](https://github.com/grpc/grpc-node/issues/2176) which looks the same, but I've been able to workaround it by enabling sending keepalive pings with `'grpc.keepalive_time_ms'` option, can a similar option be added to connect-node library?

timostamm:
Investigating https://github.com/bufbuild/connect-es/issues/463 and https://github.com/bufbuild/connect-es/issues/576, I found situations where both HTTP/1.1 and HTTP/2 client requests on Node.js can hang. This PR fixes them.

The issues were found by adding test coverage. We do not have definitive proof that this fixes #463 because we do not have a reproducible case, but I am fairly confident that it does.

### DIFF
--- a/packages/connect-node/src/node-universal-client.spec.ts
+++ b/packages/connect-node/src/node-universal-client.spec.ts
@@ -13,25 +13,25 @@
 // limitations under the License.
 
 import * as http2 from "http2";
-import type * as net from "net";
+import * as http from "http";
+import { ConnectError, connectErrorFromReason } from "@bufbuild/connect";
 import { createAsyncIterable } from "@bufbuild/connect/protocol";
 import { createNodeHttpClient } from "./node-universal-client.js";
-import { encodeEnvelope } from "@bufbuild/connect/protocol";
+import { useNodeServer } from "./use-node-server-helper.spec.js";
 
-describe("Node.js http2 API", function () {
-  it("should see reset codes from the client side on the server", async function () {
-    const server = await startServer();
-    await h2RequestWithReset(server.baseUrl, http2.constants.NGHTTP2_CANCEL);
-    const { rstCode } = await server.stop();
-    expect(rstCode).toBe(http2.constants.NGHTTP2_CANCEL);
-  });
+// Polyfill the Headers API for Node versions < 18
+import "./node-headers-polyfill.js";
 
-  /**
-   * Issues an H2 request, and immediately resets with the given code.
-   */
-  async function h2RequestWithReset(baseUrl: string, rstCode: number) {
+describe("node http/2 client closing with RST_STREAM with code CANCEL", function () {
+  let serverReceivedRstCode: number | undefined;
+  const server = useNodeServer(() =>
+    http2.createServer().on("stream", (stream) => {
+      stream.on("close", () => (serverReceivedRstCode = stream.rstCode));
+    })
+  );
+  it("should send RST_STREAM frame to the server", async function () {
     return new Promise<void>((resolve) => {
-      http2.connect(baseUrl, (session: http2.ClientHttp2Session) => {
+      http2.connect(server.getUrl(), (session: http2.ClientHttp2Session) => {
         const stream = session.request(
           {
             ":method": "POST",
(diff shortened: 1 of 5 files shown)


packages/connect-node/src/node-universal-client.spec.ts
```diff
@@ -40,118 +40,802 @@ describe("Node.js http2 API", function () {
...
-      const signal = AbortSignal.abort();
-      const client = createNodeHttpClient({
-        httpVersion: "2",
-        baseUrl: "http://example.com",
-        keepSessionAlive: false,
-      });
-      try {
-        await client({
-          url: "http://example.com",
+describe("universal node http client", function () {
```
timostamm, line 65: These added tests spin up a Node.js HTTP/1.1 or HTTP/2.2 server that aborts incoming requests. We run a universal Node.js client (which is used by all Transports from `@bufbuild/connect-node`) against the servers. We test closing the request at several stages, for example before the first response byte is written, or while the request body is still read.

packages/connect-node/src/node-universal-client.ts
```diff
@@ -345,22 +341,20 @@ function h2Request(
...
+        // Node.js http2 streams that are aborted via an AbortSignal close with
+        // an RST_STREAM with code INTERNAL_ERROR.
+        // To comply with the mapping between gRPC and HTTP/2 codes, we need to
+        // close with code CANCEL.
+        // See https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#errors
+        // See https://www.rfc-editor.org/rfc/rfc7540#section-7
+        const rstCode =
+          reason instanceof ConnectError && reason.code == Code.Canceled
+            ? H2Code.CANCEL
+            : H2Code.INTERNAL_ERROR;
```
timostamm, line 359: No logical changes here, just making use of the added enumeration for error codes.

packages/connect-node/src/node-universal-client.ts
```diff
@@ -553,8 +551,10 @@ function createSentinel(signal?: AbortSignal): Sentinel {
   const s = Object.assign(p, c);
 
   function onSignalAbort() {
+    // AbortSignal.reason was added in Node.js 17.2.0, we cannot rely on it.
     c.reject(
-      new ConnectError("operation was aborted via signal", Code.Canceled)
+      signal?.reason ??
+        new ConnectError("This operation was aborted", Code.Canceled)
```
timostamm, line 560: Unrelated to the bug fixes. With Node.js v16 being EOL soon, we can make better use of AbortSignal.reason.

packages/connect-node/src/node-universal-client.ts
```diff
@@ -132,23 +134,15 @@ function createNodeHttp1Client(
...
-              sentinel.reject(
-                new ConnectError("node response aborted", Code.Aborted)
-              )
-            );
-            response.on("timeout", () =>
-              sentinel.reject(
-                new ConnectError("node response timed out", Code.Aborted)
-              )
+            sentinel.catch((reason) =>
+              response.destroy(connectErrorFromNodeReason(reason))
```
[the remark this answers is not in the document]
smaye81: Will this address the issue of explicit cancellations from the client and how the server handles those?

smaye81 (approved):

timostamm:
> Will this address the issue of explicit cancellations from the client and how the server handles those?

Do you mean https://github.com/bufbuild/connect-es/issues/576? No, it does not.

smaye81:
> > Will this address the issue of explicit cancellations from the client and how the server handles those?
> 
> Do you mean #576? No, it does not.

Ah no, I meant this discussion:  https://github.com/bufbuild/connect-es/discussions/315#discussioncomment-4829888 (I think they're different issues).

PR #620, opened 2023-05-05
Add client support for Connect HTTP GET requests

jchadwick-buf:
Adds support for [Connect HTTP GET requests](https://connect.build/docs/protocol#unary-get-request) on the client, both in Node.JS and the browser. Requires specifying the transport option `useHttpGet` for now to enable support.

Server implementation: #624


packages/connect/src/protocol-connect/get-request.ts
```diff
@@ -0,0 +1,90 @@
...
+ */
+export function transformConnectPostToGetRequest<
+  I extends Message<I> = AnyMessage,
+  O extends Message<O> = AnyMessage
+>(
+  request: UnaryRequest<I, O>,
+  message: Uint8Array,
+  useBase64: boolean
+): UnaryRequest<I, O> {
+  let query = "?connect=v1";
```
timostamm, line 53: Can use `protocolVersion` from packages/connect/src/protocol-connect/version.ts here.

packages/connect/src/protocol-connect/get-request.ts
```diff
@@ -0,0 +1,90 @@
...
+  if (useBase64) {
+    // TODO(jchadwick-buf): Three regex replaces seems excessive.
+    // Can we make protoBase64.enc more flexible?
+    return protoBase64
+      .enc(message)
+      .replace(/\+/g, "-")
+      .replace(/\//g, "_")
+      .replace(/=+$/, "");
+  } else {
+    return encodeURIComponent(new TextDecoder().decode(message));
```
timostamm, line 38: I am afraid this is lossy. TextDecoder uses UTF-8 by default and will use replacement characters for unrecognized code points.

packages/connect/src/protocol-connect/get-request.ts
```diff
@@ -0,0 +1,90 @@
...
+    query +=
+      "&encoding=" +
+      encodeURIComponent(contentType.slice(contentTypePrefix.length));
+  }
+  const compression = request.header.get(headerUnaryEncoding);
+  if (compression !== null && compression !== "identity") {
+    query += "&compression=" + encodeURIComponent(compression);
+
+    // Force base64 for compressed payloads.
+    useBase64 = true;
```
timostamm, line 65: I think we are going against the grain with percent encoding arbitrary octets. The standard libs I am familiar with  will expect UTF-8 by default, and data that is not valid UTF-8 will require bespoke code paths, if feasible at all. Case in point are dotnet's [ParseQueryString](https://learn.microsoft.com/en-us/dotnet/api/system.web.httputility.parsequerystring?view=netcore-3.1) and [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) in JS. From my understanding, RFC3986 mandates the use of UTF-8 specifically to avoid character encoding issues between different systems.

Percent-encoding JSON is fine because it is always UTF-8 (RFC8529 section 8.1), but proto binary and compressed messages should IMO always be base64 encoded.

John and @akshayjshah, what do you think?

packages/connect-node-test/src/crosstest/cacheable_unary.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+        payload: {
+          body: new Uint8Array(1024).fill(0),
+        },
+      });
+      it("with promise client", async function () {
+        const transport = transportFactory({
+          useHttpGet: true,
+          interceptors: [ensureGetRequest],
+        });
+        const client = createPromiseClient(TestService, transport);
```
timostamm, line 91: The interceptor is a valid and nice solution, but the preferred way to access response headers is via a call option, and I think we should prefer it in the tests too. 

For example:

```ts
function ensureGetRequest(header: Headers) {
  expect(header.has("request-protocol")).toBeTrue();
  if (header.get("request-protocol") === "connect") {
    expect(header.has("get-request")).toBeTrue();
  }
}
```

With:

```ts
const response = await client.cacheableUnaryCall(request, {onHeader: ensureGetRequest});
```

Could you update the four places?

packages/connect/src/protocol-connect/get-request.ts
```diff
@@ -0,0 +1,90 @@
...
+  headerUnaryContentLength,
+  headerUnaryEncoding,
+} from "./headers.js";
+
+const contentTypePrefix = "application/";
+
+function encodeMessageForUrl(message: Uint8Array, useBase64: boolean): string {
+  if (useBase64) {
+    // TODO(jchadwick-buf): Three regex replaces seems excessive.
+    // Can we make protoBase64.enc more flexible?
```
timostamm, line 31: protobuf-es uses this for JSON, and it doesn't seem ideal to add complexity, bundle size there for a feature that is not used by protobuf-es. I agree this isn't pretty, but I am not sure we have a really good alternative.

packages/connect/src/protocol/transport-options.ts
```diff
@@ -113,4 +113,11 @@ export interface CommonTransportOptions {
    * The default limit is the maximum supported value of ~4GiB.
    */
   writeMaxBytes: number;
+
+  /**
+   * Controls whether or not the fetch client should use Connect GET
+   * requests when available, on side-effect free methods. Defaults
+   * to false.
+   */
+  useHttpGet?: boolean;
```
timostamm, line 122: I think we should enable GET by default. Let's discuss offline.

packages/connect/src/protocol-connect/get-request.ts
jchadwick-buf: Agreed, and we're working to clarify the spec as discussed offline. The client doesn't try to encode binary as percent-encoding, and we have another PR to discuss this issue in at the moment, so I'll mark this resolved now.

packages/connect/src/protocol-connect/get-request.ts
jchadwick-buf: It's probably not great that it isn't obvious the base64 path is ALWAYS taken for binary data, but since it is, I'm marking this resolved. If I can make the control flow clearer that would be nice, I'm open to suggestions.

packages/connect/src/protocol-connect/get-request.ts
jchadwick-buf: Done

packages/connect-node-test/src/crosstest/cacheable_unary.spec.ts
jchadwick-buf: Thanks, that is a great improvement. Done.

packages/connect/src/protocol-connect/get-request.ts
timostamm: I suggest a doc comment on `transformConnectPostToGetRequest()`:

> If `message` is not valid UTF-8, `useBase64` must be true.

And to enforce it with `new TextDecoder("utf-8", {fatal: true})`.

packages/connect/src/protocol/transport-options.ts
timostamm: We discussed offline. I completely forgot about size constraints :facepalm:

packages/connect-web/src/connect-transport.ts
```diff
@@ -102,6 +103,13 @@ export interface ConnectTransportOptions {
    * Options for the binary wire format.
    */
   binaryOptions?: Partial<BinaryReadOptions & BinaryWriteOptions>;
+
+  /**
+   * Controls whether or not the fetch client should use Connect GET
```
timostamm, line 108: Can you avoid "fetch client" here and in transport-options.ts? 

For example "Controls whether or not Connect GET requests should be used when ...".

Since Node.js does not use the fetch API fetch(), it could be a bit confusing otherwise.

timostamm (commented):
We'll need a PR description. Aside from the comment nit below, and https://github.com/bufbuild/connect-es/pull/620#discussion_r1189696352, this looks ready to merge to me.

PR #623, opened 2023-05-08
Close the session on error in client transports

timostamm:
By default, we close HTTP/2 connections after each request in Node.js. 

But because of an oversight, the code handling the close never ran if we encountered an error, for example an unexpected response content type, or just an error encoded in the response body or response trailers of a streaming RPC.

### DIFF
--- a/packages/connect/src/protocol-connect/transport.ts
+++ b/packages/connect/src/protocol-connect/transport.ts
@@ -19,13 +19,6 @@ import type {
   PartialMessage,
   ServiceType,
 } from "@bufbuild/protobuf";
-import {
-  appendHeaders,
-  Code,
-  ConnectError,
-  runStreaming,
-  runUnary,
-} from "../index.js";
 import type {
   StreamRequest,
   StreamResponse,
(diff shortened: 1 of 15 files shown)


packages/connect/src/protocol-connect/transport.ts
```diff
@@ -78,6 +80,7 @@ export function createTransport(opt: CommonTransportOptions): Transport {
         opt.jsonOptions,
         opt
       );
+      const ac = createLinkedAbortController(signal);
```
timostamm, line 83: This is the relevant change: We create an AbortSignal and pass it to the HTTP client, and we wrap the rest of the Transport logic in a try-catch block and abort the signal on error. For HTTP/2, this closes the stream (and session), for HTTP/1.1 it calls destroy() on request and response.

packages/connect/src/protocol-connect/validate-response.ts
```diff
@@ -46,12 +45,7 @@ export function validateResponse(
       `HTTP ${status}`,
       codeFromHttpStatus(status)
     );
-    if (
-      methodKind == MethodKind.Unary &&
-      parsedType &&
-      !parsedType.stream &&
-      !parsedType.binary
-    ) {
+    if (methodKind == MethodKind.Unary && parsedType && !parsedType.stream) {
```
timostamm, line 48: This is an unrelated fix. For Connect unary, we always try to parse an error from the response body if we receive a non-successful HTTP status - not just for JSON, but for protobuf binary as well. This has slipped through our test suite so far, but is covered now.

smaye81 (approved):

PR #624, opened 2023-05-08
Add server support for Connect HTTP GET requests

jchadwick-buf:
Adds support for handling [Connect HTTP GET requests](https://connect.build/docs/protocol#unary-get-request) in servers. We also expose `protocolName` and `requestMethod` on handler context so that handlers can know how the RPC request was initiated.


packages/connect-node-test/src/helpers/test-routes.ts
```diff
@@ -57,8 +57,10 @@ const testService: ServiceImpl<typeof TestService> = {
     ]);
   },
 
-  cacheableUnaryCall(/*request*/) {
-    throw new ConnectError("TODO", Code.Unimplemented);
+  cacheableUnaryCall(request, context) {
+    // TODO: Need to be able to determine request protocol/HTTP method.
+    context.responseHeader.set("request-protocol", "unknown");
```
timostamm, line 62: We can add two properties to HandlerContext:

```ts
  readonly protocolName: string;
  readonly requestMethod: string;
```

Both values are available when we call createHandlerContext() in the handler-factory.ts files. We should wait for consensus on https://github.com/bufbuild/connect-go/pull/502.

packages/connect-express/src/express-connect-middleware.ts
```diff
@@ -74,7 +74,7 @@ export function expressConnectMiddleware(
     res: express.Response,
     next: express.NextFunction
   ) {
-    const uHandler = paths.get(req.url);
+    const uHandler = paths.get(req.url.split("?", 2)[0]);
```
timostamm, line 77: I think a short comment explaining that this strips the query from the request path will be helpful for future us 🙂

packages/connect-node-test/src/helpers/test-routes.ts
jchadwick-buf: Do you think we should block this PR on resolving this TODO? I'd like to get consensus on the other issue, but not 100% sure how long it will take either. It's not a huge deal though, since this isn't a blocker for the client part, which I think is more important anyways.

packages/connect-express/src/express-connect-middleware.ts
jchadwick-buf: Done.

packages/connect/src/protocol-connect/handler-factory.ts
```diff
@@ -12,7 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import type { MethodInfo } from "@bufbuild/protobuf";
+import { MethodIdempotency, protoBase64 } from "@bufbuild/protobuf";
+import type { JsonValue, MethodInfo } from "@bufbuild/protobuf";
 import { Message, MethodKind } from "@bufbuild/protobuf";
```
reyawn, line 17: The `import` (not `import type`) blocks should probably be merged here.

jchadwick-buf: Good catch, fixed.

packages/connect-node-test/src/helpers/test-routes.ts
jchadwick-buf: Since we have consensus on connect-go now, I went ahead and implemented this as-planned.

timostamm (commented):
Hey John, just two code style comments from me. I believe this is ready to merge otherwise.
Some more tests would be nice to have, so I put up https://github.com/bufbuild/connect-es/pull/629, based on this PR.

packages/connect/src/protocol-connect/handler-factory.ts
```diff
@@ -78,11 +85,16 @@ import {
 import { codeToHttpStatus } from "./http-status.js";
 import { errorToJsonBytes } from "./error-json.js";
 import { trailerMux } from "./trailer-mux.js";
-import { requireProtocolVersion } from "./version.js";
+import {
+  requireProtocolVersionHeader,
+  requireProtocolVersionParam,
+} from "./version.js";
 import { parseTimeout } from "./parse-timeout.js";
+import { paramMessage, paramBase64 } from "./query-params.js";
```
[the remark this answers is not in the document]
jchadwick-buf: I ran organize imports. I didn't really realize it handled our import style correctly, because some other tsserver functions don't seem to handle it very well. Thanks for the tip.

timostamm (approved):
LGTM, thank you for this!

Please add a PR description so our release changelog is comprehensible.

jchadwick-buf:
Thanks for your help + additional tests!

PR #629, opened 2023-05-12
Add more tests for GET support

timostamm:
This adds a little more coverage for GET in transports and handlers. We prove that clients issue a GET request with an URL like `http://example.com/TestService/UnaryNoSideEffects?connect=v1&encoding=proto&base64=1&message=CHs`, and we prove that handlers can handle such a request.


jchadwick-buf (approved):
Thanks, this looks like a good addition.

PR #630, opened 2023-05-12
Add createRouterTransport() to the connect-node-test crosstests

dimitropoulos:
I want to make sure this is moving in the right direction, so I've only done `connect-node-test` and will follow up with `connect-web-test` if this is correct.

We don't have good coverage for createRouterTransport() yet, but we really want to be certain that it behaves like other transports that use the network. We can use the crosstests for this.

The crosstest are, a schema and a set of tests defined in bufbuild/connect-crosstest. We have two implementations in the connect-es repository:
- connect-node-test: runs the Connect, gRPC, and gRPC-web transports with various options from @bufbuild/connect node against various servers
- connect-web-test: runs the Connect and gRPC-web transports from @bufbuild/connect-web with various options against various servers, in a headless web browser, on Node.js, optionally in a local web browser, and against several browsers and devices on browserstack (in CI).

We are also about to add connect-edge-test, for using the same pattern for code running in edge environments.

Ideally, we would add a router transport to all three test packages, so that we have confidence that it behaves as expected in all environments.

### DIFF
--- a/packages/connect-node-test/src/crosstest/cancel_after_begin.spec.ts
+++ b/packages/connect-node-test/src/crosstest/cancel_after_begin.spec.ts
@@ -28,38 +28,44 @@ describe("cancel_after_begin", function () {
     }
   }
 
-  servers.describeTransports((transport) => {
-    const servers = createTestServers();
-    beforeAll(async () => await servers.start());
-    it("with promise client", async function () {
-      const client = createPromiseClient(TestService, transport());
-      const controller = new AbortController();
-      async function* input() {
-        yield {
-          payload: {
-            body: new Uint8Array(),
-            type: PayloadType.COMPRESSABLE,
-          },
-        };
-        await new Promise((resolve) => setTimeout(resolve, 1));
-        controller.abort();
-        yield {
-          payload: {
-            body: new Uint8Array(),
-            type: PayloadType.COMPRESSABLE,
-          },
-        };
-      }
-      try {
-        await client.streamingInputCall(input(), {
-          signal: controller.signal,
-        });
-        fail("expected error");
-      } catch (e) {
-        expectError(e);
-      }
-    });
-  });
+  servers.describeTransportsExcluding(
+    [
+      "@bufbuild/connect (ConnectRouter, binary)",
+      "@bufbuild/connect (ConnectRouter, JSON)",
+    ],
+    (transport) => {
+      const servers = createTestServers();
+      beforeAll(async () => await servers.start());
+      it("with promise client", async function () {
+        const client = createPromiseClient(TestService, transport());
+        const controller = new AbortController();
+        async function* input() {
+          yield {
+            payload: {
+              body: new Uint8Array(),
+              type: PayloadType.COMPRESSABLE,
+            },
+          };
+          await new Promise((resolve) => setTimeout(resolve, 1));
+          controller.abort();
+          yield {
+            payload: {
+              body: new Uint8Array(),
+              type: PayloadType.COMPRESSABLE,
+            },
+          };
+        }
+        try {
+          await client.streamingInputCall(input(), {
+            signal: controller.signal,
+          });
+          fail("expected error");
+        } catch (e) {
+          expectError(e);
+        }
+      });
+    }
+  );
 
   afterAll(async () => await servers.stop());
 });
--- a/packages/connect-node-test/src/crosstest/unresolvable_host.spec.ts
+++ b/packages/connect-node-test/src/crosstest/unresolvable_host.spec.ts
@@ -15,11 +15,9 @@
 import {
   Code,
   ConnectError,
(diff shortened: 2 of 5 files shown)


packages/connect-node-test/src/node-readme.spec.ts
```diff
@@ -60,34 +61,53 @@ describe("node readme", function () {
...
+    httpVersion: "1.1" as const,
+  };
+
   it("createConnectTransport()", async function () {
     // A transport for clients using the gRPC protocol with Node.js `http` module
-    const transport = createConnectTransport({
-      baseUrl: "https://demo.connect.build",
-      httpVersion: "1.1",
-    });
+    const transport = createConnectTransport(optionsHttp1);
```
timostamm, line 75: This spec file tests that the examples we give in https://www.npmjs.com/package/@bufbuild/connect-node work.

Can you leave the transport options verbatim?

packages/connect-node-test/src/crosstest/cancel_after_begin.spec.ts
```diff
@@ -28,38 +28,44 @@ describe("cancel_after_begin", function () {
...
-        fail("expected error");
-      } catch (e) {
-        expectError(e);
-      }
-    });
-  });
+  servers.describeTransportsExcluding(
+    [
+      "@bufbuild/connect (ConnectRouter, binary)",
+      "@bufbuild/connect (ConnectRouter, JSON)",
```
timostamm, line 34: I would expect this test case to work with the router transport too. How is it failing?

packages/connect-node-test/src/crosstest/unresolvable_host.spec.ts
```diff
@@ -28,14 +30,31 @@ import { TestService } from "../gen/grpc/testing/test_connect.js";
...
   const optionsHttp1 = {
-    baseUrl: "https://unresolvable-host.some.domain",
+    ...baseOptions,
     httpVersion: "1.1" as const,
   };
+  const unresolvableRouter = ({ service }: ConnectRouter) => {
+    const unavailableRoute = () => {
+      throw new ConnectError(
+        "getaddrinfo ENOTFOUND unresolvable-host.some.domain",
+        Code.Unavailable
```
timostamm, line 48: This works 👍 

But by definition, the router transport only serves the routes you give to it. Any other route will hit raise this error: https://github.com/bufbuild/connect-es/blob/fc9cd63226289cc792c0e108c1aeefd802216510/packages/connect/src/protocol/universal-handler-client.ts#L36-L39

I can't think of a meaningful way to run the router transport with this test case, I think we can simply omit it here.

Could you instead add a test in packages/connect/src/router-transport.spec.ts to cover the behavior for calling an RPC with the router transport that isn't registered?

timostamm (commented):
If all but the unresolvable_host test succeed with this transport, that's pretty awesome.

packages/connect-node-test/src/node-readme.spec.ts
dimitropoulos: sure :) 0f0adcd7e3e78c254af3b531bf0b361b7ce2d9ba

packages/connect-node-test/src/crosstest/cancel_after_begin.spec.ts
dimitropoulos: it fails with:

```
Failures:
1) cancel_after_begin @bufbuild/connect (ConnectRouter, JSON) with promise client
  Message:
    Failed: expected error
  Stack:
        at <Jasmine>
        at UserContext.<anonymous> (file:///Users/dimitri/src/github.com/bufbuild/connect-es/packages/connect-node-test/dist/esm/crosstest/cancel_after_begin.spec.js:70:17)
```

packages/connect-node-test/src/crosstest/unresolvable_host.spec.ts
dimitropoulos: > I think we can simply omit it here.

sure: removed it in 7042dd57ad41a7d6191093b2417b8c57264d9208

dimitropoulos: > Could you instead add a test in packages/connect/src/router-transport.spec.ts to cover the behavior for calling an RPC with the router transport that isn't registered?

sure: 3bdbd362a8fbb959fa927245374d9a1f250b25a2

packages/connect-node-test/src/crosstest/cancel_after_begin.spec.ts
timostamm: Okay, we have to fix this, but not here.

timostamm (approved):
LGTM. Fine to follow up with the same in connect-web-test in a separate PR. 

Please update the PR description so we have a comprehensible release changelog.

PR #633, opened 2023-05-15
Add createRouterTransport() to the connect-web-test crosstests

dimitropoulos:
This PR is a continuation of #630 which adds crosstest coverage for connect-web.


packages/connect-web-test/src/cancellation.spec.ts
```diff
@@ -28,89 +28,96 @@ describe("explicit cancellation with AbortController", function () {
...
-          caughtError = true;
-          expect(e).toBeInstanceOf(ConnectError);
-          if (e instanceof ConnectError) {
-            expect(e.code).toBe(Code.Canceled);
+  describeTransportsExcluding(
+    [
+      // TODO(TCN-1763) support client-side cancellation in createRouterTransport()
+      "@bufbuild/connect-web (ConnectRouter, JSON)",
+      "@bufbuild/connect-web (ConnectRouter, binary)",
+    ],
```
dimitropoulos, line 36: for posterity, the messages here are 

```
explicit cancellation with AbortController @bufbuild/connect-web (ConnectRouter, JSON) with callback client works for unary method
  Message:
    Failed: expected callback client to swallow AbortError
    
    
explicit cancellation with AbortController @bufbuild/connect-web (ConnectRouter, JSON) with promise client works for server-streaming method
  Message:
    Expected false to be true.
```

timostamm (approved):

PR #635, opened 2023-05-16
Support timeouts in clients

timostamm:
In https://github.com/bufbuild/connect-es/pull/591, we added support for timeouts in handlers. For better reliability, timeouts are now honored in clients as well. 

For example, consider the following client call:

```ts
elizaClient.say({ sentence: "Hello" }, { timeoutMs: 2500 });
```

The timeout of 2.5 seconds is sent to the server in a request header (this works for the gRPC, gRPC-web, and for the Connect protocol). If the timeout is reached before the server could produce a response in time, it will send an error with Code.DeadlineExceeded instead. 

But if this error does not reach the client (for example because a connection became unresponsive), the client would wait for a response well past the timeout. With this PR, the timeout is not just sent as a header, but it is also honored in the client. If no response arrives before the timeout, the client will raise an error with Code.DeadlineExceeded itself, and stop the request.

The logic is implemented in the functions `runUnaryCall` and `runStreamingCall` from `@bufbuild/connect/protocol`, so that Transport implementations do not have to care about this detail themselves. 

The old versions of the functions - `runUnary` and `runStreaming` from `@bufbuild/connect` - are now marked deprecated.

### DIFF
--- a/packages/connect/src/protocol-grpc-web/transport.ts
+++ b/packages/connect/src/protocol-grpc-web/transport.ts
@@ -26,16 +26,16 @@ import type {
   UnaryRequest,
   UnaryResponse,
 } from "../index.js";
-import { Code, ConnectError, runStreaming, runUnary } from "../index.js";
+import { Code, ConnectError } from "../index.js";
 import type { CommonTransportOptions } from "../protocol/index.js";
 import {
   createAsyncIterable,
-  createLinkedAbortController,
   createMethodSerializationLookup,
   createMethodUrl,
   pipe,
   pipeTo,
-  transformCatchFinally,
+  runStreamingCall,
+  runUnaryCall,
   transformCompressEnvelope,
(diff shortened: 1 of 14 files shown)


packages/connect/src/protocol-grpc-web/transport.ts
```diff
@@ -71,9 +71,11 @@ export function createTransport(opt: CommonTransportOptions): Transport {
         opt.jsonOptions,
         opt
       );
-      const ac = createLinkedAbortController(signal);
-      return await runUnary<I, O>(
```
timostamm, line 75: The new `run*` functions handle abort signals and timeouts. They also trigger the abort signal if the Transport raises an error. This saves quite a bit of plumbing in Transport implementations, and seems worth the change.

smaye81 (approved):

PR #637, opened 2023-05-16
Support client-side cancellation in the router transport

timostamm:
`createRouterTransport()` added in https://github.com/bufbuild/connect-es/pull/548) creates a client for an in-memory server. It had a small flaw: Cancelling calls via an AbortSignal did not work reliably. 

This PR updates the implementation to always abort a call right away, just like the Transports that actually make calls across the network via fetch() or one of the Node.js http modules.

### DIFF
--- a/packages/connect/src/protocol/universal-handler-client.ts
+++ b/packages/connect/src/protocol/universal-handler-client.ts
@@ -14,9 +14,10 @@
 
 import { Code } from "../code.js";
 import { ConnectError } from "../connect-error.js";
-import { createAsyncIterable } from "./async-iterable.js";
+import { createAsyncIterable, pipe } from "./async-iterable.js";
 import type { UniversalHandler } from "./universal-handler.js";
 import type { UniversalClientFn } from "./universal.js";
+import { getAbortSignalReason } from "./signals.js";
 
 /**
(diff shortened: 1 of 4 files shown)


packages/connect/src/protocol/universal-handler-client.ts
```diff
@@ -38,23 +39,62 @@ export function createUniversalHandlerClient(
...
       trailer: new Headers(uServerRes.trailer),
     };
   };
 }
+
+function raceSignal<T>(signal: AbortSignal, promise: Promise<T>): Promise<T> {
+  let cleanup: (() => void) | undefined;
+  const signalPromise = new Promise<never>((_, reject) => {
+    if (signal.aborted) {
+      reject(getAbortSignalReason(signal));
```
smaye81, line 91: Could you just call `onAbort` here? Or do you intentionally want to keep them separate?

smaye81 (approved):

smaye81 (approved):

packages/connect/src/protocol/universal-handler-client.ts
timostamm: Good call, updated.

PR #638, opened 2023-05-17
Add from() and findDetails() to ConnectError

timostamm:
This adds two methods to ConnectError:

The static `from()`, which converts any error into a ConnectError, if it is not already a ConnectError:

```ts
try {
  // call an RPC
} catch (e) {
  const err: ConnectError = ConnectError.from(reason);
  console.log(err.message);
  console.log(err.code == Code.ResourceExhausted);
} 
```

This is a replacement for [connectErrorFromReason()](https://github.com/bufbuild/connect-es/blob/cf73ff6589aad854dea0bf5f01d77d8e00d1514d/packages/connect/src/connect-error.ts#L176), which is deprecated with this PR. The static method is more discoverable, and it does not require a separate import.

And the instance method `findDetails`():

```ts
import {ConnectError} from "@bufbuild/connect-web";
import {LocalizedMessage} from "./error_details_pb.js";

function handle(err: ConnectError) {
  const localized = err.findDetails(LocalizedMessage)
    .find(i => i.locale === navigator.language);
  console.log(localized?.message);
}
```

This is a replacement for [connectErrorDetails()](https://github.com/bufbuild/connect-es/blob/cf73ff6589aad854dea0bf5f01d77d8e00d1514d/packages/connect/src/connect-error.ts#L115), which is deprecated with this PR. The instance method is more discoverable, does not require a separate import, and has a more descriptive name. 

The standalone function has one override accepting multiple message types that the new method does not implement. The simple workaround is to use a registry:
```diff
- connectErrorDetails(err, TypeA, TypeB);
+ import { createRegistry } from "@bufbuild/protobuf";
+ err.findDetails(createRegistry(TypeA, TypeB));
```

This change adds 1% to the bundle size according to [our code size comparison](https://github.com/bufbuild/connect-es/tree/main/packages/connect-web-bench), but the improved developer experience seems worth it.


packages/connect/src/connect-error.ts
```diff
@@ -95,6 +95,72 @@ export class ConnectError extends Error {
...
+          details.push(data);
+        }
+        continue;
+      }
+      const type = registry.findMessage(data.type);
+      if (type) {
+        try {
+          details.push(type.fromBinary(data.value));
+        } catch (_) {
+          //
```
dimitropoulos, line 158: I realize this was copied over from the previous implementation but I'd like to better understand what the situation could be where the binary data fails to parse

packages/connect/src/connect-error.ts
```diff
@@ -95,6 +95,72 @@ export class ConnectError extends Error {
...
+   * messages. Any decoding errors are ignored, and the detail will simply be
+   * omitted from the list.
+   */
+  findDetails<T extends Message<T>>(type: MessageType<T>): T[];
+  findDetails(registry: IMessageTypeRegistry): AnyMessage[];
+  findDetails(
+    typeOrRegistry: MessageType | IMessageTypeRegistry
+  ): AnyMessage[] {
+    const registry =
+      "typeName" in typeOrRegistry
```
dimitropoulos, line 139: I'm surprised we don't have an `IsMessage` type guard yet.  we have to copy `"typeName" in` three times because of it.  It'd make this code much clearer (and less error prone since there's no exclusivity check on the `in` operator which means it'll let you check things that don't exist in either type, e.g. `typename`).  I'll put up a protobuf-es PR to add `IsMessage`.

timostamm: Hold on. We wouldn't want to use a generic type guard function here because it would come with a much larger bundle size increase. How often do you need to distinguish a message _type_? I do not believe there is a pressing need for a guard.

dimitropoulos (approved):

packages/connect/src/connect-error.ts
dimitropoulos: in that case, it couldn't hurt to explain that in the comment on line 158 for the next person after me that wonders.

PR #642, opened 2023-05-17
Ignore unknown JSON fields by default

timostamm:
When changing a schema, servers are usually updated first. With the default behavior of the JSON format to reject unknown fields, this causes an issue with in simple cases such as adding a field to a response message. https://github.com/bufbuild/connect-es/issues/613 is a good example.

This PR changes the default behavior of both clients and servers to ignore unknown fields. It is equivalent to the following option (for a Transport, or for the ConnectRouter):

```ts
jsonOptions: { ignoreUnknownFields: true },
```

After this PR is merged, clients or servers can opt in to reject unknown fields by providing `jsonOptions: { ignoreUnknownFields: false }`.


packages/connect/src/protocol/serialization.ts
```diff
@@ -206,25 +206,29 @@ type JsonSerializationOptions = Partial<JsonReadOptions & JsonWriteOptions> & {
...
+ * By default, unknown fields are ignored.
  */
 export function createJsonSerialization<T extends Message<T>>(
   messageType: MessageType<T>,
   options: JsonSerializationOptions | undefined
 ): Serialization<T> {
   const textEncoder = options?.textEncoder ?? new TextEncoder();
   const textDecoder = options?.textDecoder ?? new TextDecoder();
+  const o = options ?? {};
+  o.ignoreUnknownFields ??= true;
```
timostamm, line 219: This is the change. The rest is tests.

smaye81:
Nice. A few small questions:

1. Do we consider this a breaking change? I'm guessing no, but I can see the argument where it could be.
2. Should we mention anywhere that this deviates from the actual JSON docs? Or is it possible to work with the Protobuf team to see if ignoring unknown fields should be the default?

Also, we need to update the docs here:  https://github.com/bufbuild/protobuf-es/blob/main/docs/runtime_api.md#json

timostamm:
> Do we consider this a breaking change?

Yes, a breaking change. Only if you rely on clients and server to reject new fields. But still a breaking change.

> Should we mention anywhere that this deviates from the actual JSON docs?

Yes, I think [this docs page](https://connect.build/docs/web/choosing-a-protocol#connect) fits best.

russellhaering:
Just ran into this in production, easy to work around but just wanted to chime in that this seems like a more sensible and expected default.

arvid220u:
Agree that this should definitely be the default. One of the biggest reasons for us to use protobufs was to have backwards compatibility, and we just discovered (the hard way) that our production deployment was not backwards-compatible because of using the default options here... 

Would be great to see this PR being merged!

smaye81 (approved):

timostamm:
@russellhaering, @arvid220u, thanks for the feedback. This is going out with the next release.

PR #645, opened 2023-05-22
Avoid circular dependencies

### LINKED ISSUE #644: Circular dependencies due to barrel imports
Context: I was experimenting a bit with `bun`. I am aware that due to missing support for `http2` and `zlib` (brotli in particular) we can't use `connect-node` with `bun` yet, however `connect-web` would work! Or at least it could ...

It currently also doesn't because there are some circular dependencies / circular imports going on. This is due to the use of (internally importing from) barrel files within the code base.

Personally, I try to avoid barrel files entirely (except for exports entrypoints, ofc.) and strictly avoid importing from barrel files within the same code base.  This can lead to some nasty side effects both within the hosting code base as well as in consuming applications & bundlers (e.g. negatively affecting tree shaking in some cases).

---

I'd like to propose to nuke all internal imports from barrel files and instead simply import from the source of origin. That way, all circular dependencies are either immediately resolved or surfaced.

One of the offending circular imports:

`compression.ts` tries to import `connect-error.ts`:
https://github.com/bufbuild/connect-es/blob/main/packages/connect/src/protocol/compression.ts#L15

`connect-error.ts` imports `protocol/index.ts` which in turn imports `compression.ts`.
https://github.com/bufbuild/connect-es/blob/main/packages/connect/src/connect-error.ts#L23 clashes with

These circular imports are quite opaque when using barrel files.

This currently breaks `bun` and also causes a bunch of warnings e.g. in `rollup`

fubhy:
Fixes #644

This removes internal imports from barrel files throughout.

### DIFF
--- a/packages/connect/src/protocol-grpc-web/validate-response.ts
+++ b/packages/connect/src/protocol-grpc-web/validate-response.ts
@@ -23,7 +23,7 @@ import {
 } from "./headers.js";
 import { codeFromHttpStatus } from "../protocol-connect/http-status.js";
 import { findTrailerError } from "../protocol-grpc/trailer-status.js";
-import { Compression } from "../protocol/compression.js";
+import type { Compression } from "../protocol/compression.js";
 
 /**
  * Validates response status and header for the gRPC-web protocol.
--- a/packages/connect/src/protocol-grpc-web/validate-response.ts
+++ b/packages/connect/src/protocol-grpc-web/validate-response.ts
@@ -15,17 +15,15 @@
 import { ConnectError } from "../connect-error.js";
 import { Code } from "../code.js";
 import { parseContentType } from "./content-type.js";
-import type { Compression } from "../protocol/index.js";
-import {
-  codeFromHttpStatus,
-  findTrailerError,
-} from "../protocol-grpc/index.js";
 import {
   headerContentType,
   headerEncoding,
   headerGrpcMessage,
   headerGrpcStatus,
 } from "./headers.js";
+import { codeFromHttpStatus } from "../protocol-connect/http-status.js";
+import { findTrailerError } from "../protocol-grpc/trailer-status.js";
+import { Compression } from "../protocol/compression.js";
 
 /**
  * Validates response status and header for the gRPC-web protocol.
--- a/packages/connect/src/protocol-grpc-web/validate-response.ts
+++ b/packages/connect/src/protocol-grpc-web/validate-response.ts
@@ -21,7 +21,7 @@ import {
   headerGrpcMessage,
   headerGrpcStatus,
 } from "./headers.js";
-import { codeFromHttpStatus } from "../protocol-connect/http-status.js";
+import { codeFromHttpStatus } from "../protocol-grpc/http-status.js";
 import { findTrailerError } from "../protocol-grpc/trailer-status.js";
 import type { Compression } from "../protocol/compression.js";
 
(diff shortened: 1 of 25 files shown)


timostamm (commented):
Hey @fubhy, while we have them, we might just as well make use of the barrels to make our own imports slightly more readable. 

But the import you describe in https://github.com/bufbuild/connect-es/issues/644 is a bug. I pushed up a commit that disallows cyclic imports via an eslint plugin.

It looks like one import got left behind 🙂 

> src/protocol-connect/validate-response.ts(76,22): error TS2304: Cannot find name 'Compression'

fubhy:
Hi @timostamm! Definitely your call but personally I'd still recommend to avoid barrel files internally. Especially given the excellent code navigation tools in the IDE landscape these days I personally prefer explicit direct imports instead of opaque `index.js` here and there. I think for me that's actually less readable. But that might be a matter of preference indeed.

More significantly, In addition to the risk of accidentally introducing circular dependencies (which can be avoided with linting rules indeed), I've also seen weird behavior with bundlers in the past. Code splitting in particular.

fubhy:
(That said, I'm happy to revert those changes if you'd prefer to keep the barrel file imports regardless.)

fubhy:
TL;DR Disadvantages:

- Circular dependency hazard
- Running tests in watch mode triggers unrelated re-tests when
- Making things tougher to process & optimize for bundlers (currently trying to find the scenario that I ran into not long ago)
- Hazard of inadvertently imposing runtime dependencies on consumers (pulling in more than necessary)
- IDEs now offer multiple options for importing the same symbol (DX suffers imho)
- Opaque `index.js` files with no real semantics
- Extra steps for code navigation

... That's probably not all. Will add more here if more comes to mind ;-)

packages/connect/src/protocol-grpc-web/validate-response.ts
```diff
@@ -15,17 +15,15 @@
...
-  codeFromHttpStatus,
-  findTrailerError,
-} from "../protocol-grpc/index.js";
 import {
   headerContentType,
   headerEncoding,
   headerGrpcMessage,
   headerGrpcStatus,
 } from "./headers.js";
+import { codeFromHttpStatus } from "../protocol-connect/http-status.js";
```
timostamm, line 24:
```suggestion
import { codeFromHttpStatus } from "../protocol-grpc/http-status.js";
```

timostamm (commented):
Make sure to run `make bench`, the changes will affect the code size benchmark.

> Making it tougher for bundlers to process & optimize things

Would be interested to know what situation that was.

> IDEs now offer multiple options for importing the same symbol.

I'm aware, believe me 😆. But we're okay as long as the DX for the user is good.

fubhy:
> I'm aware, believe me laughing. But we're okay as long as the DX for the user is good.

Painfully aware? :-P Yeah... It's not the end of the world but it does annoy me sometimes.

> Would be interested to know what situation that was.

Found it. The problem was with wildcard imports / exports. Code splitting doesn't (didn't?) work with SWC in a Next.js project with those. Even with `sideEffects: false`. Not sure if that's fixed or if there's an inherent problem with that (or if that is / was just a bug or missing feature in SWC) but we fixed it by dropping the barrel files (although switching to named exports would've worked too). The point is that there are footguns :-P

timostamm (approved):
> The problem was with wildcard imports / exports

👍 I'm not sure it's relevant for _users_ of connect-es, but it doesn't hurt to be cautious. We'd also love to idiomatically support Deno at some point, so this is the right direction. 

The browserstack test failure is expected on a fork. The tests passed locally. LGTM.

Thank you for the fix 🙂

PR #657, opened 2023-05-29
Add more tests for Node.js universal handlers

timostamm:
This PR adds more test coverage for universalRequestFromNodeRequest() to guard against regression. It does not change any logic.

### DIFF
--- a/packages/connect-node/src/node-universal-handler.spec.ts
+++ b/packages/connect-node/src/node-universal-handler.spec.ts
@@ -17,21 +17,23 @@ import * as http2 from "http2";
 import * as http from "http";
 import { universalRequestFromNodeRequest } from "./node-universal-handler.js";
 import { ConnectError } from "@bufbuild/connect";
-import { readAllBytes } from "@bufbuild/connect/protocol";
+import { getNodeErrorProps } from "./node-error.js";
+import {
+  assertByteStreamRequest,
+  readAllBytes,
+  type UniversalServerRequest,
+} from "@bufbuild/connect/protocol";
 
 // Polyfill the Headers API for Node versions < 18
 import "./node-headers-polyfill.js";
 
 describe("universalRequestFromNodeRequest()", function () {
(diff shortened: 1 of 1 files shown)


packages/connect-node/src/node-universal-handler.spec.ts
```diff
@@ -63,50 +65,91 @@ describe("universalRequestFromNodeRequest()", function () {
...
-        expect(universalRequestSignal.reason.message).toBe(
+      expect(serverRequest?.signal).toBeInstanceOf(AbortSignal);
+      expect(serverRequest?.signal.aborted).toBeTrue();
+      expect(serverRequest?.signal.reason).toBeInstanceOf(Error);
+      if (serverRequest?.signal.reason instanceof Error) {
+        expect(serverRequest.signal.reason.name).toBe("AbortError");
+        expect(serverRequest.signal.reason.message).toBe(
           "This operation was aborted"
         );
       }
```
dimitropoulos, line 76: since `signal.reason` is typed `any`, the correct way to pass this information to the type system is with an `asserts` function:

```ts
function assertIsError(error: unknown): asserts error is Error {
  if (!(error instanceof Error)) {
    throw new Error(`input ${error} is not an instance of Error`)
  }
};
```

Then we can avoid the (usually noop) conditional in a test

```suggestion
  assertIsError(serverRequest?.signal.reason);
  expect(serverRequest.signal.reason.name).toBe("AbortError");
  expect(serverRequest.signal.reason.message).toBe(
    "This operation was aborted"
  );
```

dimitropoulos (approved):
so just to make sure I follow: was there something specific that happened recently that raised this?

packages/connect-node/src/node-universal-handler.spec.ts
timostamm: I would prefer this to be built into the test framework. Maintaining assertion functions as helpers is not ideal.

PR #668, opened 2023-06-05
Modify parsing of response for Connect unary requests

smaye81:
This modifies the response parsing for Connect unary requests based on the serialization format. If `useBinaryFormat` is true, it will use the output type's `fromBinary` against `response.arrayBuffer`.

If the serialization format is JSON, it will use the output type's `fromJson` method against `response.json`.

Note that this fixes two outstanding issues:

* It allows React-Native to be used with Connect unary requests without the need for patches. See https://github.com/bufbuild/connect-es-integration/pull/602 for more information. With this change in place, the React-Native example in Connect-ES Integration works over Connect and unary. Note that streaming still is out of scope.

* This also fixes the hidden issues with Svelte and universal SSR. Tests were run with the Svelte example in Connect-ES Integration using this change and universal SSR is working as intended:

**No additional XHR requests issued from the browser**

image

**Hydration data in the document returned from the initial `load` call:**

image


smaye81:
@timostamm I saw some tests for Connect and JSON format already present in the `crosstests` folder, but I can add more if there's other scenarios you'd like to verify.

packages/connect-web/src/connect-transport.ts
```diff
@@ -204,7 +204,15 @@ export function createConnectTransport(
...
-            message: parse(new Uint8Array(await response.arrayBuffer())),
+            message: useBinaryFormat
+              ? method.O.fromBinary(
+                  new Uint8Array(await response.arrayBuffer()),
+                  options.binaryOptions
+                )
+              : method.O.fromJson(
+                  (await response.json()) as JsonValue,
+                  options.jsonOptions
+                ),
```
timostamm, line 215: We are still pulling `parse` into the bundle. Would this shave off a couple of bytes?

```suggestion
            message: useBinaryFormat
              ? parse(new Uint8Array(await response.arrayBuffer()))
              : method.O.fromJson(
                  (await response.json()) as JsonValue,
                  options.jsonOptions
                ),
```

timostamm (commented):
Now that we support a feature that relies on the specific behavior of calling Response#json, we need to add a test to make sure we do not regress. Can you add test that provides a mocked `fetch` to the transport?

smaye81:
> Now that we support a feature that relies on the specific behavior of calling Response#json, we need to add a test to make sure we do not regress. Can you add test that provides a mocked `fetch` to the transport?

Sure. Do you want the mocked `fetch` to do anything special, or just delegate to the global `fetch`? They both have to return a `Promise<Response>`, so I'm not sure if we want to mock out any special circumstances?

timostamm:
> It's a bit tough to use the spies for the Response object because the individual tests don't currently know the type of transport or options used (i.e. whether useBinaryFormat is true or false). So, I modified the list of crosstests to provide this information. This could probably be organized better, but I wanted to get your thoughts first.

There is a simple solution for this problem: Don't use describeTransports. Pushed up https://github.com/bufbuild/connect-es/pull/668/commits/6369528f8d333649ede5f854b75ad3e160b88249 as an example.

You can also avoid using a server completely. Pushed up https://github.com/bufbuild/connect-es/pull/668/commits/c236657163c306b6c4b794276eadf9ed0b90af68 as an example. This one also covers error responses. Errors are always JSON, even when the request was made with the binary format. This is easily missed, so we should cover it.

Please use the examples as you see fit, but:
- Let's not put these tests into the crosstests directory: The directory is an implementation of the test cases defined in [bufbuild/connect-crosstest](https://github.com/bufbuild/connect-crosstest), and we can't just add one. 
- Let's also not complicate the crosstests by providing the protocol a transport uses. The goal of the crosstest is to prove independency of the protocol. `describeTransportsExcluding` is a sufficient escape hatch for cases like HTTP/1.1 and full-duplex.

PR #674, opened 2023-06-13
Ignore unknown JSON fields in Connect unary responses

smaye81:
This ignores unknown fields by default if the format is JSON when parsing a response. This was put into place in #642, but this is actually a special case that doesn't follow the code path in #642 and must be handled separately.

### DIFF
--- a/packages/connect-web/src/connect-transport.ts
+++ b/packages/connect-web/src/connect-transport.ts
@@ -36,6 +36,7 @@ import {
   createClientMethodSerializers,
   createEnvelopeReadableStream,
   createMethodUrl,
+  getJsonOptions,
   encodeEnvelope,
   runStreamingCall,
   runUnaryCall,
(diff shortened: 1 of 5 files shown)


packages/connect-web/src/connect-transport.ts
```diff
@@ -200,6 +200,14 @@ export function createConnectTransport(
...
             response.headers
           );
+
+          // Ignore unknown fields by default if we need to parse JSON format.
+          // Note this is done automatically via the createClientMethodSerializers call above,
+          // but this is a special path that doesn't use that function so we need to set it
+          // explicitly here.
+          if (options.jsonOptions) {
+            options.jsonOptions.ignoreUnknownFields ??= true;
+          }
```
timostamm, line 210: This code block does not run if no `jsonOptions` are given, and unknown fields will not be ignored.

packages/connect/src/protocol/serialization.ts
```diff
@@ -44,6 +44,17 @@ export interface Serialization<T> {
   parse: (data: Uint8Array) => T;
 }
 
+/**
+ * Returns a Json Options object with defaults set for any properties not provided
+ */
```
timostamm, line 49:
```suggestion
/**
 * Sets default JSON serialization options for connect-es.
 * 
 * With standard protobuf JSON serialization, unknown JSON fields are 
 * rejected by default. In connect-es, unknown JSON fields are ignored 
 * by default.
 */
```

timostamm (approved):
Just one suggestion for the JSDoc.

PR #681, opened 2023-06-20
Handle GOAWAY frames with NO_ERROR in Node.js clients

timostamm:
In [v0.10.0](https://github.com/bufbuild/connect-es/releases/tag/v0.10.0), we added support for basic keepalive for Node.js HTTP/2 clients (implemented in https://github.com/bufbuild/connect-es/pull/673). 

We made a mistake and expected all GOAWAY frames to close a client session, but that is not the case for the HTTP/2 error code NO_ERROR:

> NO_ERROR (0x0):  The associated condition is not a result of an error.  For example, a GOAWAY might include this code to indicate graceful shutdown of a connection.

The graceful shutdown is further specified:

> GOAWAY allows an endpoint to gracefully stop accepting new streams while still finishing processing of previously established streams. This enables administrative actions, like server maintenance.

The Node.js `http2` module implements this behavior, and raises an error when new request is issued on a session. This is the issue we are seeing with https://github.com/bufbuild/connect-es/issues/680.

This PR fixes https://github.com/bufbuild/connect-es/issues/680 by opening a new connection if the current one has received a GOAWAY with NO_ERROR. Connections that are gracefully shutting down are maintained until all streams have finished.

### DIFF
--- a/packages/connect-node/src/http2-session-manager.ts
+++ b/packages/connect-node/src/http2-session-manager.ts
@@ -72,13 +72,16 @@ export interface Http2SessionOptions {
 }
 
 /**
- * Manage a single HTTP/2 connection and keep it alive with PING frames.
+ * Manage an HTTP/2 connection and keep it alive with PING frames.
  *
  * The logic is based on "Basic Keepalive" described in
  * https://github.com/grpc/proposal/blob/0ba0c1905050525f9b0aee46f3f23c8e1e515489/A8-client-side-keepalive.md#basic-keepalive
  * as well as the client channel arguments described in
  * https://github.com/grpc/grpc/blob/8e137e524a1b1da7bbf4603662876d5719563b57/doc/keepalive.md
  *
+ * Usually, the managers tracks exactly one connection, but if a connection
+ * receives a GOAWAY frame with NO_ERROR, the connection is maintained until
+ * all streams have finished, and new requests will open a new connection.
  */
 export class Http2SessionManager {
   /**
@@ -276,7 +279,9 @@ export class Http2SessionManager {
       | StateVerifying
       | StateReady
   ): void {
+    this.s.onExitState?.();
     if (this.s.t == "ready" && this.s.isShuttingDown()) {
+      // Maintain connections that have been asked to shut down.
       const sd = this.s;
       this.shuttingDown.push(sd);
       sd.onClose = sd.onError = () => {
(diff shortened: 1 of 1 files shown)


packages/connect-node/src/http2-session-manager.ts
```diff
@@ -653,23 +686,40 @@ function ready(
...
+      if (streamCount == 0 && nodeMajor >= 18) {
+        conn.destroy(
+          new ConnectError(
+            "received GOAWAY without any open streams",
+            Code.Canceled
+          ),
+          http2.constants.NGHTTP2_NO_ERROR
+        );
+      }
+    }
```
jeanp413, line 722: In case it's relevant:
- grpc.js seems to always close the connection https://github.com/grpc/grpc-node/blob/186718cb70af708c75584909b21169289d7363b2/packages/grpc-js/src/transport.ts#L198-L212
- the server will also do the same, close the connection after sending `goaway` https://github.com/grpc/proposal/blob/master/A8-client-side-keepalive.md#server-enforcement

jeanp413: >     // Node.js v16 closes a connection on its own when it receives a GOAWAY
>     // frame and there are no open streams (emitting a "close" event and
>     // destroying the session), but more recent versions do not.

Does it mean the bug shouldn't happen in node 16? I'm using node 16 and I can confirm it does happen

timostamm: I'm not sure if `@grpc/grpc-js` is really _always_ closing the connection when receiving a GOAWAY. I've updated the PR description with details about the graceful shutdown. I would expect an HTTP/2 client to drain streams first. It is possible that `@grpc/grpc-js` does not do so for historical reasons.

Note that there is a distinction between error code NO_ERROR and actual error codes, and also the special behavior in case of ENHANCE_YOUR_CALM (last paragraph in [basic keepalive](https://github.com/grpc/proposal/blob/master/A8-client-side-keepalive.md#basic-keepalive)).

Regarding the comment: The `http2` module implements graceful shutdown. It rejects new streams, and it automatically closes the connection after all streams have finished. But if a connection _without_ open streams receives a GOAWAY with NO_ERROR, behavior differs between versions. Seeing https://github.com/bufbuild/connect-es/issues/680 in v16 is expected.

smaye81 (approved):

PR #687, opened 2023-06-22
Assert that the HTTP/2 connection is open

timostamm:
Assert that the HTTP/2 connection is open when entering the "ready" state in the connection manager.
This will help isolate the cause of https://github.com/bufbuild/connect-es/issues/683.


packages/connect-node/src/http2-session-manager.ts
```diff
@@ -305,11 +305,7 @@ export class Http2SessionManager {
       case "verifying":
         state.verified.then(
           (value) => {
-            if ("t" in value) {
-              this.setState(value);
-            } else {
-              this.setState(ready(value, this.options));
-            }
+            this.setState(value);
```
timostamm, line 308: Unrelated cleanup. This was a remnant from earlier in the implementation, when the promise resolved to a state or to a http2 connection that is no longer necessary. The else block is dead code.

srikrsna-buf (approved):

PR #692, opened 2023-06-26
Defer resolving globalThis.fetch until calling endpoint

paul-sachs:
When using a polyfilled (or patched) fetch, we need to make sure the fetch is retrieved at call site, rather than being cached earlier. This allows tools like Sentry to properly detect fetch calls.

### DIFF
--- a/packages/connect-web-test/src/fetch.spec.ts
+++ b/packages/connect-web-test/src/fetch.spec.ts
@@ -107,5 +107,31 @@ describe("custom fetch", function () {
       expect(response.json).toHaveBeenCalledTimes(1); // eslint-disable-line @typescript-eslint/unbound-method
       expect(response.arrayBuffer).toHaveBeenCalledTimes(0); // eslint-disable-line @typescript-eslint/unbound-method
     });
+    it("should support polyfills", async function () {
+      const response = new Response(
+        new SimpleResponse({ username: "donald" }).toJsonString(),
+        {
+          headers: {
+            "Content-Type": "application/json",
+          },
+        }
+      );
+      spyOn(response, "arrayBuffer").and.callThrough();
+      spyOn(response, "json").and.callThrough();
+      const transport = createConnectTransport({
+        baseUrl: "https://example.com",
+      });
+      globalThis.fetch = () => Promise.resolve(response);
+      await transport.unary(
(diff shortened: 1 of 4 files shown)


packages/connect-web/src/grpc-web-transport.ts
```diff
@@ -157,6 +156,9 @@ export function createGrpcWebTransport(
           message: normalize(message),
         },
         next: async (req: UnaryRequest<I, O>): Promise<UnaryResponse<I, O>> => {
+          // Make sure to retrieve the fetch on request to make sure any polyfills/overrides
+          // are applied.
+          const fetch = options.fetch ?? globalThis.fetch;
```
timostamm, line 161: `stream()` below will need the same change.

packages/connect-web-test/src/fetch.spec.ts
```diff
@@ -107,5 +107,34 @@ describe("custom fetch", function () {
       expect(response.json).toHaveBeenCalledTimes(1); // eslint-disable-line @typescript-eslint/unbound-method
       expect(response.arrayBuffer).toHaveBeenCalledTimes(0); // eslint-disable-line @typescript-eslint/unbound-method
     });
+    it("should support polyfills", async function () {
```
timostamm, line 110: I wondered for a second what this means. The point is that we don't want to resolve a reference to `globalThis.fetch` _before_ a patch or polyfill has been applied, right?

This makes sense, good call 👍 

Can we find a more descriptive PR title and test name? And can we cover gRPC-web (at least unary) as well?

timostamm (commented):
Good call, but left a couple of comments. You have to run `make bench` to update the code size benchmark too.

paul-sachs:
> You have to run `make bench` to update the code size benchmark too.

I ran `make bench` and updated the readme but CI continues to fail due to a change in the size. Is there a specific version of node I need to run it with?

paul-sachs:
@timostamm Yeah, that makes a lot of sense. I was seeing that weirdness and just started to mess with beforeEach/afterEach but was looking into the GRPC thing first. I was kinda struggling to mock it out so the error idea is super helpful, thanks!

timostamm (approved):

PR #695, opened 2023-06-28
Disregard non-JSON error response bodies

smaye81:
This PR modifies the response validation so that Connect will only attempt to parse the body of a unary error if the content type is `application/json`. 

For all other error responses, the body is ignored and the code and message is built from the received HTTP error status and it's corresponding Connect error code.


packages/connect/src/protocol-connect/validate-response.spec.ts
```diff
@@ -15,51 +15,85 @@
...
-        validateResponse(methodKind, 400, new Headers());
-        fail("expected error");
+        validateResponse(
+          MethodKind.Unary,
+          400,
+          new Headers({
+            "Content-Type": "application/proto",
+          })
+        );
       } catch (e) {
```
srikrsna-buf, line 66: This test will pass even if validate response never throws. It is missing a `fail`

smaye81: Fixed. Thanks for catching that

srikrsna-buf (approved):

PR #712, opened 2023-07-06
Use built-in properties to avoid incorrect maintained conn state

mustard-mh:
This PR fixes the issue where the connection state maintained in lib was not identical to http2, resulting in no new connections being created. see also https://github.com/bufbuild/connect-es/issues/683#issuecomment-1602581662

We use `conn.destroyed` and `conn.closed` since those properties are exposed in http2 core, and where they throw `The session has been destroyed` [^1] are all checking with those properties. 

And properties are compatible [[http2/core v16.0](https://github.com/nodejs/node/blob/v16.0.0/lib/internal/http2/core.js#L1291-L1299)] [[connect-node min support v16.0](https://github.com/mustard-mh/connect-es/blob/main/packages/connect-node/package.json#L25-L27)]

[^1]: search `ERR_HTTP2_INVALID_SESSION` in `http2/core`, they all check `if (this.destroyed) { throw }`

## Related Issues
Fix https://github.com/bufbuild/connect-es/issues/680, https://github.com/bufbuild/connect-es/issues/683

## How to test
No idea, but we released a patch fork in npm and use it in our product [[1](https://github.com/gitpod-io/gitpod-vscode-desktop/blob/1b92e4a3809a339689e9740f0b4313b1454219f4/package.json#L188)], errors are gone


akosyakov:
@timostamm I can confirm no more GOAWAY and session has been destroyed errors. Our SLO went up from 90 to 97%. We were running with this change for last 5 days.

packages/connect-node/src/http2-session-manager.ts
```diff
@@ -246,7 +246,7 @@ export class Http2SessionManager {
 
   private async gotoReady() {
     if (this.s.t == "ready") {
-      if (this.s.isShuttingDown()) {
+      if (this.s.conn.destroyed || this.s.conn.closed || this.s.isShuttingDown()) {
```
timostamm, line 249: This will avoid the problem found in https://github.com/bufbuild/connect-es/issues/683#issuecomment-1648209277. I think we can simply put _all_ connections that received any GOAWAY frame into the "shutting down" state. We will need a test.

packages/connect-node/src/http2-session-manager.ts
```diff
@@ -714,15 +714,12 @@ function ready(
...
     if (errorCode === http2.constants.NGHTTP2_NO_ERROR) {
       receivedGoAwayNoError = true;
-      const nodeMajor = parseInt(process.versions.node.split(".")[0], 10);
-      // Node.js v16 closes a connection on its own when it receives a GOAWAY
-      // frame and there are no open streams (emitting a "close" event and
-      // destroying the session), but more recent versions do not.
-      // Calling close() ourselves is ineffective here - it appears that the
-      // method is already being called, see https://github.com/nodejs/node/blob/198affc63973805ce5102d246f6b7822be57f5fc/lib/internal/http2/core.js#L681
-      if (streamCount == 0 && nodeMajor >= 18) {
+      if (streamCount == 0) {
```
timostamm, line 722: As we found out in https://github.com/bufbuild/connect-es/issues/683#issuecomment-1648209277, Node.js _does_ close connections when it receives a GOAWAY with any error, but it does so after request streams have closed.

The special casing for v18 and above is because behavior differs for GOAWAY with NO_ERROR for connections without streams:

```bash
$ node -v
v16.20.0
$ node repro-goaway-close.mjs
http2.connect
conn on connect
conn on goaway, code: 0
conn on close
^C
$ node -v
v18.17.0
$ node repro-goaway-close.mjs
http2.connect
conn on connect
conn on goaway, code: 0
^C
```


<details>
<summary>repro-683.mjs</summary>

```js
import * as http2 from "http2";

// start a server that sends a GOAWAY with NO_ERROR to every session
const PORT = 8088;
const server = http2.createServer()
  .on("session", (sess) => {
    setTimeout(() => {
      sess.goaway(http2.constants.NGHTTP2_NO_ERROR)
    }, 100);
  })
  .on("request", () => {
    //
  });
await new Promise(resolve => {
  server.listen(PORT, () => resolve());
});


// open a connection
const conn = await new Promise((resolve, reject) => {
  console.log("http2.connect");
  const conn = http2.connect(`http://localhost:${PORT}`);
  conn.on("error", reject);
  conn.on("connect", () => {
    console.log("conn on connect");
    resolve(conn);
  });
  conn.on("goaway", errorCode => {
    console.log("conn on goaway, code:", errorCode);
  });
  conn.on("close", () => {
    console.log("conn on close")
  });
  conn.on("error", e => {
    console.log("conn on error", String(e));
  });
});
```
</details>

Note that v16 emits "close", v18 does not. By calling `conn.destroy()` ourselves, we force emit "error" and "close" events. You would think that calling `conn.close()` would help, but it does not. That is what the comment was referring to. We should not change this part.

timostamm (commented):
Hey @mustard-mh, thank you for the PR. I'm glad to hear the patch brought down your error numbers, but I hope you understand we're hesitant to merge without understanding what is actually happening.

packages/connect-node/src/http2-session-manager.ts
timostamm: Reverted in https://github.com/bufbuild/connect-es/pull/712/commits/4df734e6c49c33105f2ded8971cab13dbe43fbae

packages/connect-node/src/http2-session-manager.ts
timostamm: Pushed up a test in https://github.com/bufbuild/connect-es/pull/712/commits/bf4d7469304bcd4173dd05855fbcb5df31412ac8. New requests open a new connection 👍  

But the test framework reports an uncaught exception, see [run log](https://github.com/bufbuild/connect-es/actions/runs/5656284377/job/15323197715#step:5:201). That's because we remove event listeners for the connection here: https://github.com/bufbuild/connect-es/blob/4d1580fe8769a07be4aeb25f2eb4a411e90861f0/packages/connect-node/src/http2-session-manager.ts#L634-L644

Pushed up https://github.com/bufbuild/connect-es/pull/712/commits/a15680d5cba3187d5c72479a19131a9618119d70 to use isShuttingDown for all GOAWAY frames. This means we are deferring the cleanup in onExitState, and pass the test without uncaught exception.

timostamm (approved):
@mustard-mh, pushed up a test and a change that moves the logic into isShuttingDown because to not leave uncaught errors. That way, you're still credited for your contribution.

srikrsna-buf (approved):

timostamm:
CI failure is just browserstack, which does not have credentials in the fork. Since this does not affect any code tested in browserstack, we're good.

mustard-mh:
Thank you @timostamm :pray: , we will try new release on our side

akosyakov:
@timostamm We tried and unfortunately had to revert back to our fork, i.e. original @mustard-mh changes. We still see `The session has been destroyed`.

PR #724, opened 2023-07-20
Update the `WritableIterable` behavior

srikrsna-buf:
Update the `WritableIterable` behavior to improve the developer experience. Writes behave as if they are synchronous and resolve when there is one more `next` call (indicating the previous one is processed) or rejected when consumers call `throw`.

### DIFF
--- a/packages/connect/src/protocol/async-iterable.ts
+++ b/packages/connect/src/protocol/async-iterable.ts
@@ -12,7 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import type { Message, MessageType, PartialMessage } from "@bufbuild/protobuf";
 import { Code } from "../code.js";
 import { ConnectError } from "../connect-error.js";
 import type { EnvelopedMessage } from "./envelope.js";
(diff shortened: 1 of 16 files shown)


packages/connect/src/protocol/async-iterable-story.spec.ts
```diff
@@ -194,22 +194,19 @@ describe("full story", function () {
...
-      void writer.close();
+      writer.close();
 
       writer
         .write({ value: "alpha", end: false })
         .catch((e) =>
-          expect(e).toEqual(new ConnectError("cannot write, already closed"))
+          expect(e).toEqual(
+            new Error("cannot write, WritableIterable already closed")
+          )
```
srikrsna-buf, line 209: Changed it to a regular `Error` vs a `ConnectError` as this is a generic `WritableIterable` and doesn't really have anything do with connect not to mention the error is also not at all connect related.

packages/connect/src/protocol/async-iterable-story.spec.ts
```diff
@@ -265,11 +262,16 @@ describe("full story", function () {
...
       readerIt[Symbol.asyncIterator]()
         .next()
-        .then(() => fail("reads were unexpectedly resolved"))
-        .catch((e) => expect(e).toBe("READER_ERROR"));
+        .then((result) =>
+          expect(result).toEqual({ done: true, value: undefined })
+        )
+        .catch(() =>
+          fail("expected successful done result but unexpectedly rejected")
+        );
```
srikrsna-buf, line 274: Changed the behavior for next to return `{done: true}` when the writer is closed either due to an error or with call of `close`. This is what I think should be the behavior of Iterables outlined [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_iterator_protocol), based on these two paragraphs:
>If an iterator returns a result with done: true, any subsequent calls to next() are expected to return done: true as well, although this is not enforced on the language level.

>throw(exception) Optional
A function that accepts zero or one argument and returns an object conforming to the IteratorResult interface, typically with done equal to true. Calling this method tells the iterator that the caller detects an error condition, and exception is typically an [Error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error) instance.

packages/connect/src/protocol/async-iterable.ts
```diff
@@ -1283,107 +1283,123 @@ export function makeIterableAbortable<T>(
...
   write: (payload: T) => Promise<void>;
-  close: () => Promise<void>;
-  isClosed: () => boolean;
+  /**
+   * Closes the writer indicating to its consumer that no further
+   * payloads will be received.
+   *
+   * Any writes that happen after close is called will return an error.
+   */
+  close: () => void;
```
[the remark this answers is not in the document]
timostamm: Agreed, especially since this no longer raises an error for close() on a closed writable. 👍

timostamm (approved):
👍  Thank you for the work on this.

Once we are sure the DX is solid, we should export the writable from @bufbuild/connect.

PR #726, opened 2023-07-21
Fix memory leak for 'error' on session during h2 requests

timostamm:
Fixes https://github.com/bufbuild/connect-es/issues/725 by removing the event listener completely. All session events are handled by the `Http2SessionManager` since https://github.com/bufbuild/connect-es/pull/673, and I can't think of a reason we would need it here.

The reproduction for the leak from https://github.com/bufbuild/connect-es/issues/725#issuecomment-1645429051 passes now:

```bash
$ node -v
v18.17.0
$ node --trace-warnings repro-725.mjs
req 0
req 1
req 2
req 3
req 4
req 5
req 6
req 7
req 8
req 9
req 10
req 11
```

### DIFF
--- a/packages/connect-node/src/node-universal-client.ts
+++ b/packages/connect-node/src/node-universal-client.ts
@@ -289,32 +289,22 @@ function h2Request(
   }
   sm.request(method, requestUrl.pathname + requestUrl.search, headers, {}).then(
     (stream) => {
-      stream.session.on("error", sentinel.reject);
-
-      sentinel
-        .catch((reason) => {
-          if (stream.closed) {
-            return;
-          }
-          // Node.js http2 streams that are aborted via an AbortSignal close with
-          // an RST_STREAM with code INTERNAL_ERROR.
-          // To comply with the mapping between gRPC and HTTP/2 codes, we need to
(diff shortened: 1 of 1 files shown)


packages/connect-node/src/node-universal-client.ts
```diff
@@ -289,32 +289,22 @@ function h2Request(
...
-              : H2Code.INTERNAL_ERROR;
-          return new Promise<void>((resolve) => stream.close(rstCode, resolve));
-        })
-        .finally(() => {
-          stream.session.off("error", sentinel.reject);
-        })
-        .catch(() => {
-          // We intentionally swallow sentinel rejection - errors must
-          // propagate through the request or response iterables.
-        });
```
timostamm, line 317: This was originally catching rejections from a promise with just a `finally` handler - basically:

```ts
Promise.reject("err")
  .finally(() => {
    // just observing
  })
  .catch(() => {
   // swallow unhandled rejection passing through `finally`
  })
```

We already have a `catch` now, so this was not useful anymore

gilwong00 (approved):

PR #729, opened 2023-07-25
Fix edge cases with `AsyncIterable` protocol

srikrsna-buf:
This PR addresses a couple of issues with `AsyncIterable` usage:
* Call and propagate `AsyncIterator`'s `return` for request iterables. This ensures request iterables are cleaned up properly when server returns early or with an error.
* Tweaks the `Transport`'s `stream` method to accept `PartialMessage` instead of concrete ones similar to the unary method. This helps us normalize messages in one place (`run-call.ts`).
* Omit `return`/`throw` from the response iterables returned by transports/clients.
* For `bidi`, start the request as soon as the client calls the method as opposed to waiting for the first read on the response. This is required for #724 to work properly.
* Extend tests for promise client and run call functions to cover these cases.


srikrsna-buf (commented):
This still doesn't work in the case where return is called on the response iterable:
```ts
const values = [123, 456, 789];

const input = createAsyncIterable(
  values.map((value) => new Int32Value({ value }))
);

const transport = createRouterTransport(({ service }) => {
  service(TestService, {
    bidiStream: async function* (input: AsyncIterable<Int32Value>) {
      for await (const next of input) {
        yield { value: `yield ${next.value}` };
      }
    },
  });
});
const fn = createBiDiStreamingFn(
  transport,
  TestService,
  TestService.methods.bidiStream
);
const it = fn(input)[Symbol.asyncIterator]();
expect(await it.next()).toEqual({
  done: false,
  value: new StringValue({ value: "yield 123" }),
});
expect(it.return).toBeDefined();
expect(await it.return?.()).toEqual({ done: true, value: undefined });
expect(await input[Symbol.asyncIterator]().next()).toEqual({
  done: true,
  value: undefined,
});
```

packages/connect/src/transport.ts
```diff
@@ -50,6 +50,6 @@ export interface Transport {
     signal: AbortSignal | undefined,
     timeoutMs: number | undefined,
     header: HeadersInit | undefined,
-    input: AsyncIterable<I>
+    input: AsyncIterable<PartialMessage<I>>
```
timostamm, line 53: A breaking change for implementors of `Transport`, but stream() should have accepted partial messages in the first place.

timostamm (approved):
Thanks for the cleanup. Let's do some smoke tests to verify the DX and follow up with changes as necessary, but definitely LGTM.

PR #746, opened 2023-08-07
Handle multiple set-cookie headers from fetch API Headers object

### LINKED ISSUE #745: Unable to set multiple cookies from within service method handler
**Describe the bug**

We are using @bufbuild/connect-express to connect to an express application, but cannot create multiple set-cookie headers to enable browsers to set multiple cookies based on response. For this header in particular, we need the ability to set separate headers with same key instead of combining into single value.

**To Reproduce**

Simply try appending multiple values to `ctx.responseHeader` from within a service handler, and only one header gets sets. Trying to set once with multiple values fails as well.

- @bufbuild/connect-web version: `0.9.1`
- @bufbuild/connect-node version: `0.9.1`
- Frontend framework and version: `angular@14.1.0`
- Node.js version: `18.16.0`
- Browser and version:  `Google Chrome 115.0.5790.170`

tcarnes:
This is a possible solution for special handling needed to properly propagate `set-cookie` headers back to browser so that multiple cookies will be set. Fixes #745

### DIFF
--- a/packages/connect-node/src/node-universal-header.ts
+++ b/packages/connect-node/src/node-universal-header.ts
@@ -71,31 +71,30 @@ export function webHeaderToNodeHeaders(
     return undefined;
   }
   const o = Object.create(null) as http.OutgoingHttpHeaders;
+  const append = (key: string, value: string): void => {
+    key = key.toLowerCase();
+    const existing = o[key];
+    if (typeof existing == "string") {
+      o[key] = [existing, value];
+    } else if (Array.isArray(existing)) {
+      existing.push(value);
+    } else {
+      o[key] = value;
+    }
+  };
   if (Array.isArray(headersInit)) {
     for (const [key, value] of headersInit) {
-      const k = key.toLowerCase();
-      o[k] = value;
+      append(key, value);
     }
   } else if ("forEach" in headersInit) {
     if (typeof headersInit.forEach == "function") {
-        const setCookieHeaders: string[] = [];
-        headersInit.forEach((value, key) => {
-            const k = key.toLowerCase();
-            if (k === 'set-cookie') {
-                setCookieHeaders.push(value);
-            }
-            else {
-                o[k] = value;
-            }
-        });
-        if (setCookieHeaders.length > 0) {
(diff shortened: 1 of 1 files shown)


packages/connect-node/src/node-universal-header.ts
```diff
@@ -78,10 +78,19 @@ export function webHeaderToNodeHeaders(
...
+            if (k === 'set-cookie') {
+                setCookieHeaders.push(value);
+            }
+            else {
+                o[k] = value;
+            }
+        });
+        if (setCookieHeaders.length > 0) {
+            o['set-cookie'] = setCookieHeaders;
+        }
```
timostamm, line 93: This works, but I believe we have to handle it for the array (line 74) and the object literal (line 95) as well.

What do you think about adding a local function that adds multiple entries for the same key as an array?

```ts
  const append = (key: string, value: string): void => {
    key = key.toLowerCase();
    const existing = o[key];
    if (typeof existing == "string") {
      o[key] = [existing, value];
    } else if (Array.isArray(existing)) {
      existing.push(value);
    } else {
      o[key] = value;
    }
  };
```

In the forEach callback, you would simply call `append(key, value)`, and you could do the same for array and object literal.

tcarnes: Yes, that would make sense, but I believe those other two cases are intended to handle the header objects from the http and http2 packages? In both of those, you have option to set cookie value that is an array of strings, which will give the desired effect for set-cookie here. In particular, it's from the fetch API Headers object (which is used for headers in connect-express), where we don't have a way to create an array value where we need this special handling to create the array for us. The fetch API even introduced a special getSetCookie method for this purpose, which we could also use here instead, but it would cause a typing error: https://github.com/microsoft/TypeScript/issues/55270

Very happy with whatever way you'd prefer to structure this, but main thing that would fill the gap for us is something that handles multiple set-cookie headers when using fetch api.

tcarnes: Great point, addressed with last commit.

timostamm (approved):
Thank you for the contribution, Tim 🙂

PR #751, opened 2023-08-10
Propagate errors back to the handler

srikrsna-buf:
Propagate errors back to the handler.


timostamm (commented):
Apologies for taking so long to review. This looks great. So the only missing piece is making sure that the HanderContext's signal is triggered in case of an error.

Just one comment:

packages/connect/src/protocol/normalize.ts
```diff
@@ -0,0 +1,76 @@
...
+    return new type(message);
+  } catch (e) {
+    throw new ConnectError(
+      `failed to normalize message ${type.typeName}`,
+      Code.Internal,
+      undefined,
+      undefined,
+      e,
+    );
+  }
```
timostamm, line 41: It's certainly not the most important thing - correctness is certainly more important - but 3/4 of the bundle size increase of this PR come from re-throwing the error here. Do you have a specific code path in mind where this could occur? We typically wouldn't recommend to wrap constructor calls with a try-catch.

timostamm: Ah, we had two version before. Yes, let's remove the re-throw.

timostamm (approved):
LGTM!

PR #769, opened 2023-08-15
Add migration cli

paul-sachs:
Created a CLI that migrates from old packages to new.

- Using jscodeshift to guarantee that we are only modifying imports and not arbitrary code
- Testing with jest since jscodeshift have a bunch of helpers to use with jest. We can migrate those to jasmine for consistency


packages/connect-migrate/src/cli.ts
```diff
@@ -0,0 +1,98 @@
...
+    packageManager = "pnpm";
+  } else if (existsSync(path.join(dir, "package-lock.json"))) {
+    packageManager = "npm";
+  } else if (existsSync(path.join(dir, "yarn.lock"))) {
+    packageManager = "yarn";
+  }
+  assert(
+    packageManager !== undefined,
+    "Could not determine package manager based on lock file. Make sure you have a lock file (yarn.lock, package-lock.json, or pnpm-lock.yaml) in your project.",
+  );
```
timostamm, line 63: We cannot use assertions for flow control. It's easy to run the command in the wrong directory, and we have to consider it to be part of the regular flow, not an exceptional case.

In general, command line tools have to:
- exit with status 0 on success
- only crash with stack trace on an exceptional error
- print a short error message (no stack traces) and usage information on stderr and exit with non-zero code for invalid arguments
- print a short error message and exit with non-zero code in case of an error
- print usage information for the flag `-h` or `--help`
- print the version for the flag `--version` (or `-v` unless taken for verbose output)

Here is an example for parsing a command line:

```ts
interface CommandLineArgs {
  ok: true;
  help: boolean;
  version: boolean;
}

interface CommandLineError {
  ok: false;
  errorMessage: string;
}

export function parseCommandLineArgs(args: string[]): CommandLineArgs | CommandLineError {
  const parsed = {
    version: false,
    help: false
  };
  while (args.length > 0) {
    const arg = args.shift();
    switch (arg) {
      case "-h":
      case "--help":
        if (parsed.help) {
          break;
        }
        parsed.help = true;
        break;
      case "--version":
        parsed.version = true;
        break;
      default:
        return {
          ok: false,
          errorMessage: `unknown argument ${arg}`
        };
    }
  }
  return {...parsed, ok: true};
}
```

Used as:

```ts
import { parseCommandLineArgs } from "./arguments";
import process from "node:process";

const args = parseCommandLineArgs(process.argv.slice(2));
if (!args.ok) {
  process.stderr.write(`${args.errorMessage}\n`);
  process.exit(1);
}
if (args.version) {
  // todo: use version from package.json
  process.stdout.write("connect-migrate v0.1.0\n");
  process.exit(0);
}
if (args.help) {
  process.stdout.write(`USAGE: connect-migrate\n`);
  process.stdout.write(`Updates references to connect-es packages in your project to use @connectrpc.\n`);
  process.exit(0);
}

// todo: logic
process.exit(0);
```

This scales up to everything we'll need here, and makes testing trivial.

packages/connect-migrate/src/replacement-map.ts
```diff
@@ -0,0 +1,51 @@
...
+export function getReplacementImport(sourceValue: string): string | undefined {
+  const match = keys.find((key) => {
+    return sourceValue.match(new RegExp(`^${key}\\b`));
+  });
+  if (match !== undefined) {
+    return sourceValue.replace(
+      new RegExp(`^${match}\\b`),
+      replacementMap[match],
+    );
+  }
```
timostamm, line 38: Constructing regexp dynamically is fragile 😅

Can't we simply do?

```suggestion
  if (sourceValue.startsWith("@bufbuild/connect")) {
    return "@connectrpc/connnect" + sourceValue.substring("@bufbuild/connect".length);
  }
  return undefined;
```

packages/connect-migrate/src/replacement-map.ts
```diff
@@ -0,0 +1,51 @@
...
+  return undefined;
+}
+
+export function replacePackageJSONReferences(jsonString: string): string {
+  let result = jsonString;
+  for (const key of keys) {
+    result = result.replace(
+      new RegExp(`"${key}":`, "g"),
+      `"${replacementMap[key]}":`,
+    );
```
timostamm, line 48: Let's avoid constructing regex dynamically 😅 

WDYT about

```suggestion
    result = result.replace(
      "${key}", `"${replacementMap[key]}"`,
    );
```

? 

Deliberately omitted the colon, it's valid JSON to have a space between.

packages/connect-migrate/src/transforms/__testfixtures__/modify-imports/sub-directory-imports.input.ts
```diff
@@ -0,0 +1,3 @@
+import { makeAnyClient } from "@bufbuild/connect/dist/cjs/any-client";
```
timostamm, line 1: Not a valid import. Let's use

```suggestion
import { runUnaryCall } from "@bufbuild/connect/protocol";
```

packages/connect-migrate/src/transforms/__testfixtures__/modify-imports/simple.input.ts
```diff
@@ -0,0 +1,4 @@
+import { createPromiseClient } from "@bufbuild/connect";
```
timostamm, line 1: Single quotes will be very common. Are they preserved?

packages/connect-migrate/src/cli.ts
```diff
@@ -0,0 +1,98 @@
...
+    ignore: ["**/node_modules/**"],
+  });
+  for (const file of files) {
+    console.log(`Updating ${file}...`);
+    const fileContents = await readFile(file, "utf-8");
+    await writeFile(file, replacePackageJSONReferences(fileContents), "utf-8");
+  }
+  console.log("Updated package dependencies.");
+  console.log("Updating references...");
+  executeInContext(jscodeshiftExecutable, [
```
timostamm, line 76: Does jscodeshift have a programmatic API? It would be nice to avoid spawning a process...

timostamm (commented):
This looks very promising, now we just have to sort out the details. 

Blocking on jest, we already have enough dependencies. I'm okay with replacing it later, when we know better what needs testing.

paul-sachs:
@timostamm all great points, i had started on this tool separate from connect-es, so i'll migrate the packages/testing to something more akin to what's in connectrpc already. This was mostly just porting things over as is. I'll tune it up.

packages/connect-migrate/src/cli.ts
paul-sachs: It does, we can combine that directly with fastGlob and should be able to use that.

packages/connect-migrate/src/cli.ts
```diff
@@ -0,0 +1,98 @@
...
+    "--ignore-pattern=**/node_modules/**",
```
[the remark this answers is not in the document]
paul-sachs: Thankfully the output of jscodeshift is pretty thorough so it'll be obvious what files have changed. I'll check to make sure that if we call it programatically, we still get that output.

packages/connect-migrate/src/replacement-map.ts
paul-sachs: ah good call.

packages/connect-migrate/src/transforms/__testfixtures__/modify-imports/sub-directory-imports.input.ts
paul-sachs: yeah at this point I was just tossing some random cases and not considering valid imports. I can modify

packages/connect-migrate/src/transforms/__testfixtures__/modify-imports/simple.input.ts
paul-sachs: Just pushed some code up to detect the style used (based on the first import detected) and we maintain that style.

packages/connect-migrate/src/replacement-map.ts
```diff
@@ -0,0 +1,51 @@
...
+  if (match !== undefined) {
+    return sourceValue.replace(
+      new RegExp(`^${match}\\b`),
+      replacementMap[match],
+    );
+  }
+  return undefined;
+}
+
+export function replacePackageJSONReferences(jsonString: string): string {
```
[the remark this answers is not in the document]
paul-sachs: Hmm good call, I'll add a warning if install fails.

packages/connect-migrate/src/replacement-map.ts
paul-sachs: And this solution also prevents us from accidentally modifying packages we don't want to. I guess I just wanted to be very specific about which packages I replace, and the provided map of replacement packages makes that very explicit.

packages/connect-migrate/src/cli.ts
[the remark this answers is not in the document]
paul-sachs: Added a section to the readme to indicate which files are changed. That probably should suffice for now.

packages/connect-migrate/src/replacement-map.ts
paul-sachs: Actually, found a different way to do this without relying on dynamic regexp. Updated.

packages/connect-migrate/src/transforms/modify-imports.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+import { readFileSync, readdirSync } from "node:fs";
+import jscodeshift from "jscodeshift/src/core";
+import transform from "./modify-imports";
+
+// Make sure to import the testfixtures from source so we're not comparing transformed fixtures
+const fixtureLocation = join(
+  __dirname,
+  "../../../src/transforms",
+  "__testfixtures__",
+);
```
timostamm, line 25: I looked into ways to run codeshift the same way in tests and in the CLI, but the workers do give us a significant bump in performance for large numbers of files.

Since we can't run the same code paths in tests as we do in the CLI, we can just as well remove the `__testfixtures__` directory and work with strings here in the tests. 

The we end up with something like this:

```ts
import jscodeshift from "jscodeshift";
import transform from "./modify-imports.js";

function t(source: string) {
  const shift = jscodeshift.withParser("tsx");
  return transform(
    { path: "test-file", source },
    {
      jscodeshift: shift,
      j: shift,
      stats: () => {},
      report: () => {},
    },
    {},
  );
}

describe("modify-imports", () => {
  it("should modify import", () => {
    const got = `import a from "@bufbuild/connect";`;
    const want = `import a from "@connectrpc/connect";`;
    expect(t(got)?.trim()).toBe(want.trim());
  });
  it("should modify import with single quotes", () => {
    const got = `import a from '@bufbuild/connect';`;
    const want = `import a from '@connectrpc/connect';`;
    expect(t(got)?.trim()).toBe(want.trim());
  });
  it("should modify subpath import", () => {
    const got = `import a from "@bufbuild/connect/protocol";`;
    const want = `import a from "@connectrpc/connect/protocol";`;
    expect(t(got)?.trim()).toBe(want.trim());
  });
  it("should not modify irrelevant import", () => {
    const got = `import a from "@foobar/connect";`;
    const want = `import a from "@foobar/connect";`;
    expect(t(got)?.trim()).toBe(want.trim());
  });
  it("should modify require", () => {
    const got = `const a = require("@bufbuild/connect");`;
    const want = `const a = require("@connectrpc/connect");`;
    expect(t(got)?.trim()).toBe(want.trim());
  });
  it("should not modify irrelevant require", () => {
    const got = `const a = require("@foobar/connect");`;
    const want = `const a = require("@foobar/connect");`;
    expect(t(got)?.trim()).toBe(want.trim());
  });
  it("should modify require with single quotes", () => {
    const got = `const a = require('@bufbuild/connect');`;
    const want = `const a = require('@connectrpc/connect');`;
    expect(t(got)?.trim()).toBe(want.trim());
  });
});
```

I believe we don't lose any confidence with these tests (it's clear now that the file name is irrelevant), but we can remove a lot of files and ignores.

What do you think?

packages/connect-migrate/src/cli.ts
```diff
@@ -0,0 +1,132 @@
...
+  process.stdout.write("Dependencies installed.\n");
+
+  process.exit(0);
+}
+
+void main();
+
+function reinstallDependencies(dir: string) {
+  // TODO: Perhaps we can find all the lock files and run install wherever we find them. This'll help for repos without a single
+  // top level lock file.
```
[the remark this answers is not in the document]
paul-sachs: Yeah actually working on this right now, mostly to ease migration of examples.

packages/connect-migrate/src/arguments.ts
```diff
@@ -0,0 +1,57 @@
...
+interface CommandLineError {
+  ok: false;
+  errorMessage: string;
+}
+
+const usage = `USAGE: connect-migrate
+Updates references to connect-es packages in your project to use @connectrpc.
+`;
+
+export function parseCommandLineArgs(
```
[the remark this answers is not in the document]
paul-sachs: Yeah, alright, I liked having the help colocated with the argument parsing, will adjust.

packages/connect-migrate/src/transforms/modify-imports.spec.ts
paul-sachs: Yeah that sounds fine to me.

timostamm: Adding it in https://github.com/connectrpc/connect-es/pull/774

paul-sachs:
@timostamm I've fixed the issue with parsing the issue you had with the `custom-transport.ts` file and we now use different parsers for different filetypes. This does raise the risk that someone might have some custom babel setup but i've added a section to the readme to describe other ways to run our transforms.

timostamm (approved):

PR #782, opened 2023-08-23
Add version checks to migration CLI

paul-sachs:
Add the ability to force update packages when the installed versions may not satisfy requirements.

Updated with https://github.com/connectrpc/connect-es/pull/787 to migrate across breaking changes by default. 

For every future release up to version 1.0.0, we should update the constant `targetVersionConnectEs`, so that a migration from any version <1.x results in @connectrpc dependencies constrained to ^1.0.0.

Past v1, it is not necessary to update the constant. For every new major version, we can consider adding a new migration to the command.


packages/connect-migrate/src/migrate.ts
```diff
@@ -19,50 +19,179 @@ import { run } from "./run";
...
+    if (!forceUpdate) {
+      result = result.replace(
+        `"${key}"`,
+        `"${replacementMap[key].newPackage}"`,
+      );
+    } else {
+      // I wish there was a way to do this without relying on dynamic regexes
+      // but not while maintaining existing formatting
+      result = result.replace(
+        new RegExp(`"${key}"(\\s*:\\s*)"\\S*"([\\s,])`, "g"),
```
paul-sachs, line 113: If we're willing to rewrite the entire package.json file, we can simplify here, but I'm not sure we want to do that.

packages/connect-migrate/src/migrate.ts
```diff
@@ -19,50 +19,179 @@ import { run } from "./run";
...
-  "@bufbuild/connect-express": "@connectrpc/connect-express",
-  "@bufbuild/protoc-gen-connect-es": "@connectrpc/protoc-gen-connect-es",
-  "@bufbuild/protoc-gen-connect-web": "@connectrpc/protoc-gen-connect-web",
-  "@bufbuild/connect-query": "@connectrpc/connect-query",
-  "@bufbuild/protoc-gen-connect-query": "@connectrpc/protoc-gen-connect-query",
-  "@bufbuild/protoc-gen-connect-query-react":
-    "@connectrpc/protoc-gen-connect-query-react",
+  "@bufbuild/connect": {
+    newPackage: "@connectrpc/connect",
+    version: "0.13.1",
```
timostamm, line 29: We're setting the exact constraint`0.13.1`, which means that `npm update` won't update a minor version, and the next connect-migrate command is guaranteed to need --force-update.

By default, package managers use `^`. Can we do that too?

packages/connect-migrate/src/migrate.ts
timostamm: Pushed up https://github.com/connectrpc/connect-es/pull/782/commits/315540a88b560d75af9ba4c722dc75d0b042b9bc to go through JSON, formatting with two spaces for indentation. We run `npm install` too, so I think that's pretty reasonable.

timostamm (approved):

PR #783, opened 2023-08-24
Add defensive check for http2 sessions

### LINKED ISSUE #683: [internal] The session has been destroyed
version: 0.10.0

```
ConnectError: [internal] The session has been destroyed
	at i.from (/out/extension.js:2:1820120)
	at t.connectErrorFromNodeReason (/out/extension.js:2:1804959)
	at Promise.reject (/out/extension.js:2:1809452)
	at /out/extension.js:2:1811572
```

Unfortunately stack trace is minified on our side not sure which path lead to that, it seem to be an unary call. It seems somewhere stream is called when it is already destroyed.

srikrsna-buf:
Add defensive check for http2 sessions. This should fix #683. The solution is similar to the one by @mustard-mh in #712 it checks if the session is destroyed/closed and moves to closed state and retries the request with a new connection.

### DIFF
--- a/packages/connect-node/src/http2-session-manager.ts
+++ b/packages/connect-node/src/http2-session-manager.ts
@@ -202,17 +202,36 @@ export class Http2SessionManager {
     headers: http2.OutgoingHttpHeaders,
     options: Omit<http2.ClientSessionRequestOptions, "signal">,
   ): Promise<http2.ClientHttp2Stream> {
-    const ready = await this.gotoReady();
-    const stream = ready.conn.request(
-      {
-        ...headers,
-        ":method": method,
-        ":path": path,
-      },
-      options,
-    );
-    ready.registerRequest(stream);
-    return stream;
+    // Request sometimes fails with goaway/destroyed
+    // errors, we use a loop to retry.
+    //
+    // This is not expected to happen often, but it is possible that a
+    // connection is closed while we are trying to open a stream.
+    //
+    // Ref: https://github.com/nodejs/help/issues/2105
+    for (;;) {
+      const ready = await this.gotoReady();
+      try {
+        const stream = ready.conn.request(
+          {
+            ...headers,
+            ":method": method,
+            ":path": path,
+          },
+          options,
+        );
+        ready.registerRequest(stream);
+        return stream;
+      } catch (e) {
+        // Check to see if the connection is closed or destroyed
+        // and if so, move to closed state and try again.
+        if (ready.conn.closed || ready.conn.destroyed) {
+          this.setState(closed());
+          continue;
+        }
+        throw e;
+      }
+    }
   }
 
   /**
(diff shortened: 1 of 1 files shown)


packages/connect-node/src/http2-session-manager.ts
```diff
@@ -202,17 +202,36 @@ export class Http2SessionManager {
...
+        );
+        ready.registerRequest(stream);
+        return stream;
+      } catch (e) {
+        // Check to see if the connection is closed or destroyed
+        // and if so, move to closed state and try again.
+        if (ready.conn.closed || ready.conn.destroyed) {
+          this.setState(closed());
+          continue;
+        }
```
timostamm, line 231: Tearing down the broken "ready" state is important, but I think we're running into a race condition this way.

Here's a reproduction:

<details>
<summary>r.mjs</summary>

```js
import {Http2SessionManager} from "@connectrpc/connect-node";

/* eslint-disable */

const sm = new Http2SessionManager("https://demo.connect.build");
await sm.connect();

// @ts-expect-error
sm.s.conn = new Proxy(sm.s.conn, {
    get(target, prop, receiver) {
        if (prop === "destroyed") {
            return true;
        }
        return Reflect.get(...arguments);
    }
})

const r1 = sm.request("GET", "/r1", {}, {});
r1.then(() => {
    console.log("r1 resolved")
}, reason => {
    console.error("r1 rejected", reason)
});

const r2 = sm.request("GET", "/r2", {}, {});
r2.then(() => {
    console.log("r2 resolved")
}, reason => {
    console.error("r2 rejected", reason)
});

await Promise.all([r1, r2]);
console.log("all request resolved")
```
</details>

```bash
$ node r.mjs 
r2 resolved
```

The first request sets the state from "ready" to "closed" to "connecting" and the promise is suspended in gotoReady() waiting for the connection promise to settle.

But the second request is already running (it started when the first request suspended in await gotoReady). It also goes from "ready" to "closed" to "connecting", opening a second connection. The promise the first request is waiting on never settles.

I hadn't fully considered this situation either. I believe that we have to keep the logic in one place to handle concurrency properly.

timostamm: Maybe the Proxy allows us to get a better grip on this condition in tests? 

The try-catch is certainly nice, but I have a hunch that this might work out better by handling `conn.destroyed || conn.closed` in gotoReady.

timostamm (approved):
LGTM!

htranugaming:
@srikrsna-buf can you pump a version for this commit?, I still got this issue on my production site, thanks you

timostamm:
@htranugaming, this just shipped in [v0.13.2](https://github.com/connectrpc/connect-es/releases/tag/v0.13.2). Let us know if this fixes the problem.

Issue #527, opened 2023-03-13
Add interceptor capabilities to ConnectRouter


fubhy:
**Is your feature request related to a problem? Please describe.**
It's currently not easily possible to implement generic middlewares for service methods. This makes it rather tedious to enhance connect services with generic logic across multiple methods.

**Describe the solution you'd like**
I'd like to be able to implement interceptors / middlewares for my service methods or groups of service methods (Connect for Node.js). The recently introduced ConnectRouter appears to be a good candidate to register these.

These interceptors should have the ability to augment the `HandlerContext`. Ideally, this should be done with TypeScript in mind ... That said, I'm not a big fan of blanket type augmentation as used in e.g. Express, etc. for this problem. Maybe there's a better way.

**Describe alternatives you've considered**
We are currently "intercepting" requests inconveniently by wrapping all service methods with higher order functions. To keep things simple and because that's the extent of our requirements currently, we've only implemented this pattern for unary requests. So this is only a temporary workaround at the moment.

**Additional context**
Implementing interceptors would allow the development of generic, community provided interceptor modules. E.g. for integrations with OpenTelemetry (metrics & tracing): https://github.com/bufbuild/connect-es/issues/523.

Other possible use cases: 
- Automatic request (& response) validation (protoc-gen-validate)
- AuthN & AuthZ
- Rate limiting
- Caching

These could all be nicely paired with proto options based code generation.

The ConnectRouter is currently only handling POST requests. That means that, in its current form, it wouldn't be the right place to also handle things like OPTIONS requests (CORS). I don't know whether (and how) "fixing" that is within the scope of this issue. However, I can definitely see value in being able to handle CORS at service method granularity... So maybe there's room for that here somehow?

timostamm:
Related to augmenting the `HandlerContext`, the following feature requests should be considered in the design:
- https://github.com/connectrpc/connect-es/issues/708
- https://github.com/connectrpc/connect-es/issues/586
- https://github.com/connectrpc/connect-es/issues/550#issuecomment-1517526847

srikrsna-buf:
You can now try out this feature using the `next` release of connect: `npm i @connectrpc/connect@next`. The docs for the next release are hosted [here](https://connect-git-sk-esinterceptors-connectrpc.vercel.app/docs/node/interceptors)

Issue #586, opened 2023-04-17
Provide express' req/res.locals inside HandlerContext


kruegernet:
**Is your feature request related to a problem? Please describe.**
When migrating an existing express/REST service to use connect/gRPC, it will be a typical scenario that an RPC handler must work with the existing express middleware. If we cannot access req.locals or res.locals, for some users adopting connect-es will be impossible.

**Describe the solution you'd like**
Please provide access to req/res.locals via the HandlerContext. 

This request is for Connect for Node.js.

**Describe alternatives you've considered**
I would personally prefer, at least during beta, that the entire raw express req and res are exposed to us via HandlerContext. It's not clear to me how you all plan to address authentication/authorization or middleware generally, or how you intend to relate http middleware to gRPC interceptors. Until these are resolved it remains unclear how folks can migrate existing services or do serious work with connect-es.

mfecteau42:
A workaround was found by modifying this file : **@bufbuild/connect/dist/esm/protocol-connect/handler-factory.js**

By adding this code at the right place, in both functions **createUnaryHandler** and **createStreamHandler**, the res.locals variable becomes available : 
```js
if (req.locals) {
    context.locals = req.locals
}
```

srikrsna-buf:
#835

PR #811, opened 2023-09-02
Add support for server side interceptors

aschleck:
(Hopefully) fixes https://github.com/connectrpc/connect-es/issues/527

* I spent a bunch of time trying to minimize the couple of casts, I think ultimately since MethodImpl is polymorphic it's inevitable to need this type of casting.
* I don't love that interceptors are a property of the `ConnectRouterOptions`, originally I was going to put them on the `ConnectRouter`. However they cannot be on `ConnectRouter` because it immediately sets up registered methods and interceptors have to be bound inside `createMethodImplSpec` (to be explicit, if you register a method and then add an interceptor afterwards the method wouldn't be intercepted.) It would be a larger and relatively pointless refactor to make it all lazy. Putting it on `ConnectRouterOptions` just means that all interceptors will be set prior to any methods being registered.

Thank you for open sourcing this fantastic set of libraries!

### DIFF
--- a/packages/connect/src/implementation.spec.ts
+++ b/packages/connect/src/implementation.spec.ts
@@ -12,12 +12,39 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { createHandlerContext } from "./implementation.js";
+import {
+  createHandlerContext,
+  createMethodImplSpec,
+} from "./implementation.js";
+import type {
+  BiDiStreamingImpl,
+  ClientStreamingImpl,
+  ServerStreamingImpl,
+  UnaryImpl,
+} from "./method-implementation.js";
 import { Int32Value, MethodKind, StringValue } from "@bufbuild/protobuf";
 
 const TestService = {
   typeName: "handwritten.TestService",
   methods: {
+    biDiStreaming: {
+      name: "BiDiStreaming",
+      I: Int32Value,
+      O: StringValue,
+      kind: MethodKind.BiDiStreaming,
+    },
+    clientStreaming: {
+      name: "ClientStreaming",
+      I: Int32Value,
+      O: StringValue,
+      kind: MethodKind.ClientStreaming,
+    },
+    serverStreaming: {
+      name: "ServerStreaming",
+      I: Int32Value,
+      O: StringValue,
+      kind: MethodKind.ServerStreaming,
+    },
     unary: {
       name: "Unary",
       I: Int32Value,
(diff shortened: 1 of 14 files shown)


timostamm:
Whoa, thank you @aschleck :star_struck:

Right now, we're on a freeze, just about to cut a stable version 1 of connect-es. While we cannot get this merged right away, we definitely appreciate the contribution. We'll want to land this in a feature release post v1, ideally in combination for a solution for https://github.com/connectrpc/connect-es/issues/586 (context information is closely related to interceptors), which could be added as a follow-up.

aschleck:
Thank you for the quick reply @timostamm! It's a bummer to hear I missed the window but totally understand. Provided you all like this code, do you have a sense of what timeline this might be able to get released on? I'm pushing my company to totally adopt this framework because it's obviously the correct choice, but interceptors are a big part of making it nice :)

srikrsna-buf:
Hey @aschleck! Apologies for the delay and thank you for this.

I am not sure whether you looked at [client interceptors](https://connectrpc.com/docs/web/interceptors), but one nice property they have is that the signature is common for all method kinds. This is a useful property because interceptors usually are used to add generic functionality like (authentication, logging etc...) and having a single signature makes it easier to implement such interceptors. For the use cases where they only target a particular method kind they can always check for that and skip the rest.

Could you modify the design to something that has a common signature? Before you rewrite this (as it is a lot of work) it'd be great if you could write down the design and we can take it from there. Thank you!

aschleck:
Hi @srikrsna-buf, thank you for taking a look at this. I don't really have time for what you're asking and it seems like a matter of taste (trading off safety for convenience) so I'm going to have to decline, but appreciate this library and wish you folks the best of luck. Please feel free to close this.

If I were to do something generic, what I would personally lean towards is making a `createFilterInterceptor(fn: (context: HandlerContext) => boolean)` function that returns the expanded interceptor. Then you can leave this design as-is, and ensures you don't do crazy things like return unary responses for streaming RPCs due to a lack of type safety, but also lets you make the common case easy.

aschleck:
Actually now that I think about it, you can get what you want (ease of use) with a helper method `(next: AnyFn, context: HandlerContext) => AnyFn` that wraps what I've done if you start with my approach. But you can't get what I want (type safety in the core of connect-es) if you start with your approach.

srikrsna-buf:
> Hi @srikrsna-buf, thank you for taking a look at this. I don't really have time for what you're asking

Thank you for taking the time to respond, I understand.

> it seems like a matter of taste (trading off safety for convenience)

Just to be clear it will not be any less type-safe, the client interceptors implementation has to check for the `stream` property of request/response to access the body. Other fields like headers are always accessible. 

There is a possibility of returning the wrong response but at that point, you are explicitly setting the stream property and a bunch of others and not relying on the response returned from the handler, which is very unusual and almost all such cases of short-circuiting the request will do so by throwing an error rather than returning a custom response.

I'll close this, I'll be working on the interceptors next and I'll send a PR once #835 lands. I'm hopeful you'll like the design :)

PR #812, opened 2023-09-05
Rename crosstest references

smaye81:
This renames all the crosstest references and usages to use the new `connectrpc/conformance` repo.

Note this is not intended to merge (or build properly) until the crosstest repo is moved.


timostamm:
Happy to review, but that's not really possible yet. The proto namespace was changed, btw. so this will need more refactoring, not just an update to the reference. It might actually be the better approach to split it into two PRs.

smaye81:
Yeah. I suppose I would've run into that eventually but thanks for the heads up. Here's the other PR:

https://github.com/connectrpc/connect-es/pull/812

smaye81:
@timostamm should be good to review now.

timostamm (approved):
LGTM, since this refactoring only touches test code.

PR #817, opened 2023-09-08
Make`@connectrpc/connect` a peer dependency and pin the version

srikrsna-buf:
Using both `@connectrpc/connect-node` and `@connectrpc/connect-web` can result in two different versions of `@connectrpc/connect`. To avoid this we make `@connectrpc/connect` a peer dependency and pin to an exact version. This way users will be forced to use the same version for all packages.

### DIFF
--- a/packages/connect-fastify/package.json
+++ b/packages/connect-fastify/package.json
@@ -25,13 +25,11 @@
   "engines": {
     "node": ">=16.0.0"
   },
-  "dependencies": {
-    "@connectrpc/connect": "0.13.2",
-    "@connectrpc/connect-node": "^0.13.2",
-    "fastify": "^4.22.1"
-  },
   "peerDependencies": {
-    "@bufbuild/protobuf": "^1.2.1"
+    "@bufbuild/protobuf": "^1.2.1",
+    "fastify": "^4.22.1",
+    "@connectrpc/connect": "0.13.2",
+    "@connectrpc/connect-node": "0.13.2"
   },
   "files": [
     "dist/**"
--- a/package-lock.json
+++ b/package-lock.json
@@ -8100,32 +8100,30 @@
       "name": "@connectrpc/connect-express",
       "version": "0.13.2",
       "license": "Apache-2.0",
-      "dependencies": {
-        "@connectrpc/connect": "0.13.2",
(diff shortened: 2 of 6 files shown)


packages/connect-fastify/package.json
```diff
@@ -25,13 +25,11 @@
...
-    "@connectrpc/connect": "0.13.2",
-    "@connectrpc/connect-node": "^0.13.2",
-    "fastify": "^4.22.1"
-  },
   "peerDependencies": {
-    "@bufbuild/protobuf": "^1.2.1"
+    "@bufbuild/protobuf": "^1.2.1",
+    "fastify": "^4.22.1",
+    "@connectrpc/connect": "0.13.2",
+    "@connectrpc/connect-node": "0.13.2"
```
timostamm, line 32: If you write a server with Fastify, you would previously install @connectrpc/connect and @connectrpc/connect-fastify.

With this change, you also need to install @connectrpc/connect-node, whether you use clients or not.

I see two options here:

1. Keep @connectrpc/connect-node as a direct dependency, but with an exact version. If the user later decides to use clients and adds @connectrpc/connect-node, the exact peer dependency requirements will force them all to be on the same version. 
2. Use @connectrpc/connect-node as a peer dependency, but make it optional with peerDependenciesMeta.

I am not sure if hoisting in npm is guaranteed to deduplicate with option 1, and tend to think that option 2 is the better choice.

WDYT?

timostamm: My bad, you're right about option 2.

srikrsna-buf: I am doubtful about deduplication, it'd be safer to just make it a peer.

timostamm: Yes, a peer is safer - but at the cost of DX. At the very minimum, we have to update our documentation [here](https://connectrpc.com/docs/node/getting-started#start-a-server) - otherwise our tutorial leads to an error.

srikrsna-buf: If the cost of DX is okay, I'll send a PR to update the docs before the next release. We can merge that as soon as we release.

timostamm (approved):

PR #830, opened 2023-09-18
Adding msw integration

### LINKED ISSUE #825: Support API mocking via MSW
When writing a web application that makes many API calls, it is often challenging to get good code coverage without mocking the API.

The [Mock Service Worker library](https://mswjs.io/) can intercept requests on the network level and mock responses in a framework-agnostic way, and it runs on Node.js too.

MSW can already be used with Connect, but it requires mocking on the HTTP level, without any benefits that the schema provides: For example, it is easy to misspell a header name or request path - even though both are already well-defined by the schema or protocol.

It would be fantastic if there was a type-safe integration for MSW with Connect to remove the boilerplate, for example:

```ts
import { setupWorker } from 'msw'
import { service } from '@connectrpc/connect-msw'

const worker = setupWorker(
  service(ElizaService, {
    say(req) {
      return { sentence: `you said ${req.sentence}` }
    }
  }),
)

worker.start()
```

[@connectrpc/connect-playwright](https://www.npmjs.com/package/@connectrpc/connect-playwright) provides API mocking with [Playwright](https://playwright.dev/), but that means all tests must be written in Playwright, limiting the options.

**Additional context**
- Prior effort for gRPC-web: [mswjs/msw#238](https://github.com/mswjs/msw/issues/238) and a [blog post](https://lucas-levin.com/code/blog/mocking-grpc-web-requests-for-integration-testing).
- Ongoing effort to use fetch API primities in MSW: [mswjs/msw#1436](https://github.com/mswjs/msw/pull/1436)

paul-sachs:
This is a PoC of what a msw integration could look like.

Fixes #825 

## Caveats

- Built ontop of an experimental build of MSW that's built ontop of fetch primities: https://github.com/mswjs/msw/pull/1436

### DIFF
--- a/packages/connect-msw/src/create-worker-handlers.ts
+++ b/packages/connect-msw/src/create-worker-handlers.ts
@@ -1,11 +1,7 @@
 import { MethodIdempotency } from "@bufbuild/protobuf";
-import type {
-  CallOptions,
-  ConnectRouter,
-  HandlerContext,
-} from "@connectrpc/connect";
+import type { ConnectRouter, HandlerContext } from "@connectrpc/connect";
 import type { UniversalHandler } from "@connectrpc/connect/protocol";
-import { http, HttpResponse, passthrough as mswPassthrough } from "msw";
+import { http, HttpResponse, passthrough as mswPassthrough, bypass } from "msw";
 import type { RequestHandler, ResponseResolver } from "msw";
 
 const passthroughHeader = "x-passthrough";
@@ -15,20 +11,13 @@ export function passthrough(ctx: HandlerContext) {
   return {};
 }
 
-export function createBypassOptions(callOptions: CallOptions): CallOptions {
-  return {
-    ...callOptions,
-    headers: {
-      // We could make this more consistent with the exposed bypass() function provided by msw
-      // IFF we could specify interceptors on a request by request basis. At least in theory, bypass()
-      // works on a standard Request object so we'd need a converter from the interceptor UnaryRequest/StreamRequest
-      // to a standard request to pass it to bypass() and out again.
-      // Otherwise, we just need to make sure this keeps in sync with https://github.com/mswjs/msw/blob/8855956f561ad4afd4ce7ae6500ca706757bb5a9/src/core/bypass.ts
-      "x-msw-intention": "bypass",
-      ...callOptions.headers,
-    },
-  };
-}
+/**
+ * Creates a custom fetch that will bypass MSW. This can be passed to
+ * as custom transport in order to to explicitly bypass MSW.
+ */
+export const bypassFetch: typeof globalThis.fetch = (input, init) => {
+  return fetch(bypass(input, init));
+};
 
 function createResponseResolver(handler: UniversalHandler): ResponseResolver {
   return async ({ request }) => {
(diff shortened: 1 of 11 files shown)


paul-sachs:
Note: These PR is mostly for discussion and will probably end up in it's own repo eventually.

packages/connect-msw/src/create-worker-handlers.ts
```diff
@@ -0,0 +1,123 @@
...
+export function createBypassOptions(callOptions: CallOptions): CallOptions {
+  return {
+    ...callOptions,
+    headers: {
+      // We could make this more consistent with the exposed bypass() function provided by msw
+      // IFF we could specify interceptors on a request by request basis. At least in theory, bypass()
+      // works on a standard Request object so we'd need a converter from the interceptor UnaryRequest/StreamRequest
+      // to a standard request to pass it to bypass() and out again.
+      // Otherwise, we just need to make sure this keeps in sync with https://github.com/mswjs/msw/blob/8855956f561ad4afd4ce7ae6500ca706757bb5a9/src/core/bypass.ts
+      "x-msw-intention": "bypass",
```
paul-sachs, line 27: @timostamm wanted to get your thoughts on this way of adding headers to a bunch of requests. Basically this header exists to tell msw to have a request bypass all mock handlers. This can be applied to any client request (as opposed to passthrough, which is designed to just the incoming request). It would be nice to have a way to intercept the Request before it gets sent to fetch so we can more seamlessly fit with the `bypass()` function.

Not really blocking, just wanted to raise it.

timostamm: What is bypass for? I see this comment in the source:

>  Unlike "passthrough()", bypass is meant for performing additional requests within pending request resolution.

If I intercept a request, and need to make some requests to provide a mocked response, do I need to set this header to prevent MSW from intercepting these requests as well?

That's pretty straight-forward with:

```ts
client.say({sentence: "hi"}, { headers: { "x-msw-intention": "bypass" } });
```

Users could also set up a transport specifically for this use case, with a connect interceptor that always sets this header. 

Seems to me that this just needs good examples in our documentation, but please let me know if I'm missing something. 

MSW might be able to provide slightly convenience by providing a `fetch` function to response resolvers that always sets this header, and potentially inherits headers and cookies - similar to Svelte.

paul-sachs: > If I intercept a request, and need to make some requests to provide a mocked response, do I need to set this header to prevent MSW from intercepting these requests as well?

Yes that is correct. 

> Seems to me that this just needs good examples in our documentation, but please let me know if I'm missing something.

Yeah maybe docs are sufficient. I wanted to raise it to see if there was any feature requests to allow interceptors on a per request basis. I've pushed up a custom fetch that can be passed. It's a little cleaner than a custom interceptor and can still be wrapped if someone needs an additional custom fetch.

timostamm: >  I wanted to raise it to see if there was any feature requests to allow interceptors on a per request basis.

Do you mean connect interceptors as described [here](https://connectrpc.com/docs/web/interceptors)? I think it would be a bad design choice to allow the per request because it would soften the separation of concerns, and opens up questions about order of interceptors. If you want to run interceptors conditionally, https://github.com/connectrpc/connect-es/pull/841 is the way to go.

paul-sachs: Ah yeah the lack of clarity on ordering makes sense. An easy way to work around that would be to instead allow a custom fetch to be defined on a per request basis, which wouldn't introduce any ordering confusion. But for now, a custom transport to handle bypasses will suffice. I don't think it's worth much api modification just to smooth out things like msw bypass.

tunguyen-ct:
"Great news! [MSW v2.0.0](https://github.com/mswjs/msw/releases/tag/v2.0.0) has been released. Do we have an estimated timeline for shipping the connect-msw package? 😄"

PR #834, opened 2023-09-19
Add retries to node curl commands

smaye81:
This adds retries to the `curl` commands used to pull the Node binaries. It also moves the Browserstack tests to their own workflow. Since these tests always fail for contributor PRs due to credentials, it will be easier to isolate any failures that occur.

### DIFF
--- a/.github/workflows/browserstack.yaml
+++ b/.github/workflows/browserstack.yaml
@@ -9,23 +9,20 @@ on:
 permissions:
   contents: read
 jobs:
-  ci:
+  browserstack:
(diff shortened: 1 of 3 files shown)


timostamm (commented):
Since we're already moving browserstack to a separate workflow here, why not add the bade as in https://github.com/connectrpc/connect-es/pull/834/commits/a29e69536483006d9384e6299352f1bf049e16b5? I only suggested to omit the badge because browserstack has been a bit flaky for us in the past.

timostamm (approved):

Issue #818, opened 2023-09-11
Difficulty Accessing Authenticated User Object from Passport Middleware


akosyakov:
I'm using Passport's authentication middleware, which provides an authenticated user object on a request ([Passport Middleware Documentation](https://www.passportjs.org/concepts/authentication/middleware/)). I'm struggling to access this user object in services.

A potential workaround is to do a double lookup in the service based on the cookie, but I'd prefer not to. It seems counterintuitive for a library to obscure objects, especially when there's an ecosystem contributing to them. This makes integration challenging.

akosyakov:
Figured out that I can use ts augmentation to add more properties and mutate context via proxies.

srikrsna-buf:
#835

Issue #708, opened 2023-07-02
Cannot get fastify context in request handler.


acrazing:
In original fastify handler, we can access the fastify request context handy.

```js
fastify.get('/', options, function (request, reply) {
  request.log.info('Some info about the current request')
  reply.send({ hello: 'world' })
})
```

Bug it is impossible in connect-es.

```js
router.service(Health, {
  check(request, context) {
    // no any data passed here from fastify
    console.log(context)
  }
}
```

smaye81:
Hi @acrazing can you elaborate on what details of the Fastify request you're looking for? Is it the logging instance or are there other bits of info you need?

acrazing:
logger is the most urgently needed.

acrazing:
But other information is also very important, such as ip address?

I think a mechanism is needed to allow users to inject context, similar to grpc's interceptor.

acrazing:
Here is my current patch file:

```diff
diff --git a/node_modules/@bufbuild/connect-fastify/dist/cjs/fastify-connect-plugin.js b/node_modules/@bufbuild/connect-fastify/dist/cjs/fastify-connect-plugin.js
index 505235f..1983a89 100644
--- a/node_modules/@bufbuild/connect-fastify/dist/cjs/fastify-connect-plugin.js
+++ b/node_modules/@bufbuild/connect-fastify/dist/cjs/fastify-connect-plugin.js
@@ -43,7 +43,7 @@ function fastifyConnectPlugin(instance, opts, done) {
     for (const uHandler of uHandlers) {
         instance.all(uHandler.requestPath, {}, async function handleFastifyRequest(req, reply) {
             try {
-                const uRes = await uHandler((0, connect_node_1.universalRequestFromNodeRequest)(req.raw, req.body));
+                const uRes = await uHandler(Object.assign((0, connect_node_1.universalRequestFromNodeRequest)(req.raw, req.body), { extraContext: { req, reply, log: req.log } }));
                 // Fastify maintains response headers on the reply object and only moves them to
                 // the raw response during reply.send, but we are not using reply.send with this plugin.
                 // So we need to manually copy them to the raw response before handing off to vanilla Node.
diff --git a/node_modules/@bufbuild/connect-fastify/dist/types/fastify-connect-plugin.d.ts b/node_modules/@bufbuild/connect-fastify/dist/types/fastify-connect-plugin.d.ts
index 675ac8a..fdb473a 100644
[...]

Issue #550, opened 2023-03-29
Serve handler from cloudflare workers


robbertvanginkel:
**Is your feature request related to a problem? Please describe.**
I want to use connect-es to serve a connect/grpc-web handler from a cloudflare worker, filing an issue as suggested in https://github.com/bufbuild/connect-es#other-platforms.

**Describe the solution you'd like**
Similar to the adaptor that exists for node, something that I can plug into the worker's https://developers.cloudflare.com/workers/runtime-apis/fetch-event/. It would only need to support the connect & grpc-web protocols, as it seems that the workers runtime does not expose enough to implement the gRPC protocol.

**Describe alternatives you've considered**
Not using connect-es and manually writing the grpc-web/connect protocol output.

**Additional context**
I've gotten some basics to work with the following snippet:

```typescript
import { createConnectRouter } from "@bufbuild/connect";
import { UniversalHandler } from "@bufbuild/connect/protocol";
import { ElizaService } from "./gen/buf/connect/demo/eliza/v1/eliza_connect";

export default {
	async fetch(request: Request, env: Env, ctx: ExecutionContext ): Promise<Response> {
		const url = new URL(request.url);
		let router = createConnectRouter({ grpc: false, grpcWeb: true, connect: true})
		router.service(ElizaService, {
			async say(req) {
				return {sentence: `You said ${req.sentence}`};
			}
		})
		const paths = new Map<string, UniversalHandler>();
		for (const uHandler of router.handlers) {
			paths.set(uHandler.requestPath, uHandler);
		}
		const uHandler = paths.get(url.pathname ?? "");
		if (!uHandler) {
			return new Response(undefined, { status: 404 });
		}

		let response = await uHandler({
			url: url,
			header: request.headers,
			httpVersion: "2",
			method: request.method,
			body: request.body,
		})
		return new Response(
			readableStreamFromIterable(wrapIterable(response.body ?? new Uint8Array())),
			{ status: response.status, headers: response.header }
		)
	},
};

function wrapIterable(x: AsyncIterable<Uint8Array> | Uint8Array): AsyncIterable<Uint8Array> {
	if (x instanceof Uint8Array) {
		return { async *[Symbol.asyncIterator]() { yield x } }
	}
	return x;
}

function readableStreamFromIterable<T>(iterable: Iterable<T> | AsyncIterable<T>): ReadableStream<T> {
[...]

PR #835, opened 2023-09-20
Add support for user provided context in handlers

srikrsna-buf:
Add support for user provided context in handlers. With this change handlers can depend on user provided values in the context. Related issues: #818, #586, #550 and #708

Example:
Create a context key with a default value:
```ts user-ctx.ts
export interface User {
  id: string;
}
import { createContextKey } from "@connectrpc/connect";
export const kUser = createContextKey<User | undefined>(
  undefined // The default value.
);
```
Use the `contextValues` option to provide the context values for each request:
```ts index.ts
import { fastify } from "fastify";
import routes from "./connect";
import { fastifyConnectPlugin } from "@connectrpc/connect-fastify";
import { authenticate } from "./authenticate.js"; 
import { kUser } from "./user-ctx.js";

const server = fastify();

await server.register(fastifyConnectPlugin, {
 routes,
 contextValues: (req) => createContextValues().set(kUser, authenticate(req)),
});

await server.listen({
  host: "localhost",
  port: 8080,
});
```
Use the context value in the handler:
```ts routes.ts
import { ConnectRouter } from "@connectrpc/connect";
import { ElizaService } from "./gen/eliza_connect";

export default (router: ConnectRouter) => {
  // using const say
  router.service(ElizaService, {
    say: (req, { values }) => {
      const currentUser = values.get(kUser);
      if (currentUser === undefined) {
        throw new ConnectError("Unauthenticated", Code.Unauthenticated);
      }
      // ...
    },
  });
};

```


paul-sachs:
I love the way this keeps the type information associated to the context id, very cool. One question I have is if we want to expose the context values to interceptors as well?

srikrsna-buf:
> I love the way this keeps the type information associated to the context id, very cool. One question I have is if we want to expose the context values to interceptors as well?

We don't have server side interceptors yet, but if we finalize this API I will send a follow up to add this to the clients. The server side interceptors should include them when implemented.

smaye81:
This may be a contrived example, but what if the user wants to add something to context that doesn't involve the request object? I suppose they could do:

```ts
await server.register(fastifyConnectPlugin, {
 routes,
 contextValues: () => createContextValues().set(kUser, {"name": "Krishna"}),
});
```

Does it make sense then to make the `req` optional in the `contextValues` function property?

srikrsna-buf:
@smaye81 Your example will work without any changes, `req` is something we pass to the function and can be omitted if they don't need it.

timostamm (commented):
I imagine one might want to console.log context values or inspect them in a debugger. 

Should we improve behavior with an optional description string for the symbol, and by storing values as properties instead of a captured variable?

jeffsu:
Would it be helpful to consider a solution that introduces typesafe constraints between the route handlers and the context builder?

Here is a rough example of the types I'm talking about:

```typescript
// context now has typesafety
type Context<V={}> = { values: V };

// example of how a route handler can assert that it needs some data out of context
type RouterHandler<I, O, V={}> =  (input: I, context?: Context<V>) => O;

// then we can force the injection of some data in context
class Server<C extends Context> {
  addContextValue<K extends LiteralString, T>(lambda: (req: Request) => T): Server<Context<C["values"] & { K: T}>>;
}
```

srikrsna-buf:
Thank you for the suggestion @jeffsu! 

Could you give us a simple e2e example of how this will work? I am unable to see how this can guarantee the presence of a value. In the current design, users have to pass a default value.

One other thing to note is that we want to add context values to interceptors on both the client and server side (not implemented yet). Values can be added via the interceptors as well which means the values are not always available at the request start.

srikrsna-buf:
Hey @jeffsu to illustrate what I meant by adding them to interceptors I just opened #841 please take a look and let me know what you think.

One thing I noticed with the current implementation is that `ContextValues` is mutable. Should we make it immutable?

Current behavior:
```ts
const interceptor = (next) => {
     return (req) => {
          const user = req.values(kUser);
          const res = next(req); // This can modify req.values
          // req.values(kUser) !== user
     }
}
```

cc @timostamm @paul-sachs @smaye81

paul-sachs:
This level of mutability doesn't really bother me. If we made a version of context that could be deep cloned, you could do something like:

```tsx
const interceptor = (next) => {
     return (req) => {
          const user = req.values(kUser);
          const res = next({ ...req, values: req.values.clone() }); // This can modify req.values
          // req.values(kUser) === user
     }
}
```

timostamm (approved):
It's true that handlers cannot declare a dependency on a context value with this, as Jeff noted. I think this is a nice step forwards though. I say LGTM.

mblode:
Is there an example of how to use this new system to get the logger inside the connect service?

 
E.g
```
fastify.get('/', options, function (request, reply) {
  request.log.info('Some info about the current request')
  reply.send({ hello: 'world' })
})

router.service(Health, {
  check(request, context) {
    // log here
    console.log(context)
  }
}

```

srikrsna-buf:
> Is there an example of how to use this new system to get the logger inside the connect service?

It is the same as the example in the PR description, instead of 
```contextValues: (req) => createContextValues().set(kUser, authenticate(req)),```
you can just use a different key for the log type:
``` contextValues: (req) => createContextValues().set(kLog, req.log),```

And then can get the value and use it:
```
router.service(Health, {
  check(request, context) {
    context.values.get(kLog).log.info(...);
    console.log(context)
  }
}


```

**Comment by StarpTech** [comment]
Why was the implementation shipped this way? We have no async support and we can't pass a global type for contextValues. Our use case is to check authentication in this handler. Do you accept a PR?

**Comment by srikrsna-buf** [comment]
@StarpTech Typically if you're using something like Fastify/Express you will have an auth library that does the authentication checks and adds the subject info to the incoming request, and you can add the that to the context.

> we can't pass a global type for contextValues.

This is actually error prone and could have conflicts. Using a unique key lookup helps us guarantee type safety and always provide a default value.

**Comment by StarpTech** [comment]
I see your points but why should we limit the implementation to only synchronous work? I don't see the value in it.

> This is actually error prone and could have conflicts. Using a unique key lookup helps us guarantee type safety and always provide a default value.

I'm quite on the opposite side. We have to specify a shared ContextKey Type and use the right type everywhere we access it instead of maintaining the relation once

This would be more conenvient

```ts
(router: ConnectRouter) => {
    router.service<MyContextValuesType>(MyService, MyServiceServiceImpl(opts), handlerOptions);
  };
```

Access

```ts
 ctx.values.get(Symbol('foo)) // return the correct type
```

Another thing we find odd is, why is it necessary to specify the the default value when retrieving a value?

```ts
// Error: Property defaultValue is missing in type { id: symbol; } but required in type ContextKey<unknown>
ctx.values.get({
  id: Symbol('subgraph'),
});
```

srikrsna-buf:
> Access
> 
> ```ts
>  ctx.values.get(Symbol('foo')) // return the correct type, typesafe access out of the box
> ```

This is how it can be used today, instead of creating a `Symbol` the key is nothing but a type-safe proxy to a static symbol.

> Another thing we find odd is, why is it necessary to specify the the default value when retrieving a value? If the value can be undefined, it is the dev responsibility to deal with it.

Default is not needed when retrieving a value, it is only needed when creating the key. If `undefined` is needed, then the type can be set to do that: `T | undefined`. 

We just updated the docs with an [example](https://connectrpc.com/docs/node/interceptors#example). Hopefully that helps clarify the usage a little bit. This is fairly a common pattern in the JS ecosystem, a popular mechanic that works like this is the React Context.

StarpTech:
> This is how it can be used today, instead of creating a Symbol the key is nothing but a type-safe proxy to a static symbol.

Was the API updated? This is not possible in `1.1.3`. 

> Default is not needed when retrieving a value,

This doesn't work for me.

```ts
Property defaultValue is missing in type { id: symbol; } but required in type ContextKey<string | undefined>
 ctx.values.get<string | undefined>({
        id: Symbol('test'),
      });
```

> We just updated the docs with an [example](https://connectrpc.com/docs/node/interceptors#example).

There you made an example to authenticate within `contextValues`. Based on your previous suggestions, I think this should be removed.

srikrsna-buf:
The API is the same, you have to create a unique key using `createContextKey` and use that. `Symbol` constructor creates a unique symbol and cannot be used like this:
```ts
console.log(Symbol('foo') === Symbol('foo'));
// Expected output: false
```

StarpTech:
Ah, ok `createContextKey` is just a utility function. I think I know now where my confusion comes from. The API is just not very common in the Node.js space. Thanks again.

Issue #679, opened 2023-06-19
Can't enable lib checking in consumer tsconfig.json with ESM


fubhy:
**Describe the bug**

Currently, types are generated using ESM config but, unlike the files in `dist/esm`, do not have a parent `package.json` file with `type: module`. Hence, the types are considered CJS and cause typechecking errors when a consumer enables lib checking in their `tsconfig.json`.

**To Reproduce**

Set `skipLibCheck: false` in a consuming package and import from `@bufbuild/connect`. Since that imports from `@bufbuild/protobuf` (which has the types exported as ESM), you'll get the following error (and many others like it):

```
node_modules/@bufbuild/connect/dist/types/router.d.ts(1,46): error TS1479: The current file is a CommonJS module whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'. Consider writing a dynamic 'import("@bufbuild/protobuf")' call instead.
```

timostamm:
Just catching up 😅 

Good call - sounds like the "Masquerading as CJS" issue reported by arethetypeswrong. This should be fixed through https://github.com/connectrpc/connect-es/issues/801.

smaye81:
Hey @fubhy can your share your tsconfig `compilerOptions`? I'm guessing this is with `moduleResolution` of `Node16` correct?

smaye81:
This should be fixed in the v1.1.0 release.  Release notes [here](https://github.com/connectrpc/connect-es/releases/tag/v1.1.0).

PR #838, opened 2023-09-26
Integrate `arethetypeswrong`

smaye81:
This integrates the [`arethetypeswrong`](https://github.com/arethetypeswrong/arethetypeswrong.github.io/tree/main) CLI tool for checking types in published packages. It also adds the call to lint the types in CI that it is easy to spot regressions.

This was tested with `examples-es` using:
- react/esbuild (ESM with esbuild)
- react/rollup (ESM with rollup)
- react/webpack (ESM with webpack)
- react/webpack-cjs (CJS with webpack)

### DIFF
--- a/packages/connect-express/package.json
+++ b/packages/connect-express/package.json
@@ -9,18 +9,26 @@
   },
   "sideEffects": false,
   "scripts": {
-    "clean": "rm -rf ./dist/cjs/* ./dist/esm/* ./dist/types/*",
-    "build": "npm run build:cjs && npm run build:esm+types",
-    "build:cjs": "tsc --project tsconfig.json --module commonjs --outDir ./dist/cjs && echo >./dist/cjs/package.json '{\"type\":\"commonjs\"}'",
-    "build:esm+types": "tsc --project tsconfig.json --module ES2015 --verbatimModuleSyntax --outDir ./dist/esm --declaration --declarationDir ./dist/types"
+    "clean": "rm -rf ./dist/cjs/* ./dist/esm/*",
+    "build": "npm run build:cjs && npm run build:esm",
+    "build:cjs": "tsc --project tsconfig.json --module commonjs --outDir ./dist/cjs --declaration --declarationDir ./dist/cjs && echo >./dist/cjs/package.json '{\"type\":\"commonjs\"}'",
+    "build:esm": "tsc --project tsconfig.json --module ES2015 --verbatimModuleSyntax --outDir ./dist/esm --declaration --declarationDir ./dist/esm",
+    "lint:types": "attw --pack"
   },
   "main": "./dist/cjs/index.js",
   "type": "module",
-  "types": "./dist/types/index.d.ts",
+  "types": "./dist/esm/index.d.ts",
   "exports": {
-    "types": "./dist/types/index.d.ts",
-    "import": "./dist/esm/index.js",
-    "require": "./dist/cjs/index.js"
+    ".": {
+      "import": {
+        "types": "./dist/esm/index.d.ts",
+        "default": "./dist/esm/index.js"
+      },
+      "require": {
+        "types": "./dist/cjs/index.d.ts",
+        "default": "./dist/cjs/index.js"
+      }
+    }
   },
   "engines": {
     "node": ">=16.0.0"
--- a/packages/connect-express/package.json
+++ b/packages/connect-express/package.json
@@ -13,7 +13,7 @@
     "build": "npm run build:cjs && npm run build:esm",
     "build:cjs": "tsc --project tsconfig.json --module commonjs --outDir ./dist/cjs --declaration --declarationDir ./dist/cjs && echo >./dist/cjs/package.json '{\"type\":\"commonjs\"}'",
     "build:esm": "tsc --project tsconfig.json --module ES2015 --verbatimModuleSyntax --outDir ./dist/esm --declaration --declarationDir ./dist/esm",
-    "lint:types": "attw --pack"
+    "attw": "attw --pack"
   },
   "main": "./dist/cjs/index.js",
   "type": "module",
--- a/packages/connect-express/package.json
+++ b/packages/connect-express/package.json
@@ -17,7 +17,6 @@
   },
   "main": "./dist/cjs/index.js",
   "type": "module",
-  "types": "./dist/esm/index.d.ts",
   "exports": {
     ".": {
       "import": {
--- a/packages/connect/package.json
+++ b/packages/connect/package.json
@@ -10,55 +10,81 @@
   },
   "sideEffects": false,
   "scripts": {
-    "clean": "rm -rf ./dist/cjs/* ./dist/esm/* ./dist/types/*",
+    "clean": "rm -rf ./dist/cjs/* ./dist/esm/*",
     "generate": "buf generate src/protocol-grpc/proto",
-    "build": "npm run build:cjs && npm run build:esm+types && node scripts/update-user-agent.mjs",
-    "build:cjs": "tsc --project tsconfig.json --module commonjs --outDir ./dist/cjs && echo >./dist/cjs/package.json '{\"type\":\"commonjs\"}'",
-    "build:esm+types": "tsc --project tsconfig.json --module ES2015 --verbatimModuleSyntax --outDir ./dist/esm --declaration --declarationDir ./dist/types && echo >./dist/esm/package.json '{\"type\":\"module\", \"sideEffects\":false}'",
-    "jasmine": "jasmine --config=jasmine.json"
+    "build": "npm run build:cjs && npm run build:esm && node scripts/update-user-agent.mjs",
+    "build:cjs": "tsc --project tsconfig.json --module commonjs --outDir ./dist/cjs --declaration --declarationDir ./dist/cjs && echo >./dist/cjs/package.json '{\"type\":\"commonjs\"}'",
+    "build:esm": "tsc --project tsconfig.json --module ES2015 --verbatimModuleSyntax --outDir ./dist/esm --declaration --declarationDir ./dist/esm && echo >./dist/esm/package.json '{\"type\":\"module\", \"sideEffects\":false}'",
+    "jasmine": "jasmine --config=jasmine.json",
+    "lint:types": "attw --pack"
   },
   "main": "./dist/cjs/index.js",
-  "types": "./dist/types/index.d.ts",
+  "types": "./dist/esm/index.d.ts",
   "exports": {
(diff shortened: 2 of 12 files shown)


packages/connect-express/package.json
```diff
@@ -9,18 +9,26 @@
...
+    "clean": "rm -rf ./dist/cjs/* ./dist/esm/*",
+    "build": "npm run build:cjs && npm run build:esm",
+    "build:cjs": "tsc --project tsconfig.json --module commonjs --outDir ./dist/cjs --declaration --declarationDir ./dist/cjs && echo >./dist/cjs/package.json '{\"type\":\"commonjs\"}'",
+    "build:esm": "tsc --project tsconfig.json --module ES2015 --verbatimModuleSyntax --outDir ./dist/esm --declaration --declarationDir ./dist/esm",
+    "lint:types": "attw --pack"
   },
   "main": "./dist/cjs/index.js",
   "type": "module",
-  "types": "./dist/types/index.d.ts",
+  "types": "./dist/esm/index.d.ts",
```
smaye81, line 20: Unclear what this should be. This property is set for older versions of TypeScript (< 4.7) that don't support conditional exports, but those could be either ESM or CJS, so it's not entirely clear what the types fallback should be.

packages/connect/package.json
```diff
@@ -10,55 +10,81 @@
...
+        "types": "./dist/cjs/protocol-grpc-web/index.d.ts",
+        "default": "./dist/cjs/protocol-grpc-web/index.js"
+      }
     }
   },
   "typesVersions": {
     "*": {
       "protocol": [
-        "./dist/types/protocol/index.d.ts"
+        "./dist/esm/protocol/index.d.ts"
```
smaye81, line 78: Likewise with above, `typesVersions` is a fallback for older versions of TypeScript that do not prioritize subpath exports. However, it's unclear if this should be ESM or CJS?

packages/connect/src/protocol-connect/end-stream.ts
```diff
@@ -21,7 +21,7 @@ import { errorFromJson, errorToJson } from "./error-json.js";
 import { appendHeaders } from "../http-headers.js";
 import { ConnectError } from "../connect-error.js";
 import { Code } from "../code.js";
-import type { Serialization } from "../protocol";
+import type { Serialization } from "../protocol/serialization.js";
```
smaye81, line 24: This fixes an [Internal Resolution Error](https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/InternalResolutionError.md) with Node 16.

timostamm (commented):
> This was tested with `examples-es` using:
> 
> * react/esbuild (ESM with esbuild)
> * react/rollup (ESM with rollup)
> * react/webpack (ESM with webpack)
> * react/webpack-cjs (CJS with webpack)

That's great, now we know that bundlers work as expected with the changed exports.

But besides bundlers, Node.js and TypeScript also resolve modules. I wonder how well `attw` guards against mistakes in our exports for the three different consumers. It would be really nice to know what is covered and what isn't. Can we find out?

packages/connect/src/protocol-connect/end-stream.ts
timostamm: Reported by `attw` I assume?

I can reproduce with:

<details open>
<summary>tsconfig.json</summary>

```json
{
  "files": ["test.ts"],
  "compilerOptions": {
    "module": "node16",
    "skipLibCheck": false
  }
}
```
</details>


<details open>
<summary>test.ts</summary>

```ts
import {endStreamFlag} from "@connectrpc/connect/protocol-connect";

console.log(endStreamFlag)
```
</details>

```bash
$ npx tsc                                                                                    
../connect/dist/esm/protocol-connect/end-stream.d.ts:3:36 - error TS2834: Relative import paths need explicit file extensions in EcmaScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Consider adding an extension to the import path.

3 import type { Serialization } from "../protocol";
                                     ~~~~~~~~~~~~~


Found 1 error in ../connect/dist/esm/protocol-connect/end-stream.d.ts:3
```

Nice, this is a genuine issue found by `attw`. 

The doc you linked provides further guidance:

> Library authors should always compile/check with --module node16 --moduleResolution node16 or nodenext (note that module implies the corresponding moduleResolution value, but not vice versa), because this is the only TypeScript mode that recognizes Node’s strict module resolution algorithm for ESM.

The `Node` module resolution algorithm we're using is effectively Node v10. I think it's a good idea to switch to `node16` across the board, but I believe this PR is much easier to reason about if we do it separately (separate PR).

timostamm:
Were you able to reproduce https://github.com/connectrpc/connect-es/issues/679 and confirm that it's fixed with this?

packages/connect-express/package.json
```diff
@@ -9,18 +9,26 @@
...
   "scripts": {
-    "clean": "rm -rf ./dist/cjs/* ./dist/esm/* ./dist/types/*",
-    "build": "npm run build:cjs && npm run build:esm+types",
-    "build:cjs": "tsc --project tsconfig.json --module commonjs --outDir ./dist/cjs && echo >./dist/cjs/package.json '{\"type\":\"commonjs\"}'",
-    "build:esm+types": "tsc --project tsconfig.json --module ES2015 --verbatimModuleSyntax --outDir ./dist/esm --declaration --declarationDir ./dist/types"
+    "clean": "rm -rf ./dist/cjs/* ./dist/esm/*",
+    "build": "npm run build:cjs && npm run build:esm",
+    "build:cjs": "tsc --project tsconfig.json --module commonjs --outDir ./dist/cjs --declaration --declarationDir ./dist/cjs && echo >./dist/cjs/package.json '{\"type\":\"commonjs\"}'",
+    "build:esm": "tsc --project tsconfig.json --module ES2015 --verbatimModuleSyntax --outDir ./dist/esm --declaration --declarationDir ./dist/esm",
+    "lint:types": "attw --pack"
```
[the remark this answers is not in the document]
smaye81: Yep, done.

packages/connect/src/protocol-connect/end-stream.ts
smaye81: It was uncovered by `attw`, yeah. And yeah, sounds good re: `node16`. We can def do that in a follow-up PR.

smaye81:
> > This was tested with `examples-es` using:
> > 
> > * react/esbuild (ESM with esbuild)
> > * react/rollup (ESM with rollup)
> > * react/webpack (ESM with webpack)
> > * react/webpack-cjs (CJS with webpack)
> 
> That's great, now we know that bundlers work as expected with the changed exports.
> 
> But besides bundlers, Node.js and TypeScript also resolve modules. I wonder how well `attw` guards against mistakes in our exports for the three different consumers. It would be really nice to know what is covered and what isn't. Can we find out?

I ran some tests on a branch in `examples-es` that uses just Node by having a single file (`main.js`) that imports `@connectrpc/connect` from my local machine (using `npm link`). I was able to run `node main.js` and verify that the import was working fine by printing out the imported symbols.

I also did this same test with a file `test.ts` which imported `@connectrpc/connect`. I then ran `tsc` against this file to verify things were working correctly.

I'm not sure if this answers your question or provides sufficient confidence that this works for Node and TS. Let me know though.

packages/connect/package.json
timostamm: Same as in https://github.com/connectrpc/connect-es/pull/838#discussion_r1338555849, I'm expecting this to be relevant with the Node10 version resolution only, and since Node10 did not support ESM, these should point to the CJS types.

timostamm:
> I ran some tests on a branch in `examples-es` that uses just Node by having a single file (`main.js`) that imports `@connectrpc/connect` from my local machine (using `npm link`). I was able to run `node main.js` and verify that the import was working fine by printing out the imported symbols.

Nice. I suppose with scripts like these?

<details><summary>t.mjs</summary>

```js
import { ConnectError as a} from "@connectrpc/connect";
import { createFetchClient as b } from "@connectrpc/connect/protocol";
import { createTransport as c } from "@connectrpc/connect/protocol-grpc";
import { createTransport as d } from "@connectrpc/connect/protocol-grpc-web";
import { createTransport as e } from "@connectrpc/connect/protocol-connect";
const ok = !!a && !!b && !!c && !!d && !!e;
console.log(ok ? "ok" : "fail")
```
</details>


<details><summary>t.cjs</summary>

```js
const { ConnectError: a } = require("@connectrpc/connect");
const { createFetchClient: b } = require("@connectrpc/connect/protocol");
const { createTransport: c } = require("@connectrpc/connect/protocol-grpc");
const { createTransport: d } = require("@connectrpc/connect/protocol-grpc-web");
const { createTransport: e } = require("@connectrpc/connect/protocol-connect");
const ok = !!a && !!b && !!c && !!d && !!e;
console.log(ok ? "ok" : "fail")
```
</details>

Running both scripts with Node.js LTS from v10 to v20 (`fnm exec --using=v20 node t.cjs`): Only v10 with `t.mjs` fails, which is expected because this version doesn't support ESM.

`t.mjs` only fails with Node.js v10 -  expected because it never supported ESM 👍 

Now to break it:

```bash
$ cd packages/connect
$ npm run build
$ mv protocol.js protocol.js-x
$ npx attw --pack
```

No error from `attw`, but now Node.js v10 also fails `t.cjs` :boom: 

It's a bummer that `attw` does not validate these .js exports. They're handwritten and most susceptible to break. 

This makes me wonder: Clearly we have to support Node.js v10 module resolution for types, since it's not uncommon to use `"moduleResulution": "Node"` in a tsconfig. But there is no point in support Node.js v10 itself at runtime - it's been EOL for more than two years. 

So I wonder whether we can actually safely delete `protocol.js`, `protocol-connect.js`, etc. 🤔 
This could potentially break some bundlers, so it's best to keep them for now, and investigate that separately.

> I also did this same test with a file test.ts which imported @connectrpc/connect. I then ran tsc against this file to verify things were working correctly.

 Seems sufficient 👍

The only open questions from my end are https://github.com/connectrpc/connect-es/pull/838#discussion_r1337576110 and https://github.com/connectrpc/connect-es/pull/838#issuecomment-1737395643.

packages/connect/package.json
smaye81: Got it. Updated.

smaye81:
> Were you able to reproduce #679 and confirm that it's fixed with this?

Whoops, sorry, missed this. Yes, I was able to recreate #679 and can confirm that this fixes that issue.

timostamm (approved):
LGTM!

PR #670, opened 2023-06-06
Support fetch api request stream

oott123:
This fixes #669 for connect transport. Maybe we should support it for grpc transport too.

PR #840, opened 2023-09-28
fix: browser fetch can support streaming request bodies

tshedor:
This is a bug fix related to client-streaming requests. 

Invoking a `MethodKind.ClientStreaming` method using a `connect` or `grpc-web` transport immediately throws the error "The fetch API does not support streaming request bodies."

This, however, appears to be a dated catch. Streaming as a request body is supported [by most modern browsers](https://caniuse.com/?search=TransformStream) using the default `fetch`. 

This PR does a [very light feature check](https://developer.chrome.com/articles/fetch-streaming-requests/#feature-detection) to ensure backwards compatibility instead of checking against the method's kind.

### DIFF
--- a/packages/connect-web/src/connect-transport.ts
+++ b/packages/connect-web/src/connect-transport.ts
@@ -270,7 +270,7 @@ export function createConnectTransport(
       async function createRequestBody(
         input: AsyncIterable<I>,
       ): Promise<Uint8Array> {
-        if (method.kind != MethodKind.ServerStreaming) {
+        if (!TransformStream) {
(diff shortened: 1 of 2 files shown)


tshedor:
This may be a half-baked duplicate of #670 , and if that's the case, I'm sorry for adding to the maintainer's administrative work

timostamm:
Hey Tim, thanks for bringing it up! Unfortunately, only Chrome supports streaming request bodies yet, but it's not obvious from the documentation out there.

PR #841, opened 2023-09-29
Add context values to clients

srikrsna-buf:
Add context values to clients. This adds support for `ContextValues` in clients via `CallOptions` and `Interceptors`: 

```ts
import { ElizaService } from "gen/...";
import { createPromiseClient } from "@connectrpc/connect";
import transport from "./transport.js";
import kUser from "user-context.js";

const client = createPromiseClient(ElizeService, trasport);

await client.say({ sentence: "Hi" }, { values: createContextValues().set(kUser, { ... }) });
```

Which can be accessed in an interceptor:

```ts
const tokenInterceptor = (next) => {
    return (req) => {
           req.header.set("Authorization", `Bearer ${req.values.get(kUser).token}`);
           return next(req);
    };
};
```

### DIFF
--- a/packages/connect-node/src/node-universal-handler.ts
+++ b/packages/connect-node/src/node-universal-handler.ts
@@ -127,7 +127,7 @@ export function universalRequestFromNodeRequest(
     header: nodeHeaderToWebHeader(nodeRequest.headers),
     body,
     signal: abortController.signal,
-    values: contextValues,
+    contextValues: contextValues,
   };
 }
 
--- a/packages/connect-web-bench/README.md
+++ b/packages/connect-web-bench/README.md
@@ -10,5 +10,5 @@ it like a web server would usually do.
 
 | code generator | bundle size        | minified               | compressed           |
 |----------------|-------------------:|-----------------------:|---------------------:|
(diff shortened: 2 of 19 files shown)


paul-sachs:
Love the simplicity. My one tiny nitpick is that the key on the Request object is just `values`, which on first read feels like it's some dictionary representation of the content of the request. Perhaps calling it `context` instead? I guess `values` just feels a touch generic.

srikrsna-buf:
> Perhaps calling it `context` instead? I guess `values` just feels a touch generic.

I also prefer that, but a potential problem I see is that when we implement server-side interceptors, one of the possible designs is to just reuse the interceptor types that we use on the client. `context` could be confusing to users as they might expect `HandlerContext`.

If an alternate design for interceptors is possible then I don't see a problem with naming this context. My efforts to get TS to work have been unsuccessful so far I tried a couple of designs and the ones we have for the client seem the most promising but nevertheless we can wait until at least we finalize the design

srikrsna-buf:
I changed the name to `contextValues` I also updated `UniversalServerRequest`. The only place it is called `values` is on `HandlerContext`.

smaye81 (approved):

Issue #839, opened 2023-09-27
Investigate removing subpath fallback files


smaye81:
Currently, we have fallback files in `packages/connect` for consumers that don't support subpath exports (`protocol.js`, `protocol-connect.js`, `protocol-grpc.js`, and `protocol-grpc-web.js`). 

However, these files may not be needed as it is basically supporting Node.js v10 module resolution. Since Node 10 has been EOL for more than two years, we could potentially remove these altogether.

This could potentially break some bundlers, so we'd have to conduct an investigation into whether we could safely remove them.

For further context, see https://github.com/connectrpc/connect-es/pull/838#issuecomment-1737869511

timostamm:
We have removed the fallback files in https://github.com/bufbuild/protobuf-es/pull/804.
If this does not cause any issues, we can safely do it here as well.

timostamm:
With Parcel and Metro, subpath exports need to be enabled, see https://github.com/bufbuild/protobuf-es/blob/main/MANUAL.md#parcelmetro-fails-to-resolve-imports

timostamm:
V2 removes subpath fallbacks for Node10.

PR #842, opened 2023-09-29
Add ESM wrapper to avoid dual package hazard

timostamm:
This fixes the dual package hazard with the solution proposed in https://github.com/bufbuild/protobuf-es/issues/509#issuecomment-1634670621. 

We've tested the approach in https://github.com/connectrpc/examples-es/pull/1002 in various versions of Node.js, esbuild, parcel, rollup, vite, and webpack. Tree-shaking continues to work as expected with the "module" condition, and the dual package hazard is resolved with the esm wrappers in the "import" condition.

Note that we bump dependencies on `@bufbuild/protobuf` to `^1.3.3` with this PR, meaning that [1.3.3](https://github.com/bufbuild/protobuf-es/releases/tag/v1.3.3) or later needs to be installed. This is necessary because users are very likely to run into a dual package hazard otherwise.


packages/connect/src/protocol/is-protobuf-message.ts
```diff
@@ -0,0 +1,34 @@
...
+// limitations under the License.
+
+import { Message } from "@bufbuild/protobuf";
+
+/**
+ * Temporary workaround for dual package hazard caused by @bufbuild/protobuf
+ *
+ * @private Internal code, does not follow semantic versioning.
+ */
+export function isProtobufMessage<T extends Message<T>>(
```
timostamm, line 22: This is a temporary workaround for the dual package hazard caused by protobuf-es. Since the ESM wrapper delegates to CJS, we're internally using `Message` from protobuf-es in CJS, but the tests are providing `Message` from protobuf-es in ESM.

I don't want to land this - we have to fix and release protobuf-es first, than revert https://github.com/connectrpc/connect-es/pull/842/commits/b57115ce70005b256530815db2917fb6e4604984 here.

packages/connect-web-bench/README.md
```diff
@@ -10,5 +10,5 @@ it like a web server would usually do.
 
 | code generator | bundle size        | minified               | compressed           |
 |----------------|-------------------:|-----------------------:|---------------------:|
-| connect        | 113,448 b | 49,872 b | 13,465 b |
+| connect        | 113,741 b | 50,030 b | 13,533 b |
```
timostamm, line 13: The number are only slightly bumped because of https://github.com/connectrpc/connect-es/pull/842/commits/b57115ce70005b256530815db2917fb6e4604984. 
So tree-shaking is working as expected via the "module" export condition - at least in esbuild.

paul-sachs: I can confirm that tree shaking also still works in webpack. Currently looking at parcel, though having problems getting parcel tree shaking at all.

paul-sachs: Further digging into parcel looks like the it would resolve to cjs in the old version and seems to resolve to esm with the new version. However, both don't seem to be doing any tree shaking either way.

smaye81:
For my findings re: general interop between bundlers, I tested **Webpack**, **Parcel**, **Rollup**, **esbuild**, and **Vite**.

For **Webpack**, **Parcel**, **Rollup**, and **esbuild**, I can get the dual package hazard to occur and can confirm installing the fix tarball fixes the problem in all 4 bundlers.

For **Vite**, I can't seem to get the dual package hazard to occur. I tried various incantations in my `vite.config.js` file, but every output builds fine and all tests for the dual package hazard pass. They also pass with the installed fix tarball, also, so it may not be an issue. Not sure if I'm missing something in the config though.

paul-sachs:
Investigations into tree shaking look good as well. No significant difference between bundle sizes between the current 1.0 and this version. Tried with both direct and intermediary examples, using **webpack**, **esbuild**, and **parcel**. Parcel was the only slight outlier in that I couldn't get tree shaking to work at all, regardless of which version I used

timostamm:
Sounds great, @smaye81. Can you push the tests up somewhere (examples-es) so we can re-use for Node16 module resolution and https://github.com/connectrpc/connect-es/issues/839 as discussed?

timostamm:
> For **Vite**, I can't seem to get the dual package hazard to occur. I tried various incantations in my `vite.config.js` file, but every output builds fine and all tests for the dual package hazard pass. They also pass with the installed fix tarball, also, so it may not be an issue. Not sure if I'm missing something in the config though.

I can reproduce the error with vite. In the branch `tstamm/dual-package-hazard-example` in examples-es:

Append to `react/vite/src/App.tsx`:

```ts
import { ConnectError as a } from "@connectrpc/connect";
import { createFetchClient as b } from "@connectrpc/connect/protocol";
import { createTransport as c } from "@connectrpc/connect/protocol-grpc";
import { createTransport as d } from "@connectrpc/connect/protocol-grpc-web";
import { createTransport as e } from "@connectrpc/connect/protocol-connect";

const direct = {
    a,
    b,
    c,
    d,
    e,
};

// @ts-ignore
import { intermediary } from "intermediary";

for (const name of "abcde".split("")) {
    // @ts-ignore
    const directSymbol = direct[name];
    const interSymbol = intermediary[name];
    if (!directSymbol) {
        throw `missing direct ${name}`;
    }
    if (!interSymbol) {
        throw `missing intermediary ${name}`;
    }
    if (directSymbol === interSymbol) {
        console.log(`${name} OK`)
    } else {
        console.error(`${name} FAIL`)
    }
}
```

```
cd react/vite
npm ci
cp -rp ../../dual-package-hazard-example/intermediary node_modules
npm start
```

Screenshot 2023-10-04 at 16 42 57

timostamm:
> Investigations into tree shaking look good as well. No significant difference between bundle sizes between the current 1.0 and this version. Tried with both direct and intermediary examples, using **webpack**, **esbuild**, and **parcel**. Parcel was the only slight outlier in that I couldn't get tree shaking to work at all, regardless of which version I used

Thanks, @paul-sachs. Can you push the code up? It does not need to be polished at all, just repeatable.

paul-sachs:
@timostamm pushed up to your dual-package-hazard-example. Just need to run `bundle` command in the consumer package and inspect the resulting file sizes for all the different bundlers.

fubhy:
This is great! I appreciate all the work and research that went into this. Looking forward to the next release :)

packages/connect/src/protocol/is-protobuf-message.ts
timostamm: Updated to [@bufbuild/protobuf v1.3.3](https://github.com/bufbuild/protobuf-es/releases/tag/v1.3.3) in https://github.com/connectrpc/connect-es/pull/842/commits/567b5c99933daf62175ac4cbe1e74db874842025.

Reverted this temporary workaround in https://github.com/connectrpc/connect-es/pull/842/commits/598f1e40ae24b770366677e067680c090e80cd60.

timostamm: Repeated the tests from https://github.com/connectrpc/examples-es/pull/1002 successfully.

paul-sachs (approved):

PR #843, opened 2023-09-29
Add support for graceful shutdown in fastify

### LINKED ISSUE #802: Use Fastify hooks to shut down services gracefully
When setting up a server, it is possible to pass a [shutdownSignal](https://github.com/connectrpc/connect-es/blob/a03b06539b97264800790b4a317d3187e094c8a7/packages/connect/src/protocol/universal-handler.ts#L104):

```ts
  /**
   * To shut down servers gracefully, this option takes an AbortSignal.
   * If this signal is aborted, all signals in handler contexts will be aborted
   * as well. This gives implementations a chance to wrap up work before the
   * server process is killed.
   * Abort this signal with a ConnectError to send a message and code to
   * clients.
   */
  shutdownSignal?: AbortSignal;
```

This signal will also trigger the [signal in the HandlerContext](https://github.com/connectrpc/connect-es/blob/a03b06539b97264800790b4a317d3187e094c8a7/packages/connect/src/implementation.ts#L84) available to all RPC implementations, allowing them to close any open resources.

This is a convenient feature for microservices that might teared down frequently, but the feature is not well integrated into server plugins.

Fastify has a [close and preClose hooks](https://fastify.dev/docs/latest/Reference/Hooks/#preclose) that [@connectrpc/connect-fastify](https://www.npmjs.com/package/@connectrpc/connect-fastify) could use to trigger the signal automatically.

**Additional context**
Usually, services are allowed a grace period to shut down cleanly, but it is unclear how the service could signal back that it is finished.

srikrsna-buf:
Closes #802.

Once `Fastify.close` is called it waits for requests to complete and immediately blocks new ones. We use the `preClose` hook, which is called right when `close` is called, to start a shutdown timeout for requests and close them if the time runs out.

The other option is just to document this so that users can set the `shutdownSignal` as we do in this PR.


packages/connect-fastify/src/fastify-connect-plugin.ts
```diff
@@ -60,6 +74,16 @@ export function fastifyConnectPlugin(
...
     opts.acceptCompression = [compressionGzip, compressionBrotli];
   }
+  if (opts.shutdownTimeoutMs !== undefined) {
+    const shutdownController = createLinkedAbortController(opts.shutdownSignal);
+    opts.shutdownSignal = shutdownController.signal;
+    instance.addHook("preClose", (done) => {
+      setTimeout(() => {
+        shutdownController.abort(opts.shutdownError);
+      }, opts.shutdownTimeoutMs);
+      done();
```
timostamm, line 84: Shouldn't `done` be called _after_ the grace period? So in the setTimeout callback?

srikrsna-buf: Either should be fine, Fastify waits for the requests to finish even there are no `preClose`hooks.

timostamm (approved):

PR #861, opened 2023-10-03
Fix early event loop exit on nodejs when H2 sessions are in the verify state

### LINKED ISSUE #860: Pinging idle connection before sending request is ignored and node process exits unexpectedly
**Describe the bug**

After being idle from a long time and calling a method, the second call is skipped completely and the node process exits unexpectedly.
This is because the ping call in the verify state is done with the socket unrefed.

**To Reproduce**
Tested this on linux using latest version:

```js
    const transport = createConnectTransport({
        baseUrl: serviceUrl.toString(),
        httpVersion: '2',
        interceptors: [authInterceptor],
        useBinaryFormat: true,
       pingIntervalMs: 120000,
    });

    const service = createPromiseClient(MyService, transport);

        const response1 = await service.method({});
        console.log(`===> ${new Date()} On method response}`, response1);

        console.log(`===> ${new Date()} Going to sleep`);
        await timeout(10*60*1000);
        console.log(`===> ${new Date()} Awake`);

        // This await is ignored and the node process exits unexpectedly
        const response2 = await service.method({});
        console.log(`===> ${new Date()} On method response2}`, response2);
```

**Additional context**

srikrsna-buf:
Fix early event loop exit on nodejs when H2 sessions are in the verify state. We `unref` the connections to avoid idle connections stopping the process from exiting. This results in a case where process exits when verification is required (last pinged time elapsed ping interval). We always `ref` before a manual `ping` and `unref` if it succeeds with empty streams. 

Used a worker to test the behavior. Not super happy about it but seemed better than a child process.

Fixes #860.

### DIFF
--- a/packages/connect-node/.npmignore
+++ b/packages/connect-node/.npmignore
@@ -2,3 +2,4 @@
 /*.json
 /dist/*/**/*.spec.js
 /dist/*/**/*.spec.d.ts
+/dist/*/testdata/
--- a/packages/connect-node/src/http2-session-manager.spec.ts
+++ b/packages/connect-node/src/http2-session-manager.spec.ts
@@ -16,6 +16,7 @@ import { useNodeServer } from "./use-node-server-helper.spec.js";
 import * as http2 from "http2";
 import { Http2SessionManager } from "./http2-session-manager.js";
 import { ConnectError } from "@connectrpc/connect";
+import { Worker } from "worker_threads";
 
 describe("Http2SessionManager", function () {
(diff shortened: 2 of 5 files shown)


timostamm (approved):
LGTM. 

Can you update the PR title description with more details? They are basically our changelog, and we should take the extra couple of minutes it takes to make it really easy to see what's changed from release notes, with a link to a PR providing details.

For example, this PR title:

> Fix early event loop exit on Node.js when H2 sessions are in the `verify` state

... would be more likely to ring a bell for a user who saw an unexpected early exit in their client app.

PR #876, opened 2023-10-07
Fix multiple read attempts in fetch client

srikrsna-buf:
Fix multiple read attempts in fetch client. I caught this while working on https://github.com/connectrpc/examples-es/pull/1075. The fix similar to how we read the iterable: https://github.com/connectrpc/connect-es/blob/51470ebc229130150d5ba4147fe32b9c142cb83f/packages/connect/src/protocol/universal-fetch.ts#L129-L130

### DIFF
--- a/packages/connect/src/protocol/universal-fetch.ts
+++ b/packages/connect/src/protocol/universal-fetch.ts
@@ -152,9 +152,9 @@ function iterableToReadableStream(
 function iterableFromReadableStream(
   body: ReadableStream<Uint8Array> | null,
 ): AsyncIterable<Uint8Array> {
+  const reader = body?.getReader();
   return {
     [Symbol.asyncIterator](): AsyncIterator<Uint8Array> {
-      const reader = body?.getReader();
       return {
         async next() {
           if (reader !== undefined) {


timostamm (commented):
From https://github.com/connectrpc/examples-es/pull/1075:

> Using the fetch client from connect, the client for server steaming throws an error at the very end of the steam with: Invalid state: ReadableStream is locked.

Seems like some code is calling `[Symbol.asyncIterator]()` twice. If this is in our code base, it would be great to understand where this happens, and avoid it. Did you take a look at how the duplicate call occurs?

srikrsna-buf:
> Seems like some code is calling `[Symbol.asyncIterator]()` twice. If this is in our code base, it would be great to understand where this happens, and avoid it. Did you take a look at how the duplicate call occurs?

It seems to be this line: https://github.com/connectrpc/connect-es/blob/e0bffbab4e75e19fd7eeb9eadabe050941d39e5f/packages/connect/src/protocol/async-iterable.ts#L596-L599 in `pipe`.

timostamm:
Yeah, looks like that line in async-iterable.ts would do it. I believe we should avoid that. 

If you iterate over an Array or Map, you would expect a second iteration to restart from the first entry. Under the hood, `[Symbol.iterator]` is called once per iteration. I would expect `[Symbol.asyncIterator]` to have the same semantics. By calling it twice for what is effectively a single iteration, we dip out toes into undefined behavior.

timostamm (approved):

PR #884, opened 2023-10-25
Introduce ConnectRouter.rpc overload to not require full ServiceType

paul-sachs:
Partly copied from https://github.com/bufbuild/protobuf-es/pull/594:

`ConnectRouter.rpc` overload that doesn't require the full service is easier to use for one off methods that don't require an entire service to be defined.


paul-sachs:
@timostamm resuming the conversation from https://github.com/bufbuild/protobuf-es/pull/594

> 1. Can we omit methods from service and still use a full ServiceType for the property?

I'm not 100% sure what you mean. The omitted field does allow us to assign a full ServiceType value to this, it just doesn't allow us to rely on it downstream.

> Do we need localName? It makes it much more cumbersome to construct a MethodType from just a ServiceType.

From what I can see, it's not required. @srikrsna-buf was there a specific use case you were thinking of?

packages/connect/src/router.ts
```diff
@@ -131,14 +138,42 @@ export function createConnectRouter(
...
-          protocols,
-        ),
-      );
+    rpc(
+      serviceOrMethod: ServiceType | MethodType,
+      methodOrImpl: MethodInfo | MethodImpl<MethodInfo<AnyMessage>>,
+      implementationOrOptions?:
+        | MethodImpl<MethodInfo<AnyMessage>>
+        | Partial<UniversalHandlerOptions>,
+      options?: Partial<UniversalHandlerOptions>,
```
paul-sachs, line 147: There's probably a better way to do, similar to how it was overridden in `packages/connect/src/implementation.ts` but I just couldn't figure it out. I tried using the same array format but it caused a weird typescript error. I assume it's because unlike in implementation, this `rpc` is not declared near the function overloads.

packages/connect/src/implementation.ts
```diff
@@ -251,14 +252,36 @@ export type ServiceImplSpec = {
  * Create an MethodImplSpec - a user-provided implementation for a method,
  * wrapped in a discriminated union type along with service and method metadata.
  */
+export function createMethodImplSpec<M extends MethodType>(
+  method: M,
+  impl: MethodImpl<M>,
+): MethodImplSpec;
```
timostamm, line 258: I would prefer if we only support `MethodType` in `ConnectRouter.rpc` in this PR to keep the change well-scoped.

packages/connect/src/router-transport.spec.ts
```diff
@@ -224,4 +225,25 @@ describe("createRoutesTransport", function () {
       });
     });
   });
+  describe("supports self describing method type", function () {
```
timostamm, line 228: Great call to add a test 🙏 

We're not really testing a feature of `router-transport.ts`, but a feature of `router.ts`. Can you instead move the test to a new `router.spec.ts` file? You can simply inspect `ConnectRouter.handlers` to verify.

packages/connect/src/method-type.ts
```diff
@@ -0,0 +1,83 @@
...
+ */
+export interface MethodTypeBiDiStreaming<
+  I extends Message<I>,
+  O extends Message<O>,
+> extends mtShared,
+    MethodInfoBiDiStreaming<I, O> {}
+
+/**
+ * MethodType represents a self-contained method type. It must contain
+ * references to the service that implements it.
```
timostamm, line 65: Nice. Ultimately, this should be moved to protobuf-es. Can you add a short explanation and link to https://github.com/bufbuild/protobuf-es/pull/594 ?

packages/connect/src/router.ts
```diff
@@ -63,6 +64,11 @@ export interface ConnectRouter {
     impl: MethodImpl<M>,
     options?: Partial<UniversalHandlerOptions>,
   ): this;
+  rpc<M extends MethodType>(
+    method: M,
+    impl: MethodImpl<M>,
+    options?: Partial<UniversalHandlerOptions>,
+  ): this;
```
timostamm, line 71: Some comments on the methods would be nice to have now.

packages/connect/src/router.ts
```diff
@@ -131,14 +138,42 @@ export function createConnectRouter(
...
-      const { protocols } = whichProtocols(options, base);
-      handlers.push(
-        createUniversalMethodHandler(
-          createMethodImplSpec(service, method, implementation),
-          protocols,
-        ),
-      );
+    rpc(
+      serviceOrMethod: ServiceType | MethodType,
+      methodOrImpl: MethodInfo | MethodImpl<MethodInfo<AnyMessage>>,
```
timostamm, line 143: No need to specify the default type parameter:

```suggestion
      methodOrImpl: MethodInfo | MethodImpl<MethodInfo>,
```

packages/connect/src/router.ts
```diff
@@ -131,14 +138,42 @@ export function createConnectRouter(
...
+      implementationOrOptions?:
+        | MethodImpl<MethodInfo<AnyMessage>>
+        | Partial<UniversalHandlerOptions>,
+      options?: Partial<UniversalHandlerOptions>,
+    ) {
+      if ("typeName" in serviceOrMethod) {
+        const { protocols } = whichProtocols(options, base);
+
```
timostamm, line 151: You will still need to cast, but you can do it at one place:

```ts
      let service: ServiceType;
      let method: MethodInfo;
      let impl: MethodImpl<MethodInfo>;
      let opt: Partial<UniversalHandlerOptions> | undefined;
      if ("typeName" in serviceOrMethod) {
        service = serviceOrMethod;
        method = methodOrImpl as MethodInfo;
       // ...
     } else {
       // ...
     }
     // create handler
```

Then the casts aren't interleaved with creating the handler, and they become a bit easier to reason about.

paul-sachs: Oh yeah, that is WAY better. I'll update.

packages/connect/src/router.ts
```diff
@@ -52,17 +53,31 @@ import type { ProtocolHandlerFactory } from "./protocol/protocol-handler-factory
...
+   * Provides implementation for a single RPC given service and associated method.
+   */
   rpc<M extends MethodInfo>(
     service: ServiceType,
     method: M,
     impl: MethodImpl<M>,
     options?: Partial<UniversalHandlerOptions>,
   ): this;
+  /**
+   * Provides implementation for a single RPC given a method type.
```
timostamm, line 74: I'm in favor of the proposed connect-query API, but we're still working with a draft PR with "experimental" in the name. This new feature will be out of place if we don't ship the new API for whatever reason.

Can we mark it `@private`? I wouldn't have any concerns publishing it then.

```suggestion
   * Provides implementation for a single RPC given a method type.
   * 
   * @private This is an experimental API. Please do not rely on it yet.
```

timostamm (commented):
Looks great to me. 

@srikrsna-buf, do you know of a use case for `localName` we missed? https://github.com/bufbuild/protobuf-es/pull/594#issuecomment-1780950808

timostamm (approved):

paul-sachs:
Continuing PR in fork based PR: https://github.com/connectrpc/connect-es/pull/925

PR #890, opened 2023-10-27
Add conformance client for node

srikrsna-buf:
Add conformance client for node. 

Tested with following config:
```yaml
features:
  versions:
    - HTTP_VERSION_1
    - HTTP_VERSION_2
  protocols:
    - PROTOCOL_CONNECT
    - PROTOCOL_GRPC
    - PROTOCOL_GRPC_WEB
  codecs:
    - CODEC_PROTO
    - CODEC_JSON
  compressions:
    - COMPRESSION_IDENTITY
    - COMPRESSION_GZIP
    - COMPRESSION_BR
  supportsTls: true
  supportsTlsClientCerts: true
  supportsConnectGet: true
  supportsHalfDuplexBidiOverHttp1: true
  requiresConnectVersionHeader: false
  supportsMessageReceiveLimit: true

```


packages/connect-node-conformance/src/transport.ts
```diff
@@ -0,0 +1,79 @@
...
+    case ConformanceCompression.IDENTITY:
+      break;
+  }
+
+  const sharedOptions = {
+    baseUrl,
+    httpVersion,
+    useBinaryFormat: req.codec === Codec.PROTO,
+    sendCompression,
+  };
```
timostamm, line 61: Since we're using `Any`, we have to provide a type registry in the JSON options here.

packages/connect-node-conformance/src/client.ts
```diff
@@ -0,0 +1,48 @@
...
+    },
+  },
+});
+
+const req =
+  flags.json === true
+    ? ClientCompatRequest.fromJsonString(readFileSync(0, "utf-8"))
+    : ClientCompatRequest.fromBinary(
+        new Uint8Array(Buffer.from(readFileSync(0, "binary"), "binary"))
+      );
```
timostamm, line 25: Conformance clients will have to read multiple requests from stdin. For example from an earlier exploration:

```ts
import {ClientCompatRequest, ClientCompatResponse} from "../gen/connectrpc/conformance/v1alpha1/client_compat_pb";
import {protoDelimited} from "@bufbuild/protobuf";

export interface ClientUnderTest {
    name: string;
    version: string;
    run(input: AsyncIterable<ClientCompatRequest>): AsyncIterable<ClientCompatResponse>;
}

export function runNodeJs(client: ClientUnderTest): void {
    const args = process.argv.slice(2);
    if ((args.length === 1 && args[0] === "-v") || args[0] === "--version") {
        process.stdout.write(`${client.name} ${client.version}\n`);
        process.exit(0);
    }
    if (args.length !== 0) {
        process.stderr.write(
            `${client.name} accepts length-delimited ${ClientCompatRequest.typeName} on stdin and writes length-delimited ${ClientCompatResponse.typeName} to stdout\n`,
        );
        process.exit(1);
    }
    runClientWithStdio(client).then(() => {
        process.exit(0);
    }, reason => {
        process.stderr.write(`${client.name} failed: ${String(reason)}\n`);
        process.exit(1);
    });
}

async function runClientWithStdio(client: ClientUnderTest) {
    const input = protoDelimited.decStream(ClientCompatRequest, process.stdin);
    for await (const res of client.run(input)) {
        const bytes = protoDelimited.enc(res)
        process.stdout.write(bytes);
    }
}
```

packages/connect-node-conformance/src/invoke.ts
```diff
@@ -0,0 +1,288 @@
...
+        onTrailer(trailers) {
+          resTrailers = convertToProtoHeaders(trailers);
+        },
+      },
+    );
+    payloads.push(csRes.payload!);
+  } catch (e) {
+    error = ConnectError.from(e);
+    resHeaders = convertToProtoHeaders(error.metadata);
+    resTrailers = convertToProtoHeaders(error.metadata);
```
timostamm, line 168: There reference client does not use error metadata at all for streaming: https://github.com/connectrpc/conformance/blob/34f2af36604771b02138b70db3f0f4502b8d957b/internal/app/referenceclient/impl.go#L157-L162

It's good that we provide combined headers _and_ trailers in error metadata, but it doesn't feel right to populate both `ClientResponseResult.response_headers` and `ClientResponseResult.response_trailers` with them.

If our clients don't invoke the `onHeader` and `onTrailer` callbacks on error, I think we should change them to do so.

srikrsna-buf: With interceptors, I am not sure how we can do that. 

The way I see it is that the promise clients are just one flavor of API, we also have the callback flavor, but should we be testing them as part of conformance? We can just as easily make a client that is close to how requests and responses are done in go and use that for conformance or better yet just use the transport directly.

packages/connect-node-conformance/src/invoke.ts
```diff
@@ -0,0 +1,288 @@
...
+      onTrailer(trailers) {
+        resTrailers = convertToProtoHeaders(trailers);
+      },
+    })) {
+      payloads.push(msg.payload!);
+    }
+  } catch (e) {
+    error = ConnectError.from(e);
+    resHeaders = convertToProtoHeaders(error.metadata);
+    resTrailers = convertToProtoHeaders(error.metadata);
```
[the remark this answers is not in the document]
smaye81: This was intentional, yeah and is related to what you mentioned -- that the API provides no way to distinguish between headers and trailers in the error metadata. I do see your point though that we only set it on the `responseHeaders`. This also seems related to the below question. @jhump what are your thoughts here?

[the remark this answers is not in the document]
smaye81: Gotcha, so the tl;dr would be that just setting them into headers for now is fine until we make this more sound in the future. Sound good?

packages/connect-node-conformance/src/invoke.ts
timostamm: My bad, I didn't have the `Transport` types correctly in mind. We can't invoke onTrailer callbacks on error. We should not start differentiating in `ConnectError`. 

It still doesn't feel right to populate both `ClientResponseResult.response_headers` and `ClientResponseResult.response_trailers` with combined metadata. We need at least an explanatory comment.

srikrsna-buf: Added the comment and also switched to only adding the header, if not already populated.

packages/connect-web-test/src/gen/connectrpc/conformance/v1/messages_pb.ts
```diff
@@ -30,7 +30,7 @@
 
 // Message definitions to be used by integration test service definitions.
 
-// @generated by protoc-gen-es v1.5.0 with parameter "ts_nocheck=false,target=ts"
+// @generated by protoc-gen-es v1.4.2 with parameter "ts_nocheck=false,target=ts"
```
timostamm, line 33: You need to merge in main again or rebase, re-generate code an re-format with the updated prettier, otherwise, this will likely fail CI when merged to main. 

I guess this will also make the diff here much smaller, and make it easier to follow the actual change.

timostamm (commented):
The code looks good to me, thanks for adding the comments! But please see comment below.

packages/connect-web-test/src/gen/connectrpc/conformance/v1/messages_pb.ts
srikrsna-buf: It was not the merge, took me some time to figure out. The npm version resolution logic kept installing `1.4.2` on this branch but is `1.5.0` in main. It also added the integrity values this time around. Should be good now

timostamm (approved):

PR #891, opened 2023-10-27
Update Node versions tested

smaye81:
Updates the Node versions used in testing:

* Removes Node 19 from tests as this has reached EOL and is not an LTS release.
* Adds Node 21 to the list of tested Node versions
* Updates Node 20 to its latest release as of 10/27 (v20.9.0)
* Updates `connect-web-test` to run on Node versions >=18 (Node 16 does not support the Fetch API)

In addition, this also adds the `http1.1` flag to the curl calls to download the Node tarballs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -201,10 +201,12 @@ testnode: $(BIN)/node16 $(BIN)/node18 $(BIN)/node20 $(BIN)/node21 $(BUILD)/conne
 	$(MAKE) conformanceserverstop
 
 .PHONY: testwebnode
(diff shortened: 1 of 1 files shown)


smaye81:
@timo should we update this Node 20? https://github.com/connectrpc/connect-es/pull/891/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R204

timostamm:
Yes, good idea to extend `testwebnode` to run on all versions - except for the CI runtime penalty, but that just means we should improve the setup sooner.

timostamm (approved):

PR #892, opened 2023-10-30
Update nvmrc to v20.9.0

smaye81:
This updates the .nvmrc file to always use the latest minor version of Node v20 (current LTS)


timostamm (approved):

smaye81:
@timostamm wdyt about this instead to always use the latest LTS?

timostamm:
Maybe? I'm not sure that it's supported by all things that read .nvmrc files. Is that standardized anywhere? I'm also a bit worried that it will update the major version on a whim - npm, which is tied to Node.js, is used heavily in build and release.

smaye81:
It looks like nvm and fnm both support it, but your other point about major versions is probably more salient here. Maybe we go with 20.x so that we at least stay updated within a major version?

timostamm:
If `20.x` works, this sounds good to me.

timostamm (approved):

PR #921, opened 2023-11-16
Add explicit exports for node

smaye81:
Related to issue [#610](https://github.com/bufbuild/protobuf-es/issues/610) experienced in Protobuf-ES. This adds explicit exports for `node` into the package.json for all published packages in Connect-ES. In addition, it modifies the path for the top-level `import` statement to use ESM instead of the proxy. 

For more context, check [PR #611](https://github.com/bufbuild/protobuf-es/pull/611) in Protobuf-ES as well as the above issue.

Also note that the ideal setting for the `node` imports condition should be something along the lines of:

```json
  "exports": {
    ".": {
      "node": {
        "import": "./dist/proxy/index.js",
      },
      "module": "./dist/esm/index.js",
      "import": "./dist/esm/index.js",
    }
  },
```
Since any `require` statements for Node _should_ find the top-level `require` condition.

However, doing so causes the `attw` checks to fail with a [Fallback Condition](https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/FallbackCondition.md) error. So, we have added an additional `"require": "./dist/cjs/index.js"` inside the `node` condition to satisfy this check. But, it's unclear whether this is necessary and whether this check is entirely accurate on `attw`'s part. 

The claim is that the `require` condition is being resolved due to a TypeScript bug. The `attw` docs state that Node's resolution algorithm "**requires that resolution stop once a target filename has been tried, whether or not that file could be found**". TypeScript, though, will continue to evaluate conditions if a file cannot be found. Hence, why this check fails - `attw` thinks the `require` condition is only being resolved because of a bug in TypeScript.

This does not seem to be Node's behavior in practice, though. Node [docs](https://nodejs.org/api/packages.html#nested-conditions) clearly state:

> Conditions continue to be matched in order as with flat conditions. If a nested condition does not have any mapping it will continue checking the remaining conditions of the parent condition. In this way nested conditions behave analogously to nested JavaScript if statements.

In addition, this has been verified with some local checks. So, the tl;dr here is that the `attw` **Fallback Condition** check seems like it _may_ be incorrect, but needs deeper investigation. In the meantime, we are adding this additional `node.require` condition to satisfy the check until we can definitively say whether it's needed.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -15,7 +15,7 @@ module.exports = {
     "packages/*/dist/**",
     "node_modules/**",
   ],
-  plugins: ["@typescript-eslint", "node", "import"],
+  plugins: ["@typescript-eslint", "n", "import"],
   // Rules and settings that do not require a non-default parser
   extends: ["eslint:recommended"],
   rules: {
@@ -37,10 +37,10 @@ module.exports = {
           },
           settings: {
             "import/resolver": {
-              "typescript": {
-                "project": "packages/*/tsconfig.json",
-              }
-            }
+              typescript: {
+                project: "packages/*/tsconfig.json",
+              },
+            },
           },
           extends: [
             "plugin:@typescript-eslint/recommended",
@@ -51,7 +51,8 @@ module.exports = {
           rules: {
             "@typescript-eslint/strict-boolean-expressions": "error",
             "@typescript-eslint/no-unnecessary-condition": "error",
-            "@typescript-eslint/array-type": "off", // we use complex typings, where Array is actually more readable than T[]
+            // we use complex typings, where Array is actually more readable than T[]
+            "@typescript-eslint/array-type": "off",
             "@typescript-eslint/switch-exhaustiveness-check": "error",
             "@typescript-eslint/prefer-nullish-coalescing": "error",
             "@typescript-eslint/no-unnecessary-boolean-literal-compare":
@@ -60,7 +61,11 @@ module.exports = {
             "@typescript-eslint/no-base-to-string": "error",
             "import/no-cycle": "error",
             "import/no-duplicates": "error",
-            "import/consistent-type-specifier-style": ["error", "prefer-top-level"], // TS 4.5 adds type modifiers on import names, but we want to support lower versions
+            // TS 4.5 adds type modifiers on import names, but we want to support lower versions
+            "import/consistent-type-specifier-style": [
+              "error",
+              "prefer-top-level",
+            ],
           },
         };
       }),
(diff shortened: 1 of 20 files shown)


packages/connect-express/package.json
```diff
@@ -20,6 +20,10 @@
   "main": "./dist/cjs/index.js",
   "exports": {
     ".": {
+      "node": {
+        "require": "./dist/cjs/index.js",
+        "import": "./dist/proxy/index.js"
+      },
       "module": "./dist/esm/index.js",
       "require": "./dist/cjs/index.js",
       "import": "./dist/proxy/index.js"
```
timostamm, line 29: This doesn't look right. The point of adding the `node` condition is only use the ESM wrapper on Node.js, because it will fail for other ESM consumers. The ESM wrapper simply delegates to CJS, but only Node.js can import CJS from ESM.

You still have the `import` condition pointing to the ESM wrapper. ESM consumers that don't enter the `node` condition will use this, and fail:

```
"import": "./dist/proxy/index.js"
```

See https://github.com/bufbuild/protobuf-es/issues/610#issuecomment-1802152966

timostamm (changes requested):

packages/connect-express/package.json
smaye81: I see. So, the plain `import` condition **not** nested inside the `node` condition should point to `./dist/esm/index.js`?

Note that #610 is also doing the same thing I'm doing here. I'm assuming that should also be changed?

timostamm: > I see. So, the plain import condition **not** nested inside the node condition should point to ./dist/esm/index.js?

Correct. 

> Note that https://github.com/bufbuild/protobuf-es/issues/610#issuecomment-1802152966 is also doing the same thing I'm doing here. I'm assuming that should also be changed?

Do you mean the exports in my comment, or https://github.com/bufbuild/protobuf-es/pull/611?

If possible, it's best to keep changes to a minimum. So make the change here, test it in examples-es, _then_ come back to https://github.com/bufbuild/protobuf-es/pull/611.

smaye81: I mean the actual changes as implemented in bufbuild/protobuf-es#611. They also specify a plain import condition pointing to the ESM wrapper.

smaye81: Updated the paths for the top-level import condition. If this looks good, I'll create the PR for examples-es.

.eslintrc.cjs
```diff
@@ -73,23 +73,23 @@ module.exports = {
     {
       files: ["**/*.{js,mjs,cjs}"],
       parserOptions: {
-          ecmaVersion: 13, // ES2022 - https://eslint.org/docs/latest/use/configure/language-options#specifying-environments
+        ecmaVersion: 13, // ES2022 - https://eslint.org/docs/latest/use/configure/language-options#specifying-environments
```
smaye81, line 76: Not entirely sure why this is prettifying, but CI fails without it.

timostamm: See https://github.com/connectrpc/connect-es/pull/922. Best to merge 922 to main, and merge that in here.

packages/connect-express/package.json
timostamm: Here's what this uses now:

```
      "node": {
        "require": "./dist/cjs/index.js",
        "import": "./dist/proxy/index.js"
      },
      "module": "./dist/esm/index.js",
      "require": "./dist/cjs/index.js",
      "import": "./dist/esm/index.js"
```

Which is not wrong, but also not ideal, for the following reasons:

1. We should be able  to remove `module` completely here. Even from the [webpack docs](https://webpack.js.org/guides/package-exports/#reference-syntax), where this condition originates from, I don't see a reason we should need it as long as we have the `import` condition.
2. Since order matters, and we certainly prefer `import` over `require`, `import` should come first, before `require`.
3. `node.require` seems irrelevant. I would expect a CJS request from Node.js to match the `require` condition. So we should be able to remove this.

Applying these remarks, we end up with:

```
      "node": {
        "import": "./dist/proxy/index.js"
      },
      "import": "./dist/esm/index.js", 
      "require": "./dist/cjs/index.js",
```

This differs from the suggestion in https://github.com/bufbuild/protobuf-es/issues/610#issuecomment-1802152966 to respect ordering, see point 2.

timostamm (approved):
LGTM!

(Please make sure the PR description is up to date)

PR #924, opened 2023-11-22
Always capture header in `ConnecError`

srikrsna-buf:
In server streaming for all protocols and unary for gRPC/gRPC-Web if an error occurs after the headers are received `ConnectError.metadata` doesn't include them. Instead they are only part of the `onHeader` callback. 

We change the behavior to always capture header in `ConnecError`. 

This was caught by the new conformance tests for client in #890.

### DIFF
--- a/packages/connect-web/src/grpc-web-transport.ts
+++ b/packages/connect-web/src/grpc-web-transport.ts
@@ -213,7 +213,7 @@ export function createGrpcWebTransport(
           if (trailer === undefined) {
             throw "missing trailer";
           }
-          validateTrailer(trailer);
+          validateTrailer(trailer, response.headers);
           if (message === undefined) {
             throw "missing message";
           }
@@ -279,7 +279,7 @@ export function createGrpcWebTransport(
             }
             trailerReceived = true;
             const trailer = trailerParse(data);
-            validateTrailer(trailer);
+            validateTrailer(trailer, undefined);
             trailer.forEach((value, key) => trailerTarget.set(key, value));
             continue;
           }
--- a/packages/connect-web/src/grpc-web-transport.ts
+++ b/packages/connect-web/src/grpc-web-transport.ts
@@ -250,6 +250,7 @@ export function createGrpcWebTransport(
         body: ReadableStream<Uint8Array>,
         foundStatus: boolean,
         trailerTarget: Headers,
+        header: Headers,
       ) {
         const reader = createEnvelopeReadableStream(body).getReader();
         if (foundStatus) {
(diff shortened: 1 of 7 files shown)


packages/connect-web/src/grpc-web-transport.ts
```diff
@@ -279,7 +279,7 @@ export function createGrpcWebTransport(
             }
             trailerReceived = true;
             const trailer = trailerParse(data);
-            validateTrailer(trailer);
+            validateTrailer(trailer, undefined);
```
timostamm, line 282: Shouldn't we pass headers here as well?

packages/connect/src/protocol-grpc-web/transport.ts
```diff
@@ -305,7 +305,7 @@ export function createTransport(opt: CommonTransportOptions): Transport {
                       );
                     }
                     trailerReceived = true;
-                    validateTrailer(chunk.value);
+                    validateTrailer(chunk.value, undefined);
```
timostamm, line 308: Same Q as above.

packages/connect/src/protocol-grpc/transport.ts
```diff
@@ -245,7 +245,7 @@ export function createTransport(opt: CommonTransportOptions): Transport {
               async function* (iterable) {
                 yield* iterable;
                 if (!foundStatus) {
-                  validateTrailer(uRes.trailer);
+                  validateTrailer(uRes.trailer, undefined);
```
timostamm, line 248: Same Q as above.

timostamm (commented):
Makes sense to me to do the [same as connect-go](https://github.com/connectrpc/connect-go/blob/45625455d7dc24049956c74e632c7a7060656cef/error.go#L225). Looks like a legit fix to me 👍 

Can you explain the change for users in the PR description?

One question from me, see comments below.

packages/connect-web/src/grpc-web-transport.ts
srikrsna-buf: For streaming requests, the `onHeader` is triggered and we do send the headers there, should we also add them here?

timostamm: The [behavior as documented for connect-go](https://github.com/connectrpc/connect-go/blob/45625455d7dc24049956c74e632c7a7060656cef/error.go#L225) suggests we should add them here.

This may or may not cause problems when we collect information for the conformance runner, and merge headers twice. Not sure what the best way forward is - perhaps we can take a look at the go conformance impl?

The fact that `onHeader` only triggers for streaming in case of an error seems less than ideal.

srikrsna-buf: I think we should follow connect-go and always return headers in metadata. We can tweak the conformance implementation if needed.

timostamm: 👍  I think this may need clarification in the conformance runner. Let's discuss elsewhere, this change is reasonable on it's own.

timostamm (approved):

PR #952, opened 2023-12-01
Add conformance client for connect-web

srikrsna-buf:
Add conformance client for connect-web. Uses [webdriver](https://webdriver.io/) to start a browser session adn run the conformance runner piping the output of the client-under-test back to the webdriver test and using `execute` to run the client on the browser.

Add a local target for chrome and firefox. I'll followup with browserstack config once this approach is finalised.


srikrsna-buf:
@timostamm Added a readme section describing the flow. It doesn't seem to render on GitHub for some reason, but renders on my machine. Will look into it. prettier also seems to have changed which changed the formatting a little bit

timostamm:
> @timostamm Added a readme section describing the flow. It doesn't seem to render on GitHub for some reason, but renders on my machine. Will look into it. 

We shouldn't need a diagram to explain how we run tests, and I'd much prefer a simple setup similar to the the gRPC-Web client tests [here](https://github.com/connectrpc/conformance/blob/main/grpcwebclient/grpcwebclient.ts).

I think it's worth taking another look at webdriver.io's [standalone mode](https://webdriver.io/docs/setuptypes#standalone-mode). It takes over stdin/stdout, preventing us from using stdio with the test runner. But it seems quite odd that it does, and we might be able to provide an upstream fix or use a local workaround. The browser output is useful, but it would be sufficient to dump it into a log file. That would bring us much closer to the simplicity of the gRPC-Web client tests.

It's a possibility that we get it to work, and integration with browserstackbreaking it again. So I'm okay with merging what we have now, and look for a better solution _after_ integration with browserstack.

FYI, puppeteer can [instrument FF](https://pptr.dev/faq) as well, so we do have options.

packages/connect-conformance/package-lock.json
```diff
@@ -0,0 +1,71 @@
+{
+  "name": "@connectrpc/connect-node-conformance",
```
timostamm, line 2: We shouldn't have a lock file here, there should only be a single one at the repo root. This may have been caused by running npm install in this package.

packages/connect-web-bench/README.md
```diff
@@ -10,5 +10,5 @@ it like a web server would usually do.
 
 | code generator | bundle size        | minified               | compressed           |
 |----------------|-------------------:|-----------------------:|---------------------:|
-| connect        | 117,734 b | 51,217 b | 13,697 b |
+| connect        | 122,099 b | 53,623 b | 14,402 b |
```
timostamm, line 13: This could be caused by new version of esbuild. We have to update esbuild separately, so that we have a good understanding of bundle size changes.

We'll update deps tomorrow - hopefully this (and changes from prettier) will automatically resolve.

srikrsna-buf:
Closing in favor of #1019

PR #959, opened 2023-12-20
Export `*TransportOptions` types for `connect-node`

jrschumacher:
Looks like there are some missing types for `connect-node`

### DIFF
--- a/packages/connect-node/src/index.ts
+++ b/packages/connect-node/src/index.ts
@@ -16,10 +16,14 @@
 import "./node-headers-polyfill.js";
 
 export { createGrpcWebTransport } from "./grpc-web-transport.js";
+export type { GrpcWebTransportOptions } from "./grpc-web-transport.js";
 export { createGrpcTransport } from "./grpc-transport.js";
+export type { GrpcTransportOptions } from "./grpc-transport.js";
 export { createConnectTransport } from "./connect-transport.js";
+export type { ConnectTransportOptions } from "./connect-transport.js";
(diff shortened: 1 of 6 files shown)


srikrsna-buf:
Thank you for the PR. 

While it is certainly useful to export the option types, to keep the API surface minimal we decided to keep them un-exported. Are there any issues you ran into because of them not being exported?

jrschumacher:
@srikrsna-buf I understand. IMO, since these are exported as `type` the API surface area will be minimal.

I encountered warnings when I was creating a wrapper around the Connect SDK. You will see that [connect-web](https://github.com/connectrpc/connect-es/blob/main/packages/connect-web/src/index.ts#L17-L18) exports the `*TransportOptions` as types so that developers can use them when creating transports.

packages/connect-node/src/index.ts
```diff
@@ -15,16 +15,16 @@
...
 
 export {
   universalRequestFromNodeRequest,
   universalResponseToNodeResponse,
 } from "./node-universal-handler.js";
 
-export { createNodeHttpClient } from "./node-universal-client.js";
-export { Http2SessionManager } from "./http2-session-manager.js";
+export { createNodeHttpClient, type NodeHttpClientOptions } from "./node-universal-client.js";
+export { Http2SessionManager, type Http2SessionOptions } from "./http2-session-manager.js";
```
srikrsna-buf, line 30: Can you use `export type` to separate the type exports?

srikrsna-buf (approved):

Issue #800, opened 2023-08-31
Remove support for Node.js v16 after EOL


timostamm:
Support for Node.js v16 ends on September 11th, 2023 ([source](https://nodejs.org/en/blog/announcements/nodejs16-eol)).

We have not decided whether to drop support on our side right away. This issue tracks tasks to do once we do drop support:

- remove polyfills ([node-headers-polyfill.ts](https://github.com/connectrpc/connect-es/blob/439e5ff3a49fec22c886aef64eddc3944e3b5702/packages/connect-node/src/node-headers-polyfill.ts), [node16-polyfill-helper.spec.ts](https://github.com/connectrpc/connect-es/blob/439e5ff3a49fec22c886aef64eddc3944e3b5702/packages/connect/src/node16-polyfill-helper.spec.ts))
- remove special casing in H2 session manager ([http2-session-manager.ts](https://github.com/connectrpc/connect-es/blob/439e5ff3a49fec22c886aef64eddc3944e3b5702/packages/connect-node/src/http2-session-manager.ts#L760))
- remove special casing in tests ([node-universal-client.spec.ts](https://github.com/connectrpc/connect-es/blob/439e5ff3a49fec22c886aef64eddc3944e3b5702/packages/connect-node/src/node-universal-client.spec.ts#L44))
- re-add sideEffects: false removed in https://github.com/connectrpc/connect-es/pull/478/files for the headers polyfill

timostamm:
We're removing support for Node.js v16 with v2.

PR #971, opened 2024-01-02
Dependabot updates

srikrsna-buf:
Dependabot updates.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,7 +20,7 @@
         "./packages/connect-conformance"
       ],
       "devDependencies": {
-        "@arethetypeswrong/cli": "^0.13.1",
+        "@arethetypeswrong/cli": "^0.13.5",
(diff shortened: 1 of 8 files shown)


timostamm:
CI fails with:

```
cd packages/connect-node && ... node16 --trace-warnings ../../node_modules/.bin/jasmine --config=jasmine.json
ReferenceError: ReadableStream is not defined
```

This is most likely [undici v6](https://github.com/nodejs/undici/releases/tag/v6.0.0), which [dropped support for Node.js v16](https://github.com/nodejs/undici/pull/2480). If that's the case, we should pin to undici 5.x until we drop support for Node.js v16. Related to https://github.com/connectrpc/connect-es/issues/800.

srikrsna-buf:
Yep, just reverted to `v5`

timostamm (approved):

PR #974, opened 2024-01-03
Override `instanceof` for `ConnectError`

srikrsna-buf:
Override `instanceof` for `ConnectError`. This will help solve the `instanceof` checks failing due to the dual package hazard (i.e. two different versions of connect causing the `instanceof` checks to fail). 

We can also simplify the changes done in #842 and export without the wrappers like we did before.


packages/connect/src/connect-error.spec.ts
```diff
@@ -137,6 +163,14 @@ describe("ConnectError", () => {
       expect(got.cause).toBe(error);
     });
   });
+  describe("instanceof", () => {
+    it("works for the actual ConnectError", () => {
+      expect(new ConnectError("foo")).toBeInstanceOf(ConnectError);
+    });
```
timostamm, line 169: Let's add tests for `null instanceof ConnectError === false`, `undefined instanceof ConnectError === false`, and `"err" instanceof ConnectError === false`

packages/connect/src/connect-error.ts
```diff
@@ -130,6 +130,25 @@ export class ConnectError extends Error {
...
+      v.name === "ConnectError" &&
+      "code" in v &&
+      Object.values(Code).includes(v.code as keyof Code) &&
+      "metadata" in v &&
+      v.metadata instanceof Headers &&
+      "details" in v &&
+      Array.isArray(v.details) &&
+      "rawMessage" in v &&
+      typeof v.rawMessage == "string" &&
+      "cause" in v
```
timostamm, line 148: `instanceof Headers` can be problematic because of polyfills. Let's remove this check.

Because TypeScript enums are lookup objects with numeric _and_ string values, let's just use `typeof v.code == "number"`.

timostamm:
> We can also simplify the changes done in https://github.com/connectrpc/connect-es/pull/842 and export without the wrappers like we did before.

Yes, but we don't need to bring back the individual "types" and "default" exports, as long as the typing files are next to the js files. The exports can simply be:

```
      "module": "./dist/esm/index.js",
      "import": "./dist/esm/index.js",
```

And I believe we can also remove the "module" export. We have to test it with https://github.com/connectrpc/examples-es/pull/1149 / https://github.com/connectrpc/examples-es/pull/1002.

timostamm (approved):
LGTM (provided a CI re-run passes)

PR #984, opened 2024-01-19
Add service and method in grpc-web unary response

minimal1:
Unary response of gRPC-Web missed a service and a method field.
So I add these like connect-transport.ts

### DIFF
--- a/packages/connect-web/src/grpc-web-transport.ts
+++ b/packages/connect-web/src/grpc-web-transport.ts
@@ -219,6 +219,8 @@ export function createGrpcWebTransport(
           }
           return <UnaryResponse<I, O>>{
             stream: false,
+            service,
+            method,
             header: response.headers,
             message,
             trailer,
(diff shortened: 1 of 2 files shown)


packages/connect-web/src/grpc-web-transport.ts
```diff
@@ -219,6 +219,8 @@ export function createGrpcWebTransport(
           }
           return <UnaryResponse<I, O>>{
```
srikrsna-buf, line 220: I think we can remove the cast here
```suggestion
          return {
```

srikrsna-buf (commented):
Thank you! Left a small suggestion

packages/connect-web/src/grpc-web-transport.ts
minimal1: 👍 thx for review

srikrsna-buf (approved):

PR #1013, opened 2024-02-12
Update conformance proto version

srikrsna-buf:
Update conformance proto version to rc2. Some of the client tests (get and cacelation) fail with rc2 runner but pass when run against latest main. So not updating the runner yet.

### DIFF
--- a/packages/connect-conformance/src/node/server.ts
+++ b/packages/connect-conformance/src/node/server.ts
@@ -32,6 +32,7 @@ import { createRegistry } from "@bufbuild/protobuf";
 import {
   BidiStreamRequest,
   ClientStreamRequest,
+  IdempotentUnaryRequest,
   ServerStreamRequest,
   UnaryRequest,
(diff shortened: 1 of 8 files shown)


packages/connect-conformance/src/node/server.ts
```diff
@@ -72,14 +75,24 @@ export function run() {
...
     case HTTPVersion.HTTP_VERSION_1:
       server = req.useTls
         ? https.createServer(serverOptions, adapter)
-        : http.createServer(adapter);
+        : http.createServer(
+            {
+              highWaterMark:
+                req.messageReceiveLimit === 0
+                  ? undefined
+                  : req.messageReceiveLimit,
```
timostamm, line 92: This doesn't seem like a good idea. Can we simply disable the tests until we solve the underlying problem?

srikrsna-buf: Pushed by mistake, removing

timostamm (approved):

Issue #1014, opened 2024-02-12
Remove "node" export condition


timostamm:
In https://github.com/connectrpc/connect-es/pull/974, we made a change that removes a common failure point for the dual package hazard. 

As a result, we can remove the ["node" export condition](https://github.com/connectrpc/connect-es/blob/e3eeb7fd20ff71f7c03327359dfe4987d5749cd9/packages/connect/package.json#L26-L29) from our package.json. The condition was intended to avoid the dual package hazard issue, but causes problems with some bundlers.

Similar issue for Protobuf-ES: https://github.com/bufbuild/protobuf-es/issues/713

PR #1017, opened 2024-02-13
Remove node export condition

smaye81:
Fixes https://github.com/connectrpc/connect-es/issues/1014.

In https://github.com/connectrpc/connect-es/pull/974, we made a change that removes a common failure point for the dual package hazard. As a result, the `node` exports condition for our packages should not be necessary and is actually problematic for some bundlers.


packages/connect-express/package.json
```diff
@@ -20,10 +20,6 @@
   "main": "./dist/cjs/index.js",
   "exports": {
     ".": {
-      "node": {
-        "import": "./dist/proxy/index.js",
-        "require": "./dist/cjs/index.js"
-      },
```
timostamm, line 26: I believe that we also need to remove the "module" condition. Otherwise, we regress on https://github.com/bufbuild/protobuf-es/issues/610#issuecomment-1798866779

I'm not sure whether removing the "module" condition will break other bundlers, so we have to test with https://github.com/connectrpc/examples-es/pull/1149 / https://github.com/connectrpc/examples-es/pull/1002 again 🙄

timostamm (commented):
Do you remember what bundler/framework stumbled over the "node" condition? Can we confirm that this fixes it?

packages/connect-express/package.json
smaye81: OK, I'll test with those branches. Re: the above question -- I can't remember what framework it was. I know connect-playwright had issues with the `node` condition, but other than that and Nuxt, I can't remember another one.

timostamm: To clarify: We _need_ to remove the "module" condition, otherwise we regress on https://github.com/bufbuild/protobuf-es/issues/610#issuecomment-1798866779. If this puts us into a lot of trouble, we have to step back to the drawing board, but right now, let's remove the condition. Can you do that?

timostamm (approved):

PR #1018, opened 2024-02-14
Add conformance tests for cloudflare server side

srikrsna-buf:
Add conformance tests for cloudflare server side.

### DIFF
--- a/.github/workflows/cloudflare.yaml
+++ b/.github/workflows/cloudflare.yaml
@@ -9,7 +9,7 @@ on:
   workflow_dispatch:
 jobs:
   test:
-    # if: github.event_name != 'pull_request' || startsWith(github.event.pull_request.title, 'Release ')
+    if: github.event_name != 'pull_request' || startsWith(github.event.pull_request.title, 'Release ')
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/cloudflare.yaml
+++ b/.github/workflows/cloudflare.yaml
@@ -9,7 +9,7 @@ on:
   workflow_dispatch:
 jobs:
   test:
-    if: github.event_name != 'pull_request' || startsWith(github.event.pull_request.title, 'Release ')
+    # if: github.event_name != 'pull_request' || startsWith(github.event.pull_request.title, 'Release ')
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/cloudflare.yaml
+++ b/.github/workflows/cloudflare.yaml
@@ -1,0 +1,34 @@
+name: Cloudflare Conformance Tests
+on:
+  schedule:
+    - cron: "0 0 * * *"
+  push:
+    tags: ["v*"]
+  pull_request:
+    branches: [main]
+  workflow_dispatch:
+jobs:
+  test:
+    if: github.event_name != 'pull_request' || startsWith(github.event.pull_request.title, 'Release ')
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
+      - uses: bufbuild/buf-setup-action@v1.29.0
+        with:
+          github_token: ${{ github.token }}
+      - name: cache
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.tmp
+            .tmp
+          key: ${{ runner.os }}-connect-web-ci-${{ hashFiles('Makefile') }}
+          restore-keys: |
+            ${{ runner.os }}-connect-web-ci-
+      - name: build
+        run: make build
+      - name: test
+        env:
+          CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+          CLOUDFLARE_WORKERS_SERVER_HOST: ${{ vars.CLOUDFLARE_WORKERS_SERVER_HOST}}
+        run: make testcloudflareconformance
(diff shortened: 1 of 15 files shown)


packages/connect-conformance/bin/conformancecloudflareserver
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env node
+
+const { run } = require("../dist/cjs/cloudflare/server.js");
```
timostamm, line 3: Can you add this to "bin" in packages/connect-conformance/package.json?

packages/connect-conformance/src/cloudflare/compression.ts
```diff
@@ -0,0 +1,81 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { Code, ConnectError } from "@connectrpc/connect";
+import type { Compression } from "@connectrpc/connect/protocol";
+
+export const compressionGzip = createCompression("gzip");
+export const compressionDeflate = {
```
timostamm, line 19: Just for my understanding: I assume we need those for (de-)compressing individual envelope payloads, and rely on automatic full-body decompression by the runtime's fetch implementation for Connect unary?

packages/connect-conformance/src/cloudflare/handler.ts
```diff
@@ -0,0 +1,98 @@
...
+          (await options.notFound?.(req, env, ctx)) ??
+          new Response("Not found", { status: 404 })
+        );
+      }
+      // Cloudflare handles compression for unary requests, so we explicitly set Accept-Encoding to identity.
+      //
+      // Cloudflare doesn't let modify the headers of the incoming request, so we need to
+      // create a new Request object and modify the headers there.
+      req = new Request(req);
+      req.headers.set("Accept-Encoding", "identity");
```
timostamm, line 89: Looks like simple and sufficient temporary workaround 👍

Can you link to the wintercg issue for context?

packages/connect-conformance/package.json
```diff
@@ -9,9 +9,12 @@
...
     "generate": "buf generate buf.build/connectrpc/conformance:v1.0.0-rc2",
     "clean": "rm -rf ./dist/cjs/*",
-    "build": "tsc --project tsconfig.json --module commonjs --outDir ./dist/cjs",
+    "build": "npm run build:cjs && npm run build:esm",
+    "build:cjs": "tsc --project tsconfig.json --module commonjs --outDir ./dist/cjs",
+    "build:esm": "tsc --project tsconfig.json --module ES2015 --outDir ./dist/esm",
     "test:node:server": "./bin/connectconformance --mode server --conf conformance-node.yaml -v ./bin/conformancenodeserver",
-    "test:node:client": "./bin/connectconformance --mode client --conf conformance-node.yaml -v ./bin/conformancenodeclient"
+    "test:node:client": "./bin/connectconformance --mode client --conf conformance-node.yaml -v ./bin/conformancenodeclient",
+    "test:cloudflare:server": "npx wrangler deploy -c wrangler-server.toml && ./bin/connectconformance --mode server --conf conformance-cloudflare-server.yaml -v ./bin/conformancecloudflareserver"
```
timostamm, line 17: Following [server_compat.proto](https://github.com/connectrpc/conformance/blob/main/proto/connectrpc/conformance/v1/server_compat.proto), I would expect `conformancecloudflareserver` to deploy.

This would allow us to honor `message_receive_limit`.

I suppose we don't because deploy time can cause the runner to time out? Are there other reasons?

packages/connect-conformance/src/cloudflare/server.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export function run() {
+  const req = ServerCompatRequest.fromBinary(
+    readFileSync(process.stdin.fd).subarray(4),
+  );
+  // Keep the process alive for the duration of the test.
+  const timeout = setInterval(() => {}, 5000);
+  process.once("SIGTERM", () => {
+    clearInterval(timeout);
+  });
```
timostamm, line 30: Can you add a comment that we do not start the server here, but in the script "test:cloudflare:server" in packages/connect-conformance/package.json, and explain why?

.github/workflows/cloudflare.yaml
```diff
@@ -0,0 +1,34 @@
...
+  schedule:
+    - cron: "0 0 * * *"
+  push:
+    tags: ["v*"]
+  pull_request:
+    branches: [main]
+  workflow_dispatch:
+jobs:
+  test:
+    if: github.event_name != 'pull_request' || startsWith(github.event.pull_request.title, 'Release ')
```
timostamm, line 12: This is great, but we need at least minimal documentation.

packages/connect-conformance/README.md has plenty of space. Can you explain there that Cloudflare conformance tests run daily, and for every release PR, and all other conformance tests run for every commit?

Then we can update .github/RELEASING.md step 4, and clarify "The PR title must start with `Release ` to run conformance tests." (linking to the conformance readme).

timostamm (commented):
Nice! Just a few comments regarding documentation:

packages/connect-conformance/src/cloudflare/compression.ts
srikrsna-buf: Yes, that's correct

packages/connect-conformance/package.json
srikrsna-buf: There could be multiple servers running in parallel

timostamm: With the added comment in packages/connect-conformance/src/cloudflare/server.ts, this works for me 👍

timostamm (approved):
Let's give @smaye81 a chance to take a look as well.

smaye81 (approved):

PR #1019, opened 2024-02-15
Add conformance tests for web

srikrsna-buf:
Add conformance tests for web. Uses `webdriverio` package to spin up local browsers and run tests. Currently tests on Chrome, Firefox and Safari (if available).

### DIFF
--- a/packages/connect-conformance/README.md
+++ b/packages/connect-conformance/README.md
@@ -3,3 +3,16 @@
 This package provides test coverage for connect-es. It implements the conformance service described [here](https://buf.build/connectrpc/conformance).
 
 It uses the [conformance runner](https://github.com/connectrpc/conformance/releases) to run the tests.
+
+## Browser
+
+Browser tests are run on every commit.
+
+### Running tests in a headless browser
+
+Run `make testwebconformance` to run the tests in a headless browser. It runs tests in chrome and firefox, safari if running on a mac. As a special case it also runs the
+tests in node.
+
+### Running tests in a local browser
+
+Run `make testwebconformancelocal` to run the tests in a local browser. This will open a chrome browser and run the tests. If you want to run the tests in a different browser, set the `CONFORMANCE_BROWSER` environment variable.
--- a/packages/connect-conformance/README.md
+++ b/packages/connect-conformance/README.md
@@ -10,7 +10,7 @@ Browser tests are run on every commit.
 
 ### Running tests in a headless browser
 
-Run `make testwebconformance` to run the tests in a headless browser. It runs tests in chrome and firefox, safari if running on a mac. As a special case it also runs the
+Run `make testwebconformance` to run the tests in a headless browser. It runs tests in chrome and firefox, safari if running on a mac. Safari requires users to enable the 'Allow Remote Automation' option in Safari's Develop menu. As a special case it also runs the
 tests in node.
 
 ### Running tests in a local browser
--- a/packages/connect-conformance/bin/conformancewebclient
+++ b/packages/connect-conformance/bin/conformancewebclient
@@ -1,0 +1,5 @@
+#!/usr/bin/env node
+
+const { run } = require("../dist/cjs/web/client.js");
+
+run();
--- a/packages/connect-conformance/src/web/client.ts
+++ b/packages/connect-conformance/src/web/client.ts
@@ -12,22 +12,53 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { ClientCompatResponse } from "../gen/connectrpc/conformance/v1/client_compat_pb.js";
-import { Buffer } from "node:buffer";
 import { remote } from "webdriverio";
 import type { RemoteOptions } from "webdriverio";
 import * as esbuild from "esbuild";
 import { parseArgs } from "node:util";
-import { readSizeDelimitedBuffers } from "../protocol.js";
+import {
+  readSizeDelimitedBuffers,
+  writeSizeDelimitedBuffer,
+} from "../protocol.js";
+import {
+  ClientCompatRequest,
+  ClientCompatResponse,
+  ClientErrorResult,
+} from "../gen/connectrpc/conformance/v1/client_compat_pb.js";
+import invoke from "../invoke.js";
+import { createTransport } from "./transport.js";
 
 const { values: flags } = parseArgs({
   args: process.argv.slice(2),
   options: {
     browser: { type: "string", default: "chrome" },
+    headless: { type: "boolean" },
   },
 });
 
 export async function run() {
+  if (flags.browser !== "node") {
+    await runBrowser();
+  }
+  for await (const next of readSizeDelimitedBuffers(process.stdin)) {
+    const req = ClientCompatRequest.fromBinary(next);
+    const res = new ClientCompatResponse({
+      testName: req.testName,
+    });
+    try {
+      const invokeResult = await invoke(createTransport(req), req);
+      res.result = { case: "response", value: invokeResult };
+    } catch (e) {
+      res.result = {
+        case: "error",
+        value: new ClientErrorResult({ message: (e as Error).message }),
+      };
+    }
+    process.stdout.write(writeSizeDelimitedBuffer(res.toBinary()));
+  }
+}
+
+async function runBrowser() {
   let capabilities: RemoteOptions["capabilities"] = {
     acceptInsecureCerts: true,
   };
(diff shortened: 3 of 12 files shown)


packages/connect-conformance/src/web/client.ts
```diff
@@ -0,0 +1,97 @@
...
+        void window.runTestCase(reqBytes).then(done);
+      },
+      Array.from(next),
+    );
+    const res = ClientCompatResponse.fromBinary(new Uint8Array(invokeResult));
+    const resData = res.toBinary();
+    const resSize = Buffer.alloc(4);
+    resSize.writeUInt32BE(resData.length);
+    process.stdout.write(resSize);
+    process.stdout.write(resData);
```
timostamm, line 82: What do you think about extracting this into a new `function writeSizeDelimitedBuffer(m: Message): void` in packages/connect-conformance/src/protocol.ts? It's a very simple piece of code, but if we're using it 3 or 4 times, it's IMO worth it to make it a function.

packages/connect-conformance/bin/conformancewebclient
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env node
+
+const { run } = require("../dist/cjs/web/client.js");
```
timostamm, line 3: Can you add this to "bin" in packages/connect-conformance/package.json?

packages/connect-conformance/src/web/transport.ts
```diff
@@ -0,0 +1,94 @@
...
+    scheme = "https://";
+  }
+  const baseUrl = `${scheme}${req.host}:${req.port}`;
+
+  switch (req.httpVersion) {
+    case HTTPVersion.HTTP_VERSION_1:
+    case HTTPVersion.HTTP_VERSION_2:
+      break;
+    case HTTPVersion.HTTP_VERSION_3:
+      throw new Error("HTTP/2 and HTTP/3 are not supported");
```
timostamm, line 47: There's a mismatch between error message and code path here. 

I think we should be good to accept all three HTTP protocol versions here. Chrome and FF have supported h3 for a long time. https://caniuse.com/http3 Information about the state in Safari is muddy, but I see requests with h3 in the network inspector in 17.2.1, browsing to random sites.

Judging from the conformance runner output, we only run tests over plaintext HTTP/1.1. Is this because the runner configuration has `supportsH2c: false`?

packages/connect-conformance/src/web/client.ts
```diff
@@ -0,0 +1,97 @@
...
+        browserName: "safari",
+      };
+      break;
+    default:
+      throw new Error(`Unsupported browser: ${flags.browser}`);
+  }
+  const browser = await remote({
+    // webdriverio prints all the logs to stdout, this will interfere with the conformance test output
+    // so we set the log level to silent
+    logLevel: "silent",
```
timostamm, line 66: I'm surprised because that's exactly what @smaye81 and I tried, and it wasn't effective. 

Was there an issue in our I/O functions? Because I see that readSizeDelimitedBuffers has changed over [readReqBuffers](https://github.com/connectrpc/conformance/blob/926b5346de2357ebfe8dceb4ac46f128c8013e5d/testing/grpcwebclient/grpcwebclient.ts#L83C17-L83C31).

One downside of just silencing the log is that it can contain helpful information. If tests fail, it will be very time-consuming to get to the bottom of an issue if the crucial pice of information is in /dev/null. Can you take a look at redirecting the log to a git-ignored file? If there is no simple solution, let's at least add add a TODO.

packages/connect-conformance/src/web/client.ts
```diff
@@ -0,0 +1,97 @@
...
+    acceptInsecureCerts: true,
+  };
+  switch (flags.browser) {
+    case "chrome":
+    case undefined:
+      capabilities = {
+        ...capabilities,
+        browserName: "chrome",
+        "goog:chromeOptions": {
+          args: ["--disable-gpu", "--headless"],
```
timostamm, line 41: Similar to the log output (comment below), it can be necessary to run tests in a non-headless browser with access to the network inspector and browser console. 

For context, here is documentation for running in a full local browser with the old crosstests: https://github.com/connectrpc/connect-es/tree/main/packages/connect-web-test#running-tests-in-a-local-browser

How can we do the same with webdriver?

timostamm (commented):
Nice! 

I wonder:

- We only run 28 tests - is this correct, or are we inadvertently skipping important tests with our config? 
- Debugging browser tests - can we get access to the network inspector and logs?
- Transports from @bufbuild/connect-web on Node.js - it supports fetch, and running the transports on Node is a valid use case that we cover with the make target `testwebnode`. How can we replace this coverage with conformance tests?

packages/connect-conformance/src/web/client.ts
srikrsna-buf: I think I know why it didn't work. There is a simpler package called `webdriver` which also accepts the same options but after reading the source code I realized it doesn't set the log values anywhere. I looked at the wrapper package `webdriverio` which does set the log level. 

I tried redirecting the logs using the `outputDir` option but that only redirects the driver logs and logs everything else to stdout.

Regarding debugging, if it is regarding setting up the browser we can just run the code directly by commenting this line and it'll pause for input right after the browser setup is done. If we remove the headless options, we can use the browser's console and inspect tab to debug.

I'll add a TODO here and also address the headless comment.

packages/connect-conformance/src/web/transport.ts
srikrsna-buf: It is because we are still using rc1. If we upgrade to rc2 there are some Idempotency tests that fail for connect-node but they seem to be a problem of the runner, because using the latest runner we pass all the tests, so I refrained from upgrading until the next release.

Regarding `HTTP/3`, the latest runner actually even fails for `HTTP/1`, because it expects the client to use `HTTP/1`, even when the reference server supports `HTTP/2`. This I think should be changed, I am yet to report this.

packages/connect-conformance/README.md
```diff
@@ -3,3 +3,16 @@
...
 
 It uses the [conformance runner](https://github.com/connectrpc/conformance/releases) to run the tests.
+
+## Browser
+
+Browser tests are run on every commit.
+
+### Running tests in a headless browser
+
+Run `make testwebconformance` to run the tests in a headless browser. It runs tests in chrome and firefox, safari if running on a mac. As a special case it also runs the
```
smaye81, line 13: Might be worth adding the bit about `Allow Remote Automation` here also.

srikrsna-buf:
> * We only run 28 tests - is this correct, or are we inadvertently skipping important tests with our config?

This is because we are still at rc1 and rc2 has a bug which is fixed in the latest runner. The latest runner runs 946 tests with the same config.

> * Debugging browser tests - can we get access to the network inspector and logs?

Yes added a section for running on a local browser, we can get access to both.

> * Transports from @bufbuild/connect-web on Node.js - it supports fetch, and running the transports on Node is a valid use case that we cover with the make target testwebnode. How can we replace this coverage with conformance tests?

Good point! I just included node as a browser option. Seemed straightforward. There's some more code we can share but I'll do that change in the conformance client PR for cloudflare.

smaye81 (approved):

packages/connect-conformance/src/web/transport.ts
timostamm: Okay, but please put a TODO at the H3 case.

timostamm (commented):
Thanks for adding @connectrpc/connect-web tests on Node! We're only running them in v20, but I think it's acceptable.

Regarding `testwebconformancelocal`, it isn't really possible to use the network inspector because the automated browser will close before you get a chance to open it.

It's crude and not ideal, but WDYT about changing the last line of `runBrowser` to:

```
  if (flags.headless === true) {
    await browser.deleteSession();
  } else {
    await browser.executeScript(`document.write("Tests done. You can inspect requests in the network explorer.")`, []);
  }
```

The conformance runner will time out, but it allows to inspect errors and the network in the browser's dev tools.

srikrsna-buf:
Great idea! Also added a flag to open the devtools on startup in local mode.

packages/connect-conformance/src/web/transport.ts
```diff
@@ -0,0 +1,94 @@
...
+    scheme = "https://";
+  }
+  const baseUrl = `${scheme}${req.host}:${req.port}`;
+
+  switch (req.httpVersion) {
+    case HTTPVersion.HTTP_VERSION_1:
+    case HTTPVersion.HTTP_VERSION_2:
+      break;
+    case HTTPVersion.HTTP_VERSION_3:
+      throw new Error("HTTP/3 is not supported");
```
timostamm, line 47: Would definitely add a TODO here, since it raises a lot of questions, see https://github.com/connectrpc/connect-es/pull/1019#discussion_r1491178819

timostamm (approved):

PR #1029, opened 2024-02-23
Remove buf setup action

smaye81:
This removes the usage of `buf-setup-action` in CI as it is currently unneeded.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -14,9 +14,6 @@ jobs:
     steps:
       - name: checkout
         uses: actions/checkout@v4
-      - uses: bufbuild/buf-setup-action@v1.29.0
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -14,9 +14,6 @@ jobs:
     steps:
       - name: checkout
         uses: actions/checkout@v4
-      - uses: bufbuild/buf-setup-action@v1.29.0
```
timostamm, line 17: Is this the only place we use it?

smaye81: Ugh, no. `ag` doesn't search hidden folders (`.github`) by default and my search missed it in `cloudflare.yml`. Fixed.

PR #1040, opened 2024-03-13
Remove old conformance tests

srikrsna-buf:
Remove old conformance tests as all of them are covered by the new conformance test suite. The old logic is also outdated and can cause problems like the failing CI in #1039.

This PR retains the simple browserstack tests, the new conformance tests are yet to be configured to run on browserstack.


packages/connect-web-test/src/cancellation.spec.ts
```diff
@@ -1,116 +0,0 @@
...
-import {
-  Code,
-  ConnectError,
-  createCallbackClient,
-  createPromiseClient,
-} from "@connectrpc/connect";
-import { describeTransports } from "./helpers/conformanceserver.js";
-import { TestService } from "./gen/connectrpc/conformance/v1/test_connect.js";
-
-describe("explicit cancellation with AbortController", function () {
```
timostamm, line 25: Do we test callback clients with the conformance test runner? If not, we need to keep coverage for this.

packages/connect-web-test/src/fetch.spec.ts
```diff
@@ -1,176 +0,0 @@
...
-import {
-  SimpleRequest,
-  SimpleResponse,
-} from "./gen/connectrpc/conformance/v1/messages_pb.js";
-import {
-  createConnectTransport,
-  createGrpcWebTransport,
-} from "@connectrpc/connect-web";
-
-describe("custom fetch", function () {
```
timostamm, line 25: Custom fetch for SSR has some peculiar requirements, and it would be bad if we regress on them by accident. So we need to keep this.

packages/connect-web-test/src/interceptor.spec.ts
```diff
@@ -1,238 +0,0 @@
...
-      };
-    } else {
-      const trailerKeys = listHeaderKeys(res.trailer).join(", ");
-      log.push(`${name} response done with trailers: ${trailerKeys}`);
-    }
-    return res;
-  };
-}
-
-describe("unary interceptors", () => {
```
timostamm, line 96: These tests are not _great_, but it seems important to keep some test coverage for interceptors in a realistic scenario.

packages/connect-node-test/src/transports.spec.ts
```diff
@@ -1,110 +0,0 @@
...
-    const t = createConnectTransport({
-      httpVersion: "2",
-      baseUrl: "https://example.com",
-      sessionManager: sm,
-    });
-    expect(t).toBeDefined();
-  });
-});
-
-describe("using a session manager to open a connection before starting an application", function () {
```
timostamm, line 69: I don't think we test elsewhere that the session manager is wired up correctly from the transport option, so it would be best to keep this.

packages/connect-node-test/src/node-readme.spec.ts
```diff
@@ -1,288 +0,0 @@
...
-  connectNodeAdapter,
-  createGrpcTransport,
-  createGrpcWebTransport,
-  createConnectTransport,
-} from "@connectrpc/connect-node";
-import { createWritableIterable } from "@connectrpc/connect/protocol";
-
-/* eslint-disable @typescript-eslint/require-await */
-
-describe("node readme", function () {
```
timostamm, line 35: Looks like this file grew to be a mixed bag of tests. It started out as test coverage for the examples we give in the package readme, and I think it's a good thing to retain.

If your concern is CI time, I think it would be better to switch to Jest and run tests in parallel. The Karma & Jasmine combo for browserstack will need to be replaced at some point anyways.

packages/connect-node-test/src/transports.spec.ts
srikrsna: I don't see it testing the transport option in any way, it seems to be just testing the `connect` which is tested in the session manager spec.

srikrsna-buf:
Closing, rework in a new PR

PR #1041, opened 2024-03-13
Remove prototype check from ConnectError

smaye81:
This removes the prototype comparison from ConnectError's `hasInstance` implementation. The prototype comparison is susceptible to the same issues that `instanceof` has and could potentially report false negatives due to the dual package hazard.

### DIFF
--- a/packages/connect-web-bench/README.md
+++ b/packages/connect-web-bench/README.md
@@ -10,5 +10,5 @@ it like a web server would usually do.
 
 | code generator | bundle size        | minified               | compressed           |
 |----------------|-------------------:|-----------------------:|---------------------:|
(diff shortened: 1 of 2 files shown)


srikrsna:
But isn't this check just an optimization for the case where the dual package hazard doesn't exist and the prototype is actually the same?

srikrsna-buf:
But isn't this check just an optimization for the case where the dual package hazard doesn't exist and the prototype is actually the same?

smaye81:
> But isn't this check just an optimization for the case where the dual package hazard doesn't exist and the prototype is actually the same?

Actually, that's true. @timostamm I think we can probably leave this in (and also add it to `PluginOptionError`)?

PR #1045, opened 2024-03-23
build(connect): export UniversalHandler and UniversalHandlerOptions

petermetz:
Currently the only way to import these types is to do something like this:

```typescript
import { UniversalHandlerOptions } from "@connectrpc/connect/dist/cjs/protocol/universal-handler";
```

but it would be preferred to be able to import it through the official API surface like this:

```typescript
import type { UniversalHandlerOptions } from "@connectrpc/connect";
```

Signed-off-by: Peter Somogyvari <peter.metz@unarin.com>

### DIFF
--- a/packages/connect/src/index.ts
+++ b/packages/connect/src/index.ts
@@ -56,3 +56,7 @@ export {
 } from "./implementation.js";
 export type { ServiceImplSpec, MethodImplSpec } from "./implementation.js";
 export { createRouterTransport } from "./router-transport.js";
+export {
+  UniversalHandler,
+  UniversalHandlerOptions,
+} from "./protocol/universal-handler.js";


timostamm:
Hey Peter, the package `@connectrpc/connect` exports these types from a subpath:

```ts
import type {UniversalHandler, UniversalHandlerOptions} from "@connectrpc/connect/protocol";

function foo(opt: UniversalHandlerOptions): UniversalHandler {
  //
}
```

petermetz:
> Hey Peter, the package `@connectrpc/connect` exports these types from a subpath:
> 
> ```ts
> import type {UniversalHandler, UniversalHandlerOptions} from "@connectrpc/connect/protocol";
> 
> function foo(opt: UniversalHandlerOptions): UniversalHandler {
>   //
> }
> ```

@timostamm Oh, I totally missed that! /facepalm
Thank you for the info!

PR #1047, opened 2024-03-26
Add context docs for nextjs

paul-sachs:
Adds documentations about how to use `contextValues` with nextjs.

Context can be a little tricky so adding some additional docs to provide some direction.


timostamm:
Thanks for the PR! 

The goal for the package README files is to provide a concise minimum of information necessary to get started, and to link to the documentation and examples for details. 

This isn't set in stone, and if there is a better approach, we can investigate it and make a consistent change. 

However, until we do, the place to add this information would be https://connectrpc.com/docs/node/server-plugins#nextjs (where the contextValues callback is already documented, albeit not with a complete example), or to add an example to the examples-es repository.

PR #1063, opened 2024-04-12
Tweak error codes according to the conformance suite

srikrsna-buf:
Tweak error codes according to the conformance suite. Users should not notice any difference as most of the tweaks are related to errors caused due to mis-implementation of the protocol.

Also update conformance runner to `v1.0.2`. This lets us shed `node-forge` and `asn1js` which we relied to generate a TLS cert.


packages/connect-web-test/src/conformance/unimplemented_server_streaming_method.spec.ts
```diff
@@ -22,7 +22,7 @@ import { TestService } from "../gen/connectrpc/conformance/v1/test_connect.js";
 import { describeTransports } from "../helpers/conformanceserver.js";
 import { Empty } from "@bufbuild/protobuf";
 
-describe("unimplemented_server_streaming_method", function () {
+xdescribe("unimplemented_server_streaming_method", function () {
```
smaye81, line 25: Is this meant to be skipped (along with the callback client one above?)

packages/connect-web/src/connect-transport.ts
```diff
@@ -268,6 +275,12 @@ export function createConnectTransport(
             break;
           }
           const { flags, data } = result.value;
+          if ((flags & compressedFlag) === compressedFlag) {
+            throw new ConnectError(
+              `protocol error: recevied unsupported compressed output`,
```
smaye81, line 280: nit: `received...`. (here and elsewhere in a few places)

packages/connect/src/promise-client.ts
```diff
@@ -163,13 +163,21 @@ export function createClientStreamingFn<
...
     if (!singleMessage) {
       throw new ConnectError(
         "protocol error: missing response message",
-        Code.Internal,
+        Code.Unimplemented,
+      );
+    }
+    if (count > 1) {
+      throw new ConnectError(
+        "protocol error: received extra messages for client steaming method",
```
smaye81, line 179: nit: 'streaming'

smaye81 (commented):
Just a thought, but is it worth splitting up the actual changes to the Connect library into their own PR(s) just to able to reason about the fixes a bit better?

srikrsna-buf:
> Just a thought, but is it worth splitting up the actual changes to the Connect library into their own PR(s) just to able to reason about the fixes a bit better?

Then we'd end up not having tests for that, or we can update the conformance version first by adding a bunch of know failing, and then add the fixes removing the know failing cases.

It didn't seem worth it because the update is just a version bump and deleting some files, and all the fixes are status code related.

packages/connect-web-test/src/conformance/unimplemented_server_streaming_method.spec.ts
srikrsna-buf: Yes, this fails for both

smaye81:
Seems fine to me, but let's wait to see what @timostamm thinks.

smaye81 (approved):

packages/connect-conformance/package.json
```diff
@@ -23,8 +27,6 @@
     "@bufbuild/protobuf": "^1.7.2",
     "@connectrpc/connect": "1.4.0",
     "@connectrpc/connect-node": "1.4.0",
-    "node-forge": "^1.3.1",
-    "asn1js": "^3.0.5",
```
timostamm, line 27: 🎉

timostamm (commented):
If we cut a release, this PR shows up in the changelog, and the title should tell users what we are changing.

The PR title doesn't give any indication that we're changing error codes, and it's not clear what exactly changes from looking at the diff.

We need better communication. For reference, see https://github.com/connectrpc/connect-go/pull/706.

I suggest to change the PR title to be about the change to the code mappings, and only mention the conformance update in the description. You can link to the connectrpc.com PR, same as the connect-go change did. If there are other changes beyond the status code changes in this PR, please list them as well.

timostamm (approved):

PR #1066, opened 2024-04-22
Add conformance tests for clients on Cloudflare workers

srikrsna-buf:
Add conformance tests for clients on Cloudflare workers. Because of the nature of the tests we have to run this on a dedicated runner where we:
- Deploy the cloudflare worker.
- Start the conformance runner in a publicly accessible manner for the worker to call the reference server.
- Forward the client conformance requests to the cloudflare worker which then calls the reference server.

CI run: https://github.com/connectrpc/connect-es/actions/runs/9090317460

PR #1079, opened 2024-05-08
Reorganize conformance tests

smaye81:
This PR implements a reorganization of the conformance tests in the repo and how they are run. 

The `connect-conformance` package previously ran all the conformance tests on its own. Now, the conformance tests have been moved to their respective packages and `connect-conformance` simply houses the common functionality used by all the packages to run conformance tests. Conformance tests have been relocated to:

* `connect-node` - houses and runs both client and server conformance tests for Connect-Node.
* `connect-web` - houses and runs client conformance tests for Connect-Web
* `connect-express` - houses and runs server conformance tests using the `connect-express` plugin.
* `connect-fastify` - houses and runs server conformance tests using the `connect-fastify` plugin.
* `connect-cloudflare` (**new**) - houses and runs server conformance tests for `connect-node` on Cloudflare. This package exists solely for holding Cloudflare conformance tests (at least for now).

The remainder of `connect-conformance` should be no scripts, no tests running. Just a helper package.

Note that the Makefile (intentionally) no longer runs the Node-related conformance tests in all supported versions of Node. A follow-up PR will add a Github action matrix that runs these tests in various Node versions.


packages/connect-conformance/package.json
```diff
@@ -8,12 +8,19 @@
...
+      "require": "./dist/cjs/index.js"
+    }
+  },
   "scripts": {
     "generate": "buf generate buf.build/connectrpc/conformance:v1.0.2",
     "clean": "rm -rf ./dist/cjs/*",
     "build": "npm run build:cjs && npm run build:esm",
-    "build:cjs": "tsc --project tsconfig.json --module commonjs --outDir ./dist/cjs",
-    "build:esm": "tsc --project tsconfig.json --module ES2015 --outDir ./dist/esm",
+    "build:cjs": "tsc --project tsconfig.json --module commonjs --outDir ./dist/cjs --declaration --declarationDir ./dist/cjs",
```
smaye81, line 22: Is it necessary to build CJS in this package?

Makefile
```diff
@@ -222,6 +222,10 @@ testnodeconformance: $(BIN)/node16 $(BIN)/node18 $(BIN)/node20 $(BIN)/node21 $(B
 	cd packages/connect-conformance && PATH="$(abspath $(BIN)):$(PATH)" node18 ./bin/connectconformance --mode client --conf conformance-node.yaml -v ./bin/conformancenodeclient
 	cd packages/connect-conformance && PATH="$(abspath $(BIN)):$(PATH)" node20 ./bin/connectconformance --mode client --conf conformance-node.yaml -v ./bin/conformancenodeclient
 	cd packages/connect-conformance && PATH="$(abspath $(BIN)):$(PATH)" node21 ./bin/connectconformance --mode client --conf conformance-node.yaml -v ./bin/conformancenodeclient
+	# Express Server
```
smaye81, line 225: Do we want to run the express and fastify tests with multiple versions of Node? If so, not quite sure how that works with `tsx`.

packages/connect-conformance/package.json
srikrsna-buf: We do it for the binaries in the `bin` directory.

smaye81: Gotcha, I figured it was something like that. Couldn't the binaries just use `import` statements though (and have `.js` extensions)?

Makefile
smaye81: OK so:
image

And further only supports Node v18 and above. So, if we did want to test with multiple versions, we could only do it with 18, 20, and 21 and we'd then have to get a little more involved and creative with how we install Node into `$(BIN)` (maybe rename `node18` to `node` before the test, etc.)

smaye81:
@srikrsna-buf we can land #1066 now whenever and I can merge main into this.

packages/connect-web/package.json
```diff
@@ -23,6 +28,9 @@
       "require": "./dist/cjs/index.js"
     }
   },
+  "devDependencies": {
+    "webdriverio": "^8.36.1"
+  },
```
timostamm, line 33: This package uses @connectrpc/connect-conformance, right? It needs a dev dependency on it.

Same for all other packages using @connectrpc/connect-conformance.

packages/connect-web/package.json
```diff
@@ -10,9 +10,14 @@
   "scripts": {
     "clean": "rm -rf ./dist/*",
     "build": "npm run build:cjs && npm run build:esm",
-    "build:cjs": "tsc --project tsconfig.json --module commonjs --outDir ./dist/cjs --declaration --declarationDir ./dist/cjs && echo >./dist/cjs/package.json '{\"type\":\"commonjs\"}'",
-    "build:esm": "tsc --project tsconfig.json --module ES2015 --verbatimModuleSyntax --outDir ./dist/esm --declaration --declarationDir ./dist/esm",
-    "attw": "attw --pack"
+    "build:cjs": "tsc --project tsconfig.build.json --module commonjs --outDir ./dist/cjs --declaration --declarationDir ./dist/cjs && echo >./dist/cjs/package.json '{\"type\":\"commonjs\"}'",
+    "build:esm": "tsc --project tsconfig.build.json --module ES2015 --verbatimModuleSyntax --outDir ./dist/esm --declaration --declarationDir ./dist/esm",
+    "attw": "attw --pack",
+    "conformance:client:chrome": "./conformance/bin/connectconformance --mode client --conf ./conformance/conformance-web.yaml -v -- ./conformance/bin/conformanceclient --browser chrome --headless",
```
timostamm, line 16: The package `@connectrpc/connect-conformance` should have a "binary" (package.json `bin` field) for the conformance test runner `connectconformance`. 

This will make the binary available everywhere, and we don't need multiple copies of the file `conformance/bin/connectconformance` everywhere.

The call here can then be:

```suggestion
    "conformance:client:chrome": "connectconformance --mode client --conf ./conformance/conformance-web.yaml -v -- ./conformance/bin/conformanceclient --browser chrome --headless",
```

Same for all other packages using the conformance test runner.

packages/connect-conformance/src/web/transport.ts
```diff
@@ -13,17 +13,17 @@
 // limitations under the License.
 
 import { createRegistry } from "@bufbuild/protobuf";
-import { ClientCompatRequest } from "../gen/connectrpc/conformance/v1/client_compat_pb.js";
+import {
+  createConnectTransport,
+  createGrpcWebTransport,
+} from "@connectrpc/connect-web";
```
timostamm, line 19: We're importing this from `@connectrpc/connect-web`, this is confusing. 

Why don't we simply move this to `packages/connect-web/conformance`?
Same for packages/connect-conformance/src/node/transport.ts - I don't see why we shouldn't  simply move it to `packages/connect-node/conformance`.

packages/connect-conformance/package.json
```diff
@@ -1,43 +1,49 @@
...
   "dependencies": {
     "@bufbuild/protobuf": "^1.7.2",
     "@connectrpc/connect": "1.4.0",
     "@connectrpc/connect-node": "1.4.0",
     "fflate": "^0.8.1",
     "tar-stream": "^3.1.7",
-    "undici": "^5.28.4",
-    "webdriverio": "^8.36.1",
-    "esbuild": "^0.20.0"
+    "undici": "^5.28.4"
```
timostamm, line 42: Above, we have a dependency on `@connectrpc/connect-node`. It should be removed.

packages/connect-conformance/package.json
```diff
@@ -1,43 +1,49 @@
...
+      "require": "./dist/cjs/index.js"
+    },
+    "./node": {
+      "import": "./dist/esm/node/index.js",
+      "require": "./dist/cjs/node/index.js"
+    },
+    "./web": {
+      "import": "./dist/esm/web/index.js",
+      "require": "./dist/cjs/web/index.js"
+    }
```
timostamm, line 17: For a solid separation of concerns, this package should provide the basics for testing, but it should include any of the code under test. Code under test belongs into a `conformance` directory of a package we want to test.

packages/connect-web/package.json
smaye81: Ah, sweet. Yeah, thanks for that. Updated.

packages/connect-conformance/src/web/transport.ts
smaye81: I was just worried if we needed this transport logic in other places as well, but I guess each one is only used in one place.

packages/connect-web/conformance/bin/conformanceclient
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env -S npx tsx
+
+import { run } from "../client.ts";
+
+run();
```
timostamm, line 5: Can't we run the client with `tsx conformance/client.ts` straight away with `connectconformance`? That seems more direct, and we wouldn't need `conformance/bin` anymore.

smaye81: Indeed we can. Updated.

timostamm (commented):
Nice, this is taking shape.

I left some comments regarding documentation and code style below. 

Regarding the test runner `connectconformance`: 
1. Currently, the version is maintained in two places, it would be better to have just a single one.
2. The shim downloads the release on the fly and caches it. This won't work for parallel runs, and we'll use parallel runs with turborepo.

We've solved both problems in protobuf-es [here](https://github.com/bufbuild/protobuf-es/tree/ca27ce6744b3cb70ed760bf913f84a4d4f8e38b4/packages/upstream-protobuf). We should solve them here as well, but I don't think it's necessary to do it in this PR.

Makefile
```diff
@@ -211,35 +211,31 @@ testnode: $(BIN)/node16 $(BIN)/node18 $(BIN)/node20 $(BIN)/node21 $(BUILD)/conne
...
-	cd packages/connect-conformance && PATH="$(abspath $(BIN)):$(PATH)" node20 ./bin/connectconformance --mode server --conf conformance-node.yaml -v ./bin/conformancenodeserver
-	cd packages/connect-conformance && PATH="$(abspath $(BIN)):$(PATH)" node21 ./bin/connectconformance --mode server --conf conformance-node.yaml -v ./bin/conformancenodeserver
-	# Client
-	cd packages/connect-conformance && PATH="$(abspath $(BIN)):$(PATH)" node16 ./bin/connectconformance --mode client --conf conformance-node.yaml -v ./bin/conformancenodeclient
-	cd packages/connect-conformance && PATH="$(abspath $(BIN)):$(PATH)" node18 ./bin/connectconformance --mode client --conf conformance-node.yaml -v ./bin/conformancenodeclient
-	cd packages/connect-conformance && PATH="$(abspath $(BIN)):$(PATH)" node20 ./bin/connectconformance --mode client --conf conformance-node.yaml -v ./bin/conformancenodeclient
-	cd packages/connect-conformance && PATH="$(abspath $(BIN)):$(PATH)" node21 ./bin/connectconformance --mode client --conf conformance-node.yaml -v ./bin/conformancenodeclient
+testnodeconformance: $(BUILD)/connect-conformance $(BUILD)/connect-node $(BUILD)/connect-fastify $(BUILD)/connect-express
```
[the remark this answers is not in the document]
timostamm: We should also (as follow-ups):
- find a solution for test coverage for `createCallbackClient`, 
- solve the `connectconformance` issue described in https://github.com/connectrpc/connect-es/pull/1079#pullrequestreview-2060075386,
- move tests we want to keep from connect-node-test to connect-node (similar for connect-web-test) and delete these packages,
- remove all remaining bits of the old crosstests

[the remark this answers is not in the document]
smaye81: Yep, sounds good. Created https://github.com/connectrpc/connect-es/issues/1080 to track all these (plus the version-in-two-places issue).

packages/connect-cloudflare/conformance/transport.ts
```diff
@@ -13,27 +13,33 @@
...
+} from "@connectrpc/connect-conformance";
 import { createTransport as createConnectTransport } from "@connectrpc/connect/protocol-connect";
 import { createTransport as createGrpcTransport } from "@connectrpc/connect/protocol-grpc";
 import { createTransport as createGrpcWebTransport } from "@connectrpc/connect/protocol-grpc-web";
 import { createFetchClient } from "@connectrpc/connect/protocol";
 import type { Compression } from "@connectrpc/connect/protocol";
 import { compressionDeflate, compressionGzip } from "./compression.js";
 
+/**
+ * Configure a transport for a client from @connectrpc/connect-node under test.
```
timostamm, line 36:
```suggestion
 * Configure a transport for a client running as a Cloudflare Worker under test.
```

timostamm (approved):

PR #1082, opened 2024-05-25
Allow specifying redirect

siosphere:
When using connect-web SDKs in edge workers, such as Vercel/NextJS, and Cloudflare, the redirect option 'error' is not supported:
```
ConnectError: [unknown] Invalid redirect value, must be one of "follow" or "manual" ("error" won't be implemented since it does not make sense at the edge; use "manual" and check the response status code).
```

There currently does not exist a way to overwrite the redirect option to get around this limitation.

This MR adds in the ability to specify redirect as a client option, while keeping the default error

### DIFF
--- a/packages/connect-web/src/connect-transport.ts
+++ b/packages/connect-web/src/connect-transport.ts
@@ -93,6 +93,13 @@ export interface ConnectTransportOptions {
   credentials?: RequestCredentials;
 
   /**
+   * Controls what the fetch client will do with redirects. The default
+   * value is "error". For reference, see
+   * https://fetch.spec.whatwg.org/#concept-request-redirect-mode
+   */
+  redirect ?: RequestRedirect
+
+  /**
    * Interceptors that should be applied to all calls running through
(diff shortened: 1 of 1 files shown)


srikrsna-buf:
Hey! Thank you for the PR. We want to support edge runtimes around the fetch implementation. We are now actively running tests against cloudflare workers. There is a [handler implementation](https://github.com/connectrpc/connect-es/blob/main/packages/connect-cloudflare/conformance/handler.ts) in the tests. That should be good starting point, if you are in immediate need for edge support. We are planning to release companion packages for various edge runtimes, but are waiting on v2 of the package.

kansson:
Hey @srikrsna-buf any update to this problem?

jakubno:
Also interested

ValentaTomas:
Was this fixed in the end with some other PR?

kansson:
> Was this fixed in the end with some other PR?

Had to patch the package myself.

PR #1083, opened 2024-05-29
Add tests for callback client to connect-web conformance tests

smaye81:
This adds logic to `connect-web` for testing Connect's callback client when running client conformance tests.

The basic approach is that a command-line arg is passed to each runnable client.ts file to specify which type of client to use. The client is then created based on this arg. What this also means is that each run of `npm run conformance:client` for these packages will run the conformance tests twice -- once with a promise client and once with a callback client.

Also, as part of this PR, the known-failures have been moved to files just for ease of discovery. As a result, there are two known failing files in `connect-web` now:

* `known-failing-callback-client.txt` - all failing tests when using the callback client across all browsers.
* `known-failing-promise-client-firefox.txt` - all failing tests when using the promise client in Firefox only.


packages/connect-cloudflare/conformance/client-worker.ts
```diff
@@ -49,7 +49,7 @@ export default createWorkerHandler({
           testName: req.testName,
         });
         try {
-          const invokeResult = await invoke(createTransport(req), req);
+          const invokeResult = await promiseInvoke(createTransport(req), req);
```
smaye81, line 52: Not sure if it is worth it to also test the callback client here in the client worker. It will be a bit more difficult to do this plus both types of clients are already being tested via the changes in `client.ts`.

packages/connect-node/conformance/known-failing-client.txt
```diff
@@ -0,0 +1,3 @@
+# When using the callback client, cancellations are calling the end-of-stream callback with an undefined error if a call is cancelled.
```
smaye81, line 1: I know this is documented as intentional, but conformance tests verify that a cancelled error is returned. Does it make sense to:

* Leave the behavior as-is and just mark these as known-failing
* Fix the behavior so that the cancelled error is passed to the `close` callback
* Modify the clients to manually create a cancelled error if the cancelFn is invoked

packages/connect-conformance/src/callback-client.ts
```diff
@@ -0,0 +1,250 @@
...
+      },
+    );
+  });
+}
+
+export function invoke(
+  transport: Transport,
+  req: ClientCompatRequest,
+): Promise<ClientResponseResult> {
+  const client = createCallbackClient(ConformanceService, transport);
```
timostamm, line 236: I forgot to mention this earlier: Can we move the export at the top of the file to make it easier to read? 

Now that we have two versions, I'd prefer if we name the function `invokeWithCallbackClient` / `invokeWithPromiseClient` to be more self-describing.

Same for promise-client.ts

packages/connect-conformance/src/index.ts
```diff
@@ -13,7 +13,8 @@
 // limitations under the License.
 
 export { default as routes } from "./routes.js";
-export { default as invoke } from "./invoke.js";
+export { invoke as promiseInvoke } from "./promise-client.js";
+export { invoke as callbackInvoke } from "./callback-client.js";
```
timostamm, line 17: Let's rename the exports where they are defined, see comment above.

packages/connect-cloudflare/conformance/known-failing-client.txt
```diff
@@ -0,0 +1 @@
+Connect Compressed Error and End-Stream/*/TLS:true/error/compressed
```
timostamm, line 1: Why do we have a new failure here? 🤔

packages/connect-conformance/src/callback-client.ts
```diff
@@ -0,0 +1,250 @@
...
+
+async function unary(
+  client: ConformanceClient,
+  req: ClientCompatRequest,
+  idempotent: boolean = false,
+) {
+  if (req.requestMessages.length !== 1) {
+    throw new Error("Unary method requires exactly one request message");
+  }
+  req.cancel;
```
timostamm, line 48:
```suggestion
```

packages/connect-web/conformance/known-failing-client.txt
```diff
@@ -0,0 +1,5 @@
+# These tests are failing as follows:
+# - in all browsers/environments with the callback client, cancellations are calling the end-of-stream callback with an undefined error if a call is cancelled.
+# - in firefox with the promise client, the error is actual error {code: 2 (unknown), message: "AbortError: The operation was aborted. "} does not match expected code 1 (canceled)
+**/server-stream/cancel-after-zero-responses
+**/server-stream/cancel-after-responses
```
timostamm, line 5: The failure with the callback client is expected, it's the same as in packages/connect-node/conformance/known-failing-client.txt

But why do we have a new failure with Firefox here? 🤔 The code did not change, so I don't understand why this uncovers a new issue with Firefox. Let's isolate this and file a bug.

timostamm (commented):
The callback client was added specifically for React and migration from grpc-web. It should be sufficient to cover the callback client only in conformance tests for connect-web - i.e. remove the callback client from the tests for connect-node and connect-cloudflare. But we have to understand the new failure in connect-cloudflare first.

packages/connect-cloudflare/conformance/client-worker.ts
timostamm: I think this is redundant, see https://github.com/connectrpc/connect-es/pull/1083#pullrequestreview-2093263022

packages/connect-node/conformance/known-failing-client.txt
timostamm: We cannot change the behavior of the callback, that would be a breaking change.

The known-failing list applies to promise client and callback client, right? I'm worried that adding the known-failing for the callback client will shadow regressions in the promise client. This seems like a bad idea.

We either need a failure-list specific to callback-clients, or we have to return a synthetic ConnectError with code `canceled` to the test runner (but obviously only in the very specific case that we cancelled the call and the close callback receives no error). I would prefer a separate failure-list because the code under test should be as simple as possible.

packages/connect-web/conformance/known-failing-client.txt
smaye81: This isn't a new failure here either. I did the same as above -- moved it to a known-failing file rather than inlining it into the package.json command. 

For most of the conformance tests in connect-web, this known-failing file only applies to usage of the callback client. If you see in the package.json, it's only used in the second call for each browser. Firefox is the exception because we see the expected failures with the promise client but also the same tests fail with the callback client in the same fashion as the other browsers. So to summarize:

chrome: only fails with callback client bc of the undefined / swallowed error passed to the end-of-stream callback.
firefox: fails with both clients -- promise is the expected one re: the mismatched error code. callback is the same failure as chrome.
safari: same failure as chrome.

packages/connect-cloudflare/conformance/known-failing-client.txt
timostamm: Ah, I missed that. Good call on adding the file 👍

smaye81:
@timostamm incorporated your feedback re: file names. Lmk what you think.

timostamm (approved):
Thanks for the great cleanup. LGTM!

PR #1102, opened 2024-06-12
Merge connect-web-test into connect-web

smaye81:
This merges the `connect-web-test` package into the `connect-web` package by:

* Removing tests made obsolete by the conformance runner
* Moving tests that are still relevant from `connect-web-test` package into the `connect-web` package
* Deleting the `connect-web-test` package

All testing for `connect-web` (karma, jasmine, browserstack, and conformance) now lives within the `connect-web` package.


packages/connect-web/README.md
```diff
@@ -55,3 +55,37 @@ for a tutorial, or take a look at [our example](https://github.com/connectrpc/co
 Connect plays nice with Vue, Svelte, Remix, Next.js, Angular and many others. Take a look at
 [our examples](https://github.com/connectrpc/examples-es) for various frameworks.
 
+# Tests
```
smaye81, line 58: I actually don't want to put this here, but just did for a discussion. Should we:
1. Omit this test section altogether from READMEs
2. Move some/all of this to `CONTRIBUTING`
3. Keep it here and just slim it down to some bare essentials
4. Keep it here as shown

I'm probably in favor of Step 1

packages/connect-web/src/browserstacktests/README.md
```diff
@@ -0,0 +1,11 @@
+# Connect-Web Browserstack Tests
```
smaye81, line 1: Note these are at `src/browserstacktests` and conformance is at the project root, which feels odd. I kept them here to mimic how it was in `connect-web-test`.

packages/connect-web/README.md
timostamm: We should definitely not have this section in this README because it shows up on npmjs.com.

The make targets `testwebnode`, `testwebbrowser`, and `testwebbrowserlocal` each only run a single file `fetch.spec.ts`. This was useful when we ran the crosstest suite.

It should suffice to run this spec only on Node.js. That means we can delete:
- scripts "karma:serve", "karma"
- file karma.serve.conf.cjs
- targets `testwebbrowser`, `testwebbrowserlocal`

The only remaining related make target is `testwebnode`. Let's rename it to `testconnectwebpackage` for consistency with `testconnectnodepackage`.

The "Tests" section in the README can then be removed.

packages/connect-web/src/browserstacktests/README.md
timostamm: Agree it would be nice. I think it should be possible to move the browserstack directory up by modifying `karma.browserstack.conf.cjs`. I think it's better to try that separately.

packages/connect-web/src/browserstacktests/buf.gen.yaml
```diff
@@ -0,0 +1,9 @@
+version: v1
+plugins:
+  - name: es
+    out: src/browserstacktests/gen
```
timostamm, line 4: How do we generate the eliza service for browserstack in make? 🤔

packages/connect-web/README.md
smaye81: I think the `karma.conf.cjs` file is also unused now so I deleted that too. So now only `karma.browserstack.conf.cjs` remains.

packages/connect-web/src/browserstacktests/buf.gen.yaml
smaye81: See the `karma:browserstack` script in package.json. It calls the `generate` command as part of the run. I wanted to isolate the generation to browserstack-related commands as much as I could.

timostamm: I think it would be least surprising to have a "generate" script similar to other packages, and to have the buf.gen.yaml at the root. 

Then running `make` will show a diff when the generated code changes, and CI will also pick up the diff (the browserstack workflow doesn't do a diffcheck). It's also easier to port to turborepo (using the same patterns everywhere means we can cache tasks). 

All karma* files are the root are only used for browserstack, so it's not an issue to have buf.gen.yaml there as well.

timostamm (approved):

PR #1104, opened 2024-06-13
Build ESM artifacts with Node16 moduleResolution

### LINKED ISSUE #887: Build ESM artifacts with Node16 module resolution
We currently build ESM artifacts with module resolution "Node", which is the Node.js v10 resolution algorithm.

Protobuf-ES had an issue in the artifacts that would have been caught by using "Node16" module resolution:  https://github.com/bufbuild/protobuf-es/issues/586 

To guard against such issues, we should switch to "Node16" module resolution for all packages in this repository.

smaye81:
Fixes #887 

This switches all packages to build ESM artifacts with Node16 module resolution. In doing so, it turns on Node16 for both `moduleResolution` and `module` by default throughout the repo. It also turns on `verbatimModuleSyntax` by default.

For CJS artifacts, the exceptions for `moduleResolution node10` and `verbatimModuleSyntax false` are now specified in the command script.


packages/connect-conformance/package.json
```diff
@@ -2,6 +2,7 @@
   "name": "@connectrpc/connect-conformance",
   "version": "1.4.0",
   "private": true,
+  "type": "module",
```
smaye81, line 5: Switching this to `module` allows us to build ESM artifacts on Node16 module resolution. But, it raises a question of whether we even need to build both here. Should we maybe only build one? And if so, should it be `commonjs` like `protoc-gen-connect-es` and `connect-migrate`?

packages/connect-node/tsconfig.build.json
```diff
@@ -1,5 +1,9 @@
 {
   "files": ["src/index.ts"],
   "include": ["src/**/*.spec.ts", "src/testdata/*.ts"],
-  "extends": "../../tsconfig.base.json"
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
```
smaye81, line 5: `connect-node` is the only package that causes problems on `Node16` module resolution. This is because of the `Long` library which is a transitive dependency through grpc-js.  It's kind of frustrating that one library brought in transitively and that further is only used for extra testing is inhibiting us here. Any thoughts? We could maybe run these extra tests on their own and create a `tsconfig.extra.json` or something. Then we could exclude them in `tsconfig.json` or `tsconfig.build.json`.

smaye81:
@timostamm here's an attempt at Node16 moduleResolution in connect-es. Left a few questions inside. Let me know what you think.

packages/connect-node/tsconfig.build.json
smaye81: Forgot to mention that we could use this workaround in the meantime also? 😬 

https://github.com/dcodeIO/long.js/issues/125#issuecomment-1897638682

packages/connect-conformance/package.json
timostamm: The package is used by both ESM and CommonJS, right? In that case, let's use the same pattern we use elsewhere.

packages/connect-node/tsconfig.build.json
timostamm: I'm okay with deleting packages/connect-node/src/extra. It's rotting here anyway. Maybe we'll make this available as a separate plugin or library. It's easy enough to pick it up from the history.

packages/connect-conformance/package.json
smaye81: `connect-cloudflare` was the only one that was CJS. But I just updated it to be `type: module` so we could probably remove the CJS build here?

timostamm: protoc-gen-connect-es and connect-migrate are just binaries, they don't export code, but connect-conformance does. Let's use the same pattern everywhere where we export code.

smaye81: Works for me.

timostamm (approved):

PR #1107, opened 2024-06-14
Move browserstack tests to project root

smaye81:
This moves the `browserstacktests` directory inside `src/` to the project root to be more consistent with the `conformance` setup.

Note that in addition, this consolidates the Karma files into that directory as well as the `buf.gen.yaml` file. This keeps all browserstack-related artifacts isolated to a single directory.

Finally, this renames the directory to `browserstack` as hopefully that name alone implies 'testing'.


timostamm (approved):
Nice!

Issue #798, opened 2023-08-31
Replace Makefile with turborepo


timostamm:
This repository is using a Makefile because it initially used Go for code generation. 

We have most tools available as npm packages by now, and in [connectrpc/connect-playwright-es](https://github.com/connectrpc/connect-playwright-es), we switched to a new turborepo setup we want to standardize on.

The plan is to switch this repository to the same setup, which should give us a decent speedup in CI and local development, and also make it easier for contributors to get started.

There are two remaining complications:
- Conformance tests are currently run against Docker servers, and we need to make sure they are running during tests.
- We are running some (but not all) test suites on several versions of Node.js. The best way forward is likely to use the [setup-node action](https://github.com/actions/setup-node/) with a matrix in the CI workflow instead.

timostamm:
The two complications are eliminated with https://github.com/connectrpc/connect-es/pull/1103 and https://github.com/connectrpc/connect-es/pull/1110.

There's a (small) new one with https://github.com/connectrpc/connect-es/pull/1079#pullrequestreview-2060075386.

PR #1110, opened 2024-06-17
Parallelize conformance test runs in CI

smaye81:
This parallelizes the various test runs for conformance in CI to minimize the duration of the entire process.

Previously, the repo was basically only running the web and conformance tests in parallel. However, each of these have various runs which were running serially. Now, each conformance suite run is split into its own Github workflow. In addition, Node version matrices were added to the Node-related conformance runs to add further coverage without affecting overall time. This results in the following all running concurrently:

Node tests:
* connect-express conformance (Node versions 18, 20, 21)
* connect-fastify conformance (Node versions 18, 20, 21)
* connect-node conformance (Node versions 18, 20, 21)

Web tests:
* connect-web conformance (Chrome)
* connect-web conformance (Firefox)
* connect-web conformance (Safari)

Note that conformance tests are still run as part of the `all` default goal, so local runs of just `make` are unaffected. However, the conformance tests are no longer a pre-req of the `test` target, so simply running `test` will not run the conformance tests. To do so, run `make testconformance`.

The above enhancements result in an improvement of approximately 8 minutes on CI runs. The overall run duration is tied to the longest individual workflow, which happens to be connect-web conformance in Chrome at 5 minutes. Therefore, the overall run is around 5 minutes, where previously it was 13 minutes or more.


smaye81:
Our setup should be OK. I checked it against webdriver's docs and it looks fine to me. It's a bit variable too. In this latest run, the safari tests seem to have taken a bit longer and chrome's tests were around 3m. So, this may be a function of how long it takes the chrome driver to start? Not entirely sure.

Also updated to further parallelize so now we're around 4 minutes or so. I don't love that proliferation of checks now and how the Makefile has kind of blown up, so if you know of a better way around this, lmk. Just feels kind of unwieldy and hard to maintain as-is.

timostamm:
Thanks for looking into chrome. 

> Also updated to further parallelize so now we're around 4 minutes or so. I don't love that proliferation of checks now and how the Makefile has kind of blown up, so if you know of a better way around this, lmk. Just feels kind of unwieldy and hard to maintain as-is.

I agree. It's a lot of boilerplate to save 1 minute.

IMO we should punt on it parallelize further with https://github.com/connectrpc/connect-es/issues/798.

smaye81:
@timostamm sounds good. I merged in your changes from #1111 and reverted the commits that further parallelized callback vs. promise client and Node clients and servers.

timostamm (approved):
LGTM!

Issue #1108, opened 2024-06-15
RST_STREAM Handling


mpataki:
**Describe the bug**

When a grpc server is abruptly closed while a streaming RPC is open, client may see a RST_STREAM http/2 frame. No grpc level status, trailer, etc. are present in the response.

I'm using a grpc transport and a PromiseClient, which represents a stream as an async iterable. This scenario results in the client terminating the iterable without error, as though the stream was closed normally/successfully.

The [GRPC HTTP/2 protocol](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#errors) states the following:

> RPC runtime implementations should interpret RST_STREAM as immediate full-closure of the stream and should propagate an error up to the calling application layer.

In my particular case, which is using a fairly vanilla Istio/Envoy proxy, the HTTP/2 code returned is NO_ERROR(0). I see no grpc status header. 

I believe the expected client behaviour would be to throw an error from the iterable/stream so the application can handle it (possibly reconnecting, etc).

**To Reproduce**

I'm testing server restarts in a k8s deployment with Istio/Envoy in front of the services. Abruptly terminating pods, Envoy is responding with the RST_STREAM frame.

**Environment:**
- @connectrpc/connect version: `1.4.0`
- @connectrpc/connect-node version: `1.4.0`
- Node.js version: 20.12.2

mpataki:
For some more context, this is how grpcurl handles this case:
```
Response trailers received:
(empty)
Sent 1 request and received 1 response
ERROR:
  Code: Internal
  Message: stream terminated by RST_STREAM with error code: NO_ERROR
```

And this is host postman handles it:
<img width="1112" alt="Pasted image 20240615075230" src="https://github.com/connectrpc/connect-es/assets/2864684/cfbb7eed-1569-4bd2-bfb7-dd695de5f075">

srikrsna-buf:
Hey! Thank you reporting this, we are looking in to this. Node seems to always report a `NO_ERROR` code even when an `RST_STREAM` is not sent from the server.

mpataki:
Ok! Thanks @srikrsna-buf. I'll stay posted. Please let me know if I can help at all

mpataki:
Hey @srikrsna-buf, curious how your initial look at this panned out? Does it seem readily solvable or should we expect to work around this issue for the foreseeable future?

[...]

PR #1113, opened 2024-06-20
Map `NO_ERROR` h2 code to `Code.Internal`

srikrsna-buf:
Map `NO_ERROR` h2 code to `Code.Internal`.

### DIFF
--- a/packages/connect-node/src/node-universal-client.ts
+++ b/packages/connect-node/src/node-universal-client.ts
@@ -317,8 +317,20 @@ function h2Request(
         }
         sentinel.reject(e);
       });
-
+      let receivedData = false;
+      stream.on("response", () => {
+        receivedData = true;
+      })
       stream.on("close", function h2StreamClose() {
+        // Node APIs don't differentiate between a `RST_STREAM` with `NO_ERROR` and 
+        // a regular close.
+        //
+        // To check if this is an error we see if any data was received, if received
+        // this is not an error.
+        if (stream.rstCode === http2.constants.NGHTTP2_NO_ERROR && !receivedData) {
+          sentinel.reject(new ConnectError(`http/2 stream closed with error code NO_ERROR (0x0)`, Code.Internal));
+          return;
+        }
         const err = connectErrorFromH2ResetCode(stream.rstCode);
         if (err) {
           sentinel.reject(err);
--- a/packages/connect-node/src/node-universal-client.ts
+++ b/packages/connect-node/src/node-universal-client.ts
@@ -320,15 +320,23 @@ function h2Request(
       let receivedData = false;
       stream.on("response", () => {
         receivedData = true;
-      })
+      });
       stream.on("close", function h2StreamClose() {
-        // Node APIs don't differentiate between a `RST_STREAM` with `NO_ERROR` and 
+        // Node APIs don't differentiate between a `RST_STREAM` with `NO_ERROR` and
         // a regular close.
         //
         // To check if this is an error we see if any data was received, if received
         // this is not an error.
-        if (stream.rstCode === http2.constants.NGHTTP2_NO_ERROR && !receivedData) {
-          sentinel.reject(new ConnectError(`http/2 stream closed with error code NO_ERROR (0x0)`, Code.Internal));
+        if (
+          stream.rstCode === http2.constants.NGHTTP2_NO_ERROR &&
+          !receivedData
+        ) {
+          sentinel.reject(
+            new ConnectError(
+              `http/2 stream closed with error code NO_ERROR (0x0)`,
+              Code.Internal,
+            ),
+          );
           return;
         }
         const err = connectErrorFromH2ResetCode(stream.rstCode);
--- a/packages/connect-node/src/node-universal-client.spec.ts
+++ b/packages/connect-node/src/node-universal-client.spec.ts
@@ -30,7 +30,7 @@ describe("node http/2 client closing with RST_STREAM with code CANCEL", function
     }),
   );
   it("should send RST_STREAM frame to the server", async function () {
-    return new Promise<void>((resolve) => {
+    await new Promise<void>((resolve) => {
       http2.connect(server.getUrl(), (session: http2.ClientHttp2Session) => {
         const stream = session.request(
           {
@@ -87,7 +87,7 @@ describe("universal node http client", function () {
   });
 
   describe("against a server that closes immediately", function () {
-    describe("over http/2", function () {
+    describe("over http/2 with code CANCEL", function () {
       let serverReceivedRequest = false;
       const server = useNodeServer(() =>
         http2.createServer((request, response) => {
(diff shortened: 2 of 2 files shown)


packages/connect-node/src/node-universal-client.ts
```diff
@@ -317,8 +317,28 @@ function h2Request(
...
+      });
       stream.on("close", function h2StreamClose() {
+        // Node APIs don't differentiate between a `RST_STREAM` with `NO_ERROR` and
+        // a regular close.
+        //
+        // To check if this is an error we see if any data was received, if received
+        // this is not an error.
+        if (
+          stream.rstCode === http2.constants.NGHTTP2_NO_ERROR &&
+          !receivedData
```
jhump, line 332: What happens if we've received data but no status with the gRPC protocol (which requires a "grpc-status" trailer)? Ideally, that would also result in an internal, so that if we receive message data but the stream is aborted before we get the final trailers, we can still report an error.

When we were both looking at this, we spent a good bit of time also looking at the "GOAWAY" case. IIUC, this doesn't address that -- if there is a "GOAWAY" frame with a "NO_ERROR" status, any streams that were created before the frame was received but with a frame ID higher than the server-indicated max ID will just hang until the connection is closed (in which case they get a network error), right? I know that's a separate bug that doesn't need to be addressed here. Just confirming if my understanding is correct and that is indeed a separate bug. I think where we left it is that it needed some more testing to see exactly what the Node HTTP/2 would do. If there is indeed another semi-related issue with that, we should make sure there is an issue to track it.

jhump (approved):
LGTM. Just one question, but nothing blocking for this change.

packages/connect-node/src/node-universal-client.ts
srikrsna-buf: For the GOAWAY case we need testing to confirm the behavior.

> What happens if we've received data but no status with the gRPC protocol (which requires a "grpc-status" trailer)? 

If we already started receiving data will the server send a NO_ERROR? I understand it can do to lame duck in which case it is not an error for the current stream but to indicate no new connections?

jhump: The lame ducking uses GOAWAY frames, which are connection-wide, not stream-specific. I thought this chunk of code was called in the face of a RST_STREAM frame, which is for terminating a single stream/operation, not the whole connection.

A server _could_ send a RST_STREAM frame with NO_ERROR at any time, even if data has already been sent. It may not be common/expected in practice, but I thought that was actually the reported bug -- that it does happen and the client doesn't handle it correctly. I think the whole reason it gets classified as "internal" in the gRPC spec is because  it is unexpected and shouldn't really happen in practice.

There is one "expected" scenario where this happens which shouldn't (hopefully) trigger this code. It should actually be ignored by the client because the client should have already reported a status from the trailers/end of body. It happens if the server ends the response stream (and, in gRPC, sends back trailers with a status), but the client never half-closed the request side of the stream. In such an instance, the RST_STREAM frame tells the client that the stream is complete and to not bother trying to send anything else for the stream, even to end/close the request side.

A Google search suggests that some HTTP/2 proxies may also do this to selectively cancel streams for which the downstream connection (to backend) has been closed or experienced some network failure (though they should really use INTERNAL_ERROR instead; but either way it would get reported to the client as an "internal" RPC error).

So with the current logic, where we have to just _ignore_ the error if data has been sent (since we can't distinguish the framework doing this to signal the stream is done vs. an actual RST_STREAM frame with NO_ERROR), we would hopefully detect that this is an error because we'd observe that the response was not complete (either incomplete/malformed body and/or missing trailers/status).

packages/connect-node/src/node-universal-client.ts
```diff
@@ -317,8 +317,28 @@ function h2Request(
         }
         sentinel.reject(e);
       });
-
+      let receivedData = false;
+      stream.on("response", () => {
+        receivedData = true;
+      });
```
timostamm, line 323: It's sensible to raise an error for a RST_STREAM with NO_ERROR received before response headers. But I don't think it fully addresses https://github.com/connectrpc/connect-es/issues/1108. The issue mentions an open streaming RPC, so the situation is much more likely to occur _after_ response headers have already been received. 

Ideally, the logic would be to raise an error for any RST_STREAM with NO_ERROR, unless we already received a status in headers or trailers.

This is impossible to implement in the universal client, since it doesn't know the status (which could be encoded in the body, or in trailers, depending on the protocol). 

But can't we verify that we receive a trailer with a grpc-status field here? https://github.com/connectrpc/connect-es/blob/v1.4.0/packages/connect/src/protocol-grpc/transport.ts#L248 It looks like we currently only raise an error from the function `validateTrailer` if we receive an error code, not if we're missing the field.

This will be more complex to test (we should also check unary, gRPC-web, and Connect), but I think this should address the issue exhaustively.

packages/connect-node/src/node-universal-client.ts
timostamm: > if there is a "GOAWAY" frame with a "NO_ERROR" status, any streams that were created before the frame was received but with a frame ID higher than the server-indicated max ID will just hang until the connection is closed (in which case they get a network error), right?

To clarify: For a graceful shutdown (GOAWAY frame with NO_ERROR), the client keeps any open streams open until they finish, and opens a new connection for new streams. So any unfinished requests continue as normal, and new requests work as normal from the user's perspective.

But you observed a different behavior with a frame ID that does not include one or more open streams, and the streams in question hang - as in they don't receive or transmit data anymore?

If this is a situation we can expect in regular operation, this does indeed look like a bug. It would be helpful to have a reproducible example.

jhump: > To clarify: For a graceful shutdown (GOAWAY frame with NO_ERROR), the client keeps any open streams open until they finish, and opens a new connection for new streams. So any unfinished requests continue as normal, and new requests work as normal from the user's perspective.

This isn't entirely correct. The GOAWAY frame includes a max stream ID. So any stream created by the client whose ID is greater than that should _not_ be allowed to finish but should instead be retried on a different connection. With the current implementation in connect-es, the max stream ID is not inspected and nothing is done with those streams, so it is possible/likely that they will simply hang until the connection is actually closed. However, this warrants further testing since it is unclear whether the Node HTTP/2 implementation does some amount of handling of GOAWAY transparently 🤷. Seems unlikely, but we should try to repro first. It is likely that fixing this in the connect-es code could require a backwards-incompatible change to this transport interface, so we should verify whether it's a bug before finalizing v2.

jhump: > So any stream created by the client whose ID is greater than that

FWIW, this only happens when there is a race condition. For example, the client creates a stream and sends the headers frame. The client is responsible for assigning stream IDs, so it assigns ID X. The server sends the GOAWAY _before_ processing the headers frame for stream X, so it may indicate a max stream ID of X - 2 (minus two, instead of one, since client-initiated streams are always odd).

packages/connect-node/src/node-universal-client.ts
jhump: Yes, definitely. And we should add conformance test cases to verify this since it's possible that we have similar issues in other client implementations.

Looking at the [specs for error codes](https://github.com/grpc/grpc/blob/master/doc/statuscodes.md), I think this would fall into the category of "Error parsing returned status", which would map to an UNKNOWN code.

srikrsna-buf: grpcurl (as stated in the issue by the user) reports a code internal, grpc-js also uses internal (https://github.com/grpc/grpc-node/issues/2569). I wonder what go does.

srikrsna-buf:
Superseded by #1205

PR #1118, opened 2024-06-26
Call cleanup in the finally of each method response instead of abort

### LINKED ISSUE #1117: Returning or throwing from a method always triggers an abort
**Describe the bug**

Normally returning or throwing from an RPC method triggers the abort event on the given context's `AbortSignal`. That's very unexpected. I'm not sure if that's intentional or a bug. Hopefully it's not intended, I don't think it should abort every time no matter what. As it stands now you have to remember to remove every abort listener you may have added to avoid accidental downstream aborting, which is really tedious.

**To Reproduce**

- @connectrpc/connect-web version: n/a
- @connectrpc/connect-node version: `1.4.0`
- Frontend framework and version: n/a
- Node.js version: `20.12.1`
- Browser and version: n/a

```proto
// test_service.proto

syntax = "proto3";

package repro.v1;

message FooRequest {
  string name = 1;
}

message FooResponse {}

service TestService {
  rpc Foo(FooRequest) returns (FooResponse);
}

```

```js
// main.js
import {
  createPromiseClient,
  createRouterTransport,
} from "@connectrpc/connect";

import { TestService } from "./gen/repro/v1/test_service_connect.js";
import { FooRequest, FooResponse } from "./gen/repro/v1/test_service_pb.js";

const transport = createRouterTransport((router) => {
  router.service(TestService, {
    async foo(req, ctx) {
      ctx.signal.addEventListener("abort", () => {
        console.log("ABORTED");
      });
      if (req.name === "throw") {
        throw new Error("thrown!");
      }
      return new FooResponse();
    },
  });
});

const client = createPromiseClient(TestService, transport);
await client.foo(new FooRequest({ name: "test" }));
try {
  await client.foo(new FooRequest({ name: "throw" }));
} catch {}

```

This will end up printing ABORTED twice, once for the first request when it returns normally and once for the second when it throws

lourd:
Fixes #1117, avoids user code from having to check the given abort signal to determine if an abort was a true abort or not, and avoids the small overhead of having to execute all abort event handlers.

### DIFF
--- a/packages/connect/src/implementation.ts
+++ b/packages/connect/src/implementation.ts
@@ -150,6 +150,7 @@ interface HandlerContextInit {
 }
 
 interface HandlerContextController extends HandlerContext {
+  cleanup(): void;
   abort(reason?: unknown): void;
 }
 
@@ -183,6 +184,7 @@ export function createHandlerContext(
     requestHeader: new Headers(init.requestHeader),
     responseHeader: new Headers(init.responseHeader),
     responseTrailer: new Headers(init.responseTrailer),
+    cleanup: deadline.cleanup,
     abort(reason?: unknown) {
       deadline.cleanup();
       abortController.abort(reason);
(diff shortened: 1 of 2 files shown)


lourd:
@srikrsna-buf @timostamm please let me know if there's anything you need from me for reviewing this!

srikrsna-buf:
Closing this, #1117 will be [addressed in v2](https://github.com/connectrpc/connect-es/issues/1117#issuecomment-2338023438)

PR #1137, opened 2024-07-10
V2: Support removal of dependencies in connect-migrate

timostamm:


packages/connect-migrate/src/lib/migrate-dependencies.ts
```diff
@@ -0,0 +1,156 @@
...
+type DependencyReplacement = {
+  from: { name: string; range: string };
+  to: { name?: string; version: string };
+};
+/**
+ * Remove all dependencies matching name and version range.
+ */
+type DependencyRemoval = {
+  remove: { name: string; range: string };
+};
```
timostamm, line 39: Instead of only DependencyReplacement, we're supporting this type as well. For the existing migration to v0.13.1, nothing changes except symbol names.

srikrsna (approved):

srikrsna-buf (approved):

PR #1141, opened 2024-07-11
V2: Support buf.gen.yaml in connect-migrate

timostamm:
Support removing plugins and updating pinned remote plugin versions in buf.gen.yaml files in the connect-migrate command.


packages/connect-migrate/src/lib/scan.ts
```diff
@@ -59,6 +61,9 @@ export function scan(
       "**/*.cjs",
       "**/*.mjs",
       "**/*.tsx",
+      // buf.gen.yaml files
+      "**/buf.gen.yaml",
+      "**/buf.*.gen.yaml",
```
timostamm, line 66: Could be relaxed to *.gen.yaml 🤔

timostamm: But I think we're fine without for a start.

srikrsna-buf (approved):

PR #1144, opened 2024-07-12
Pass ConnectRouter directly

lynicis:
I added router field into Connect-Fastify plugin options

In my project, I'm using layered architecture so I've handlers layer for each domain, handlers would depend a lot of classes also my some of handlers would be depend each other. Whatever, I need to pass it ConnectRouter directly not in callback function.

### DIFF
--- a/packages/connect-fastify/src/fastify-connect-plugin.ts
+++ b/packages/connect-fastify/src/fastify-connect-plugin.ts
@@ -50,6 +50,11 @@ interface FastifyConnectPluginOptions extends ConnectRouterOptions {
   routes?: (router: ConnectRouter) => void;
 
   /**
+   * If already got a ConnectRouter, pass this field to easily create routes
+   */
+  router?: ConnectRouter;
+
+  /**
    * If set, once `fastify.close` is called, waits for the requests to be finished for the specified duration
    * before aborting them.
    */
@@ -76,7 +81,7 @@ export function fastifyConnectPlugin(
   opts: FastifyConnectPluginOptions,
   done: (err?: Error) => void,
 ) {
-  if (opts.routes === undefined) {
+  if (opts.routes === undefined && opts.router === undefined) {
     done();
     return;
   }
@@ -93,7 +98,7 @@ export function fastifyConnectPlugin(
       done();
     });
   }
-  const router = createConnectRouter(opts);
+  const router = opts.router ?? createConnectRouter(opts);
   opts.routes(router);
 
   const uHandlers = router.handlers;


timostamm:
Thanks for the PR, Emre. Could you provide some more information about the problem this solves? For example a short piece of code that shows how the callback is a showstopper or leads to boilerplate?

Lynicis:
Firstly, thanks for quick reply Timo.

Actually I create my own ConnectRouter and pass it to my all handlers then I pass it same ConnectRouter to plugin. If I didn't function would looks like bad and getting bigger also It's make ConnectRouter more manageable I think so. I wanna abstract my "registerRoutes" methods in handlers, I'm creating that on class construct right now. I know It's not big deal or common case but I think that would be useful and I need it.

If you ask why should i abstract my registerRoutes methods because my project serving HTTP/2 (REST) and ConnectRPC protocols same time and I need to manage manual these routes. Also I don't wanna get into fastify api hell...

```typescript
routes: (router) => {
            const xHandler = new XHandler(...dependencies);
            new yHandler(yHandler, ...otherDependencies);
        },
```

timostamm:
The downside of adding this option is that it still allows to set router options. For example:

```ts
await server.register(fastifyConnectPlugin, {
    router,
    jsonOptions: {
        emitDefaultValues: true,
    }
});
```

The JSON option would not work, because we can't set them on the existing router. So this change would require a bit more work to make sure the API is reasonable. 

We have an example for Fastify here: https://github.com/connectrpc/examples-es/blob/main/fastify/server.ts
It registers regular Fastify routes, Connect routes, and a CORS plugin. 

I have to admit I haven't quite understood the issue yet. Maybe you can modify the example to demonstrate the issue?

Lynicis:
As I mentioned that PR solves that pass ConnectRouter directly not in function. Also I couldn't see that there is a jsonOptions. I want to avoid all route definitions in plugin options.

"routes" field you can't pass ConnectRouter directly because there is a function. You have to define all route definitions in there, for example;

```ts
import { fastifyConnectPlugin } from "@connectrpc/connect-fastify";
import { fastify } from "fastify";

const server = fastify();

server.register(fastifyConnectPlugin, {
        routes: (router) => {
           router.service(XService, XService.methods.xMethod, (req, context) => {});
        },
    });
```

and my PR will be able to pass the manually created ConnectRouter directly by field;
```ts
import { fastifyConnectPlugin } from "@connectrpc/connect-fastify";
import { createConnectRouter } from "@connectrpc/connect";
import { fastify } from "fastify";

const server = fastify();

const router = createConnectRouter();
router.service(XService, XService.methods.xMethod, (req, context) => {});

server.register(fastifyConnectPlugin, {
  router,
});
```

You get all route definitions in the function, also in the plugin logic it creates the ConnectRouter on its own, but my logic makes it open. 
```ts
export function fastifyConnectPlugin(
  instance: FastifyInstance,
  opts: FastifyConnectPluginOptions,
  done: (err?: Error) => void,
) {
  if (opts.routes === undefined) {
    done();
    return;
  }
 
...logic

  const router = createConnectRouter(opts);
  opts.routes(router);

...logic
}
```

timostamm:
Can you explain what issue the first code snippet has that the second one solves?

```ts
// now
server.register(fastifyConnectPlugin, {
  (router) -> 
    router.service(XService, XService.methods.xMethod, (req, context) => {}),
});

// proposed
const router = createConnectRouter();
router.service(XService, XService.methods.xMethod, (req, context) => {});
server.register(fastifyConnectPlugin, {
  router,
});
```

It's literally the same code, just with different indentation 🙂

The callback nudges you to define routes in a single place, which leads to readable and maintainable code. It's still possible to move implementations to separate files with the [helper types](https://connectrpc.com/docs/node/implementing-services#helper-types) if the code base grows.

If you can provide an example for a practical issue, we can think about accepting a router object in the fastify plugin (and not accept router options in this case).

But the first step to make an argument in favor of that would be to demonstrate a shortcoming with the current approach.

PR #1148, opened 2024-07-17
V2: Use base64Encoding options from @bufbuild/protobuf v2

timostamm:
`base64Encode()` from `@bufbuild/protobuf/wire` takes an optional argument to omit padding or to use base64url encoding.

We can use it here to replace several regex replaces.


timostamm:
The CI failure is caused by racy format / generate. This is fixed in https://github.com/connectrpc/connect-es/pull/1149, and will need to be merged into v2.

jchadwick-buf (approved):
Nice, LGTM. Checked the implementation in protobuf-es v2 to double check, LGTM too.

srikrsna-buf (approved):

PR #1156, opened 2024-07-25
Consume body before the connection is closed

Reflejo:
This change consumes the response body in case of an error with a non-json content-type for debugging purposes. Without this, it's not possible to see what the server returned on non 200 http responses. We had multiple instances where we needed this (401s, 500s, etc).

*Note: If the proposed change makes sense, I can add them to other transports*

### DIFF
--- a/packages/connect-web/src/connect-transport.ts
+++ b/packages/connect-web/src/connect-transport.ts
@@ -209,17 +209,28 @@ export function createConnectTransport(
             signal: req.signal,
             body,
           });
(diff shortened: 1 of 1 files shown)


srikrsna-buf:
Hey @Reflejo, apologies for the late review. While I understand the problem, we can't read the response with `res.text`, it expects the body to be valid utf-8. The problem your facing is specific to debugging in a local session, but this change affects all scenarios.

Unfortunately I don't have a good solution. I'm closing this PR, feel free to open an issue to track this.

PR #1157, opened 2024-07-25
Create pregenerate scripts for relevant packages

smaye81:
This creates `pregenerate` scripts for all relevant packages, which does the removal of the generated files directories. In doing so, it moves logic from the Makefile to npm.


smaye81:
@timostamm This is just a partial port of the changes from #1153. Cloudflare and Express do not currently generate any code on `main` and Connect-Node doesn't have a `testdata/` directory, so no need for the `.prettierignore` changes or the `$(GEN)` target changes.

Basically it boiled down to creating `pregenerate` scripts and moving the `rm -rf` logic from the Makefile.

timostamm (approved):
Cool, I didn't realize we don't generate these files on `main` yet 👍 

The "pregenerate" scripts are still useful on their own.

PR #1159, opened 2024-07-29
Upgrade @connect-next to NextJS v14

smaye81:
This upgrades `@connect-next` to add support for Next.js v14 as a peer dependency. Note that v13.x is still supported.

This upgrade was tested in examples-es [here](https://github.com/connectrpc/examples-es/pull/1718) to verify no issues.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,5 +1,5 @@
 {
-  "name": "connect-web",
+  "name": "connect-es",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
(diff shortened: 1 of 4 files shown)


smaye81:
@timostamm actually should we support 13 and 14? i.e. `>=13.2.4 <15`

smaye81:
I found this in the [docs](https://vercel.com/docs/frameworks/nextjs#server-side-rendering-ssr). Not sure if this is what you're looking for:

image

timostamm (approved):
LGTM!

Can you update the PR title and description to clarify this adds support for Next v14, and doesn't drop v13?

PR #1181, opened 2024-08-22
Listens to response close event for aborted request trigger on http 1

### LINKED ISSUE #1025: Requests appear aborted before handler over HTTP/1.1
**Describe the bug**

The signal `request.signal` on a `UniversalServerRequest` is aborted as soon as the request body has been read, at least over HTTP/1.1.  As a result, for most rpc handlers the request appears aborted before the handler begins (since the request has already been read).  Ideally the signal would not be aborted until the underlying connection closes.

**To Reproduce**

Create an rpc handler using node HTTP/1.1, hit the endpoint with a POST request, and read `request.signal.aborted` at the top of the handler.  It will appear aborted (i.e. take the value `true`) even though the client and server maintain a connection, and the client will be able to receive the response.

- @connectrpc/connect-node version: on latest `1.3.0`, likely going back further as well.
- Node.js version: v16+

**Additional context**

The issue most likely goes back to this change in node v16 https://github.com/nodejs/node/issues/40775, where the `'close'` event on an `IncomingMessage` indicates that you've read the whole message, whereas in the past it represented the resource underlying the request (i.e. would not close until the request/response flow ended, either due to actions taken by the server or client).  I believe the more straightforward fix may be to use the node [response object's `'close'` event](https://nodejs.org/api/http.html#event-close_2) rather than the request object's here: https://github.com/connectrpc/connect-es/blob/fa7726b598391b8dfd0e9f14e27dcc87aca74146/packages/connect-node/src/node-universal-handler.ts#L120-L121

lourd:
Fixes #1025.

This breaks some backwards compatibility. Up until now the "aborted" state of the request was always set as soon as the body of the request had been read. This is undesired, buggy behavior, as #1025 documents. For any applications using the abort controller's signal, they would be aborted every time!

The [Node.js documentation on the request close event](https://nodejs.org/api/http.html#event-close) is misleading at best, inaccurate at worst, as discussed in https://github.com/nodejs/node/issues/40775. Nevertheless, listening to the `close` event of the _response_ for aborted client requests is the functionally correct implementation. For example, see New Relic changing their instrumentation library to do so https://github.com/newrelic/node-newrelic/pull/1510

The exported `universalRequestFromNodeRequest` function now requires the `response` parameter as well.

### DIFF
--- a/packages/connect-express/src/express-connect-middleware.ts
+++ b/packages/connect-express/src/express-connect-middleware.ts
@@ -87,6 +87,7 @@ export function expressConnectMiddleware(
     }
     const uReq = universalRequestFromNodeRequest(
       req,
+      res,
       getPreparsedBody(req),
       options.contextValues?.(req),
     );
--- a/packages/connect-fastify/src/fastify-connect-plugin.ts
+++ b/packages/connect-fastify/src/fastify-connect-plugin.ts
@@ -115,6 +115,7 @@ export function fastifyConnectPlugin(
           const uRes = await uHandler(
             universalRequestFromNodeRequest(
               req.raw,
+              reply.raw,
               req.body as JsonValue | undefined,
               opts.contextValues?.(req),
             ),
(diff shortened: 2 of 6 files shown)


srikrsna-buf:
#1025 has been addressed in #1218, the solution is mostly similar to this PR but keeps the current error/api.

lourd:
You're welcome, thanks @srikrsna-buf!

When do y'all plan on publishing the next release?

timostamm:
The next release will be out by the end of the week, or early next week.

PR #1183, opened 2024-08-22
Update conformance testee for @connectrpc/connect-web

timostamm:
Several small changes, see below.

The Firefox known-failing cases are a caused by a bug in the Firefox webdriver: `AbortSignal.abort().reason instanceof Error` is `false` when run through `WebdriverIO.Browser.executeScript`.
We shouldn't need to ignore this. I plan to fix it in a follow-up.

The callback client known-failing cases happen because we swallow the AbortError for both client-side cancel _and_ server-side cancel. It makes sense for the callback-API for client-side cancel, but it would be good to not do this for the server-side.


.github/workflows/web-conformance.yaml
```diff
@@ -57,3 +57,22 @@ jobs:
             ${{ runner.os }}-connect-web-safari-conformance-
       - name: testwebsafariconformance
         run: make testwebsafariconformance
+  node:
+    runs-on: ubuntu-22.04
```
timostamm, line 61: We're running conformance tests for `@connectrpc/connect-web` on Node.js now. They run quick enough, seems worth it.

packages/connect-conformance/src/conformance.ts
```diff
@@ -58,31 +61,28 @@ async function download(url: string, path: string) {
   writeFileSync(path, new Uint8Array(await res.arrayBuffer()));
 }
 
-async function extractBin(path: string) {
-  if (path.endsWith(".zip")) {
-    const unzipped = unzipSync(readFileSync(path), {
+async function extractBin(archivePath: string, binPath: string) {
```
timostamm, line 64: Running the `connectconformance` command in parallel used to fail without an error message, because this script tried to write to the file that was already running.

Makefile
```diff
@@ -205,23 +204,27 @@ testconnectfastifyconformance: $(BUILD)/connect-fastify
...
+	npm run -w packages/connect-web conformance:client:firefox:promise
+	npm run -w packages/connect-web conformance:client:firefox:callback
 
 .PHONY: testwebsafariconformance
 testwebsafariconformance: $(BUILD)/connect-web
-	npm run -w packages/connect-web conformance:client:safari
+	npm run -w packages/connect-web conformance:client:safari:promise
+	npm run -w packages/connect-web conformance:client:safari:callback
 
-.PHONY: testwebconformancelocal
```
timostamm, line 222: This didn't work as documented. Since the npm script runs two commands with `&&`, the `--browser` flag was only appended to the second command. The replacement is explained in packages/connect-conformance/README.md

packages/connect-web/conformance/browserscript.ts
```diff
@@ -42,18 +40,17 @@ async function runTestCase(
...
+      ? await invokeWithCallbackClient(transport, req)
+      : await invokeWithPromiseClient(transport, req);
     res.result = { case: "response", value: result };
   } catch (err) {
     res.result = {
       case: "error",
-      value: new ClientErrorResult({ message: ConnectError.from(err).message }),
+      value: new ClientErrorResult({
+        message: `Failed to run test case: ${String(err)}`,
+      }),
```
timostamm, line 53: Wrapping the error isn't helpful here, this keeps more info.

packages/connect-web/conformance/client.ts
```diff
@@ -68,89 +69,81 @@ async function main() {
...
   }
-  if (flags.headless === true) {
-    await browser.deleteSession();
-  } else {
+  if (flags.openBrowser == true) {
     await browser.executeScript(
-      `document.write("Tests done. You can inspect requests in the network explorer.")`,
+      `const p = document.createElement("p");
+       p.innerText = "Tests done. You can inspect requests in the network explorer."
+       document.body.append(p);`,
```
timostamm, line 136: `document.write` raises an error in Firefox.

packages/connect-web/conformance/client.ts
```diff
@@ -34,11 +33,13 @@ const { values: flags } = parseArgs({
   options: {
     browser: { type: "string", default: "chrome" },
     headless: { type: "boolean" },
+    openBrowser: { type: "boolean" },
```
timostamm, line 36: Instead of opting in to run headless with `--headless`, you now opt in to run with UI with `--openBrowser`. This allows us to re-use the npm scripts. It's explained in packages/connect-conformance/README.md

packages/connect-web/package.json
```diff
@@ -13,11 +13,14 @@
...
-    "conformance:client:node": "connectconformance --mode client --conf ./conformance/conformance-web-node.yaml -v -- ./conformance/client.ts --browser node && connectconformance --mode client --conf ./conformance/conformance-web-node.yaml -v --known-failing @./conformance/known-failing-callback-client.txt -- ./conformance/client.ts --browser node --useCallbackClient",
-    "conformance:client:browser": "connectconformance --mode client --conf ./conformance/conformance-web.yaml -v -- ./conformance/client.ts && connectconformance --mode client --conf ./conformance/conformance-web.yaml -v --known-failing @./conformance/known-failing-callback-client.txt -- ./conformance/client.ts --useCallbackClient",
+    "conformance:client:chrome:promise": "connectconformance --mode client --conf ./conformance/conformance-web.yaml -v -- ./conformance/client.ts --browser chrome",
```
timostamm, line 16: That's an awkward amount of scripts, but it's useful to be able to run just one suite.

smaye81 (approved):

PR #1186, opened 2024-08-26
Fix conformance tests for @connectrpc/connect-web on Firefox

timostamm:
Following up on https://github.com/connectrpc/connect-es/pull/1183, this updates our test setup to avoid an issue with the Firefox webdriver:

Instead of passing our script to FF's JS engine via `WebdriverIO.Browser.executeScript`, we serve a HTML document that embeds the script. This slightly complicates our test setup, but it's well worth it, since there could be similar other issues that we just haven't run into yet.

### DIFF
--- a/packages/connect-web/package.json
+++ b/packages/connect-web/package.json
@@ -15,7 +15,7 @@
     "attw": "attw --pack",
     "conformance:client:chrome:promise": "connectconformance --mode client --conf ./conformance/conformance-web.yaml -v -- ./conformance/client.ts --browser chrome",
     "conformance:client:chrome:callback": "connectconformance --mode client --conf ./conformance/conformance-web.yaml -v --known-failing @./conformance/known-failing-callback-client.txt -- ./conformance/client.ts --browser chrome --useCallbackClient",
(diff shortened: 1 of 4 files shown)


packages/connect-web/package.json
```diff
@@ -15,7 +15,7 @@
     "attw": "attw --pack",
     "conformance:client:chrome:promise": "connectconformance --mode client --conf ./conformance/conformance-web.yaml -v -- ./conformance/client.ts --browser chrome",
     "conformance:client:chrome:callback": "connectconformance --mode client --conf ./conformance/conformance-web.yaml -v --known-failing @./conformance/known-failing-callback-client.txt -- ./conformance/client.ts --browser chrome --useCallbackClient",
-    "conformance:client:firefox:promise": "connectconformance --mode client --conf ./conformance/conformance-web.yaml -v --known-failing @./conformance/known-failing-promise-client-firefox.txt -- ./conformance/client.ts --browser firefox",
+    "conformance:client:firefox:promise": "connectconformance --mode client --conf ./conformance/conformance-web.yaml -v -- ./conformance/client.ts --browser firefox",
```
timostamm, line 18: As a result, we do not need this known-failing list anymore.

smaye81 (approved):

PR #1204, opened 2024-09-02
Fix flaky decompression error code

timostamm:
In the recent versions v20.17.0 and v16.20.2, the zip module from Node.js uses different error codes for brotli decompression. This updates the code to handle both old error codes, and new.


smaye81:
Ha nice! One small thing - the description mentions the same version. I assume the from version is wrong?

smaye81 (approved):

PR #1205, opened 2024-09-03
Throw an error on missing status in gRPC and gRPC-Web transports

### LINKED ISSUE #1108: RST_STREAM Handling

srikrsna-buf:
Throw an error on missing status in gRPC and gRPC-Web transports. 

In node, there is no way to tell if a server terminated with `RST_STREAM` with code `NO_ERROR` or a regular close. This is causing such closures to not be reported. Checking to see if we received the trailers with the status code will result in the same case. This closes #1108.

This change also triggered a minor conformance failure of unsupported header check, so this PR also adds the check to gRPC transport. We continue to omit this for gRPC-Web for [compatibility](https://github.com/connectrpc/connect-es/pull/656) with other gRPC-Web libraries


packages/connect-node/src/transport.spec.ts
```diff
@@ -0,0 +1,107 @@
+// Copyright 2021-2024 The Connect Authors
```
srikrsna-buf, line 1: Couldn't find a good place to put this test, so created a new transport related test file.

packages/connect/src/protocol-grpc/validate-trailer.ts
```diff
@@ -28,4 +31,12 @@ export function validateTrailer(trailer: Headers, header: Headers): void {
     });
     throw err;
   }
+  if (
+    !header.has(headerGrpcStatus) &&
+    !header.has(headerStatusDetailsBin) &&
+    !trailer.has(headerGrpcStatus) &&
+    !trailer.has(headerStatusDetailsBin)
+  ) {
+    throw new ConnectError("protocol error: missing status", Code.Internal);
```
timostamm, line 40: `validateResponse()` returns a `foundStatus: boolean` that only looks for the `grpc-status` field. I don't have a suggestion yet, but we should try to clean this up a bit.

packages/connect-node/src/transport.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+    server: {
+      name: "Server",
+      I: Int32Value,
+      O: StringValue,
+      kind: MethodKind.ServerStreaming,
+    },
+  },
+} as const;
+
+describe("Calls should fail with code internal on RST_STREAM no_error before trailers are received", function () {
```
timostamm, line 37: Could we possibly add this as a custom connectconformance test with the flag `--test-file`?

srikrsna-buf: There is no way to trigger an `RST_STREAM(0x00)` using a test file. 

We can indirectly trigger it by providing a raw http response that ends the response before request is finished, but that also needs code changes in the conformance runner to allow that case. It [drains the requests](https://github.com/connectrpc/conformance/blob/368943e8bfff29a06fda135b8680433516dbcde5/internal/app/referenceserver/raw_response.go#L220-L228) before sending the raw response.

packages/connect/src/protocol-grpc/validate-trailer.ts
timostamm: Discussed offline: We'll look into cleaning this up in v2.

timostamm (approved):

PR #1207, opened 2024-09-03
Update Node.js versions

timostamm:
- Add v22.7.0
- Update v20 to 20.17.0
- Update v18 to 18.20.4
- Update v16 to 16.20.2
- Remove v21 - it's not an LTS and no longer supported


packages/connect-node/src/compression.spec.ts
```diff
@@ -69,20 +69,6 @@ describe("compression", () => {
           );
         }
       });
-      it("should raise internal error on excessive readMaxBytes", async () => {
```
timostamm, line 72: This test ensures that we raise the error code "internal" if the `readMaxBytes` argument passed to `decompress()` is larger than what the decompressor supports.

We [just had to update the implementation](https://github.com/connectrpc/connect-es/pull/1204) because  error codes changed in a minor, and v22 would require us to change it again.

`readMaxBytes` is a configuration option that's verified [elsewhere](https://github.com/connectrpc/connect-es/blob/d42cff57ee0c20bfbef2f279186bdcc235a2e348/packages/connect/src/protocol/limit-io.ts#L40) to be in bounds, so I think it's best we just remove this test.

srikrsna-buf (approved):

PR #1208, opened 2024-09-04
Upgrade conformance tests to v1.0.3

### LINKED ISSUE #1161: V2: gRPC error code's no longer present in thrown ConnectError
**Describe the bug**

When a gRPC server returns/throws an error code, the V2 (alpha.1) version of `@connectrpc/connect` no longer includes the code and error message in the thrown `ConnectError` and instead throws a `ConnectError` with a status code of `Unknown`.

**To Reproduce**

In this reproduction, the gRPC server is an .NET 7 ASP.Net project and lives at `./GrpcGreeter`. It should be cross-platform and the [.NET SDK](https://dotnet.microsoft.com/en-us/download/dotnet/7.0) can build it and run it on most platforms. Start server before starting the Node.js client. There are no changes to the server in the before and after commits.

The Node.js client is at `./GrpcGreeterClientNodeJs`. Run `npm install` and `npm run dev` to launch.

The before commit is with `@connectrpc/connect-node` v1.4.0 and throws the correct error. https://github.com/ngbrown/grpc-testing/tree/6a2dd5cbfa44da1be1c55ef182d2f82493f77e52

The update to `@bufbuild/protobuf` v2.0.0 and `@connectrpc/connect-node` v2.0.0-alpha.1 is the next commit at https://github.com/ngbrown/grpc-testing/commit/baa3ac68675443a96cf25701901ac6e5742443f8.

The incorrect thrown error from v2 is:

```
ConnectError: [unknown] protocol error: missing output message for unary method
    at next (file:///D:/dev/Learning/Grpc/GrpcGreeter/GrpcGreeterClientNodeJs/node_modules/@connectrpc/connect/dist/esm/protocol-grpc/transport.js:113:31)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Object.unary (file:///D:/dev/Learning/Grpc/GrpcGreeter/GrpcGreeterClientNodeJs/node_modules/@connectrpc/connect/dist/esm/protocol-grpc/transport.js:63:20)
    at async Object.sayHello (file:///D:/dev/Learning/Grpc/GrpcGreeter/GrpcGreeterClientNodeJs/node_modules/@connectrpc/connect/dist/esm/promise-client.js:66:26)
    at async main (file:///D:/dev/Learning/Grpc/GrpcGreeter/GrpcGreeterClientNodeJs/dist/index.js:18:9) {
  rawMessage: 'protocol error: missing output message for unary method',
  code: 2,
  metadata: Headers {
    [Symbol(headers list)]: HeadersList {
      cookies: null,
      [Symbol(headers map)]: Map(0) {},
      [Symbol(headers map sorted)]: null
    },
    [Symbol(guard)]: 'none'
  },
  details: [],
  cause: undefined
}
```

The correct thrown error from v1.4 is:

```
ConnectError: [permission_denied] Unable to say hello to 'nobody'
    at findTrailerError (file:///D:/dev/Learning/Grpc/GrpcGreeter/GrpcGreeterClientNodeJs/node_modules/@connectrpc/connect/dist/esm/protocol-grpc/trailer-status.js:94:20)
    at validateResponse (file:///D:/dev/Learning/Grpc/GrpcGreeter/GrpcGreeterClientNodeJs/node_modules/@connectrpc/connect/dist/esm/protocol-grpc/validate-response.js:34:17)
    at validateResponseWithCompression (file:///D:/dev/Learning/Grpc/GrpcGreeter/GrpcGreeterClientNodeJs/node_modules/@connectrpc/connect/dist/esm/protocol-grpc/validate-response.js:52:29)
    at next (file:///D:/dev/Learning/Grpc/GrpcGreeter/GrpcGreeterClientNodeJs/node_modules/@connectrpc/connect/dist/esm/protocol-grpc/transport.js:86:45)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Object.unary (file:///D:/dev/Learning/Grpc/GrpcGreeter/GrpcGreeterClientNodeJs/node_modules/@connectrpc/connect/dist/esm/protocol-grpc/transport.js:62:20)
    at async Object.sayHello (file:///D:/dev/Learning/Grpc/GrpcGreeter/GrpcGreeterClientNodeJs/node_modules/@connectrpc/connect/dist/esm/promise-client.js:67:26)
    at async main (file:///D:/dev/Learning/Grpc/GrpcGreeter/GrpcGreeterClientNodeJs/dist/index.js:18:9) {
  rawMessage: "Unable to say hello to 'nobody'",
  code: 7,
  metadata: HeadersList {
    cookies: null,
    [Symbol(headers map)]: Map(6) {
      'content-length' => [Object],
      'content-type' => [Object],
      'date' => [Object],
      'grpc-message' => [Object],
      'grpc-status' => [Object],
      'server' => [Object]
    },
    [Symbol(headers map sorted)]: null
  },
  details: [],
  cause: undefined
}
```

**Environment:**
- @connectrpc/connect-node version: 2.0.0-alpha.1
- Frontend framework and version: None
- Node.js version: 18.20.4
- Browser and version: None

**Additional context**
I also did a WireShark trace (h2c) of my full application and the gRPC status headers of `grpc-status` and `grpc-message` are present. I'm not super familiar with HTTP/2 protocol, but they appear to be in the main header and not in a trailing header.

srikrsna-buf:
`v1.0.3` of the conformance runner adds cases to cover trailers only responses. The changes to the conformance clients follows the changes done to the reference client. 

This also changes the conformance tests for callback clients to handle cancellation tests. For unary we rely on global error handler to fail the process and streaming we rely on no of responses as a check.

fetch on node is not throwing an error on the response reader, instead it silently ends the stream. This resulted in a different error code for cancellation errors. https://github.com/nodejs/undici/issues/1940


timostamm:
> fetch on node is not throwing an error on the response reader, instead it silently ends the stream.

When a signal passed to fetch is aborted? It would be good to know whether this is expected behavior by the spec, and whether it's a known issue, and is already fixed in one of the more recent Node versions. Can you find out and add a comment in the source?

srikrsna-buf:
@timostamm I couldn't find any issue related to this but was able to pinpoint the timing:

```typescript
async function run() {
    const controller = new AbortController();
    const res = await fetch("https://example.com/", { signal: controller.signal });
    controller.abort();
    const reader = res.body?.getReader();
    console.log(await reader?.read())
}

run();
```

This prints `{ value: undefined, done: true }`.

Running the same code on chrome throws an `AbortError`.

srikrsna-buf:
Sorry, found it: https://github.com/nodejs/undici/issues/1940

packages/connect-conformance/src/callback-client.ts
```diff
@@ -148,19 +174,30 @@ async function serverStream(
...
       (err) => {
+        // Callback clients swallow client triggered cancellations don't report
+        // that as an error.
+        if (
+          err === undefined &&
+          controller.signal.aborted &&
+          abortCount == count
+        ) {
+          error = new ConnectError("operation aborted", Code.Canceled);
+        }
```
timostamm, line 200: I agree with this. It's more precise than a known-failing list.

packages/connect-conformance/src/callback-client.ts
```diff
@@ -184,6 +221,7 @@ async function serverStream(
       },
       {
         headers: reqHeader,
+        signal: controller.signal,
```
timostamm, line 224: It's preferable to cancel with the cancel function that the callback client returns - it's the API we want to test here. Can we continue to use it?

packages/connect-conformance/src/callback-client.ts
```diff
@@ -84,9 +84,34 @@ async function unary(
...
+            error ?? new ConnectError("operation aborted", Code.Canceled),
+          ),
+        }),
+      );
+    });
+    const { afterCloseSendMs } = getCancelTiming(req);
+    if (afterCloseSendMs >= 0) {
+      wait(afterCloseSendMs)
+        .then(() => controller.abort())
+        .catch(() => {});
```
timostamm, line 105:
```suggestion
      void wait(afterCloseSendMs).then(() => controller.abort());
```

It doesn't swallow errors, but tells the compiler you're ignoring rejections intentionally.

timostamm (commented):
Nice, but can you split up the changes into separate PRs? At least the fix for gRPC trailers-only responses needs to be in a separate PR so that changes and fixes are understandable from reading the changelog of a release.

packages/connect-conformance/src/callback-client.ts
srikrsna-buf: If we use `cancel` we can't check for signal.aborted

packages/connect-conformance/src/callback-client.ts
srikrsna-buf: Yes, just an effect of jumping between languages.

srikrsna-buf: Yes, just an effect of jumping between languages.

packages/connect-conformance/src/callback-client.ts
timostamm: Tracking it ourselves is straight-forward: https://github.com/connectrpc/connect-es/pull/1211

packages/connect-web/browserstack/karma-fixup-AbortController.js
```diff
@@ -16,6 +16,16 @@
 
 (function () {
   if (globalThis.AbortController && globalThis.AbortSignal) {
+    // `throwIfAborted` was added much later.
+    if (!globalThis.AbortSignal.prototype.throwIfAborted) {
```
timostamm, line 20: Ah, forgot about that when suggesting `throwIfAborted` for the @connectrcp/connect-web transports. I don't think we should start to require for v1. It's very possible that it requires someone to add additional polyfills for old browsers in our next v1 release.

We call it to work around an issue in the Node.js fetch implementation. The `signal` option was added much earlier to browsers than `throwIfAborted` (compare [init.signal](https://developer.mozilla.org/en-US/docs/Web/API/Window/fetch#browser_compatibility) to [throwIfAborted](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/throwIfAborted#browser_compatibility)). I expect browsers to _not_ have the same issue as Node.js.

Following this assumption, we can update the transports to use:

```ts
if ("throwIfAborted" in signal) {
  // We assume that implementations without `throwIfAborted` (old 
  // browsers) do honor aborted signals on `read`.
  signal.throwIfAborted();
}
```

WDYT?

timostamm (approved):
LGTM!

PR #1212, opened 2024-09-05
Avoid parallel download of conformance runner

timostamm:


packages/connect-conformance/src/conformance.ts
```diff
@@ -29,34 +29,25 @@ import { execFileSync } from "node:child_process";
...
   }
   execFileSync(binPath, process.argv.slice(2), {
     stdio: "inherit",
   });
 }
 
-function existsAndMatchesVersion(bin: string) {
-  return (
-    existsSync(bin) &&
-    execFileSync(bin, ["--version"]).toString().endsWith(version)
```
timostamm, line 56: The exec output ends with a newline, but `version` does not, so this condition is never true.

Could be fixed with `toString().trim()`, but I chose to remove the `exec` call by adding the version to the bin path instead.

srikrsna-buf (approved):

PR #1214, opened 2024-09-05
Allow cloudflare environment variables in CI

timostamm:
turborepo filters environment variables by default. We started using it in https://github.com/connectrpc/connect-es/pull/1194. This change allows variables for the cloudflare tests to be passed.


srikrsna-buf (approved):
I can confirm this works: https://github.com/connectrpc/connect-es/actions/runs/10723527098

PR #1218, opened 2024-09-09
Fix signal in handler always aborted in HTTP/1.1

### LINKED ISSUE #1025: Requests appear aborted before handler over HTTP/1.1

srikrsna-buf:
Fix signal in handler always aborted in HTTP/1.1. Closes #1025.

### DIFF
--- a/packages/connect-node/src/node-universal-handler.ts
+++ b/packages/connect-node/src/node-universal-handler.ts
@@ -66,69 +66,41 @@ export type NodeServerResponse = (
 };
 
 /**
- * Converts a UniversalServerRequest to a Node.js server request.
+ * Converts a Node.js server response to a UniversalServerRequest.
  * This function helps to implement adapters to server frameworks running
  * on Node.js. Please be careful using this function in your own code, as we
  * may have to make changes to it in the future.
+ */
+export function universalRequestFromNodeResponse(
+  nodeResponse: NodeServerResponse,
+  parsedJsonBody: JsonValue | undefined,
+  contextValues: ContextValues | undefined,
+): UniversalServerRequest {
+  return universalRequestFromNodeRequestOrResponse(
+    nodeResponse,
+    parsedJsonBody,
+    contextValues,
+  );
+}
+
+/**
+ * Converts a Node.js server request to a UniversalServerRequest.
+ * This function helps to implement adapters to server frameworks running
+ * on Node.js. Please be careful using this function in your own code, as we
+ * may have to make changes to it in the future.
+ *
+ * @deprecated Please use `universalRequestFromNodeResponse`
  */
 export function universalRequestFromNodeRequest(
   nodeRequest: NodeServerRequest,
   parsedJsonBody: JsonValue | undefined,
   contextValues: ContextValues | undefined,
 ): UniversalServerRequest {
-  const encrypted =
-    "encrypted" in nodeRequest.socket && nodeRequest.socket.encrypted;
-  const protocol = encrypted ? "https" : "http";
-  const authority =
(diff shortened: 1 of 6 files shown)


packages/connect-node/src/node-universal-handler.ts
```diff
@@ -66,69 +66,41 @@ export type NodeServerResponse = (
 };
 
 /**
- * Converts a UniversalServerRequest to a Node.js server request.
+ * Converts a Node.js server response to a UniversalServerRequest.
  * This function helps to implement adapters to server frameworks running
  * on Node.js. Please be careful using this function in your own code, as we
  * may have to make changes to it in the future.
  */
+export function universalRequestFromNodeResponse(
```
timostamm, line 74: There are two relevant exports:
- universalRequestFromNodeRequest
- universalResponseToNodeResponse

IMO the most simple solution is to keep function names, and simply add an argument to universalRequestFromNodeRequest.

We can use an override for BC in v1:

```ts
export function universalRequestFromNodeRequest(
  nodeRequest: NodeServerRequest,
  nodeResponse: NodeServerResponse,
  parsedJsonBody: JsonValue | undefined,
  contextValues: ContextValues | undefined,
): UniversalServerRequest;
/**
 * @deprecated
 */
export function universalRequestFromNodeRequest(
  nodeRequest: NodeServerRequest,
  parsedJsonBody: JsonValue | undefined,
  contextValues: ContextValues | undefined,
): UniversalServerRequest;
export function universalRequestFromNodeRequest(
  nodeRequest: NodeServerRequest,
  ...rest: [  nodeResponse: NodeServerResponse,
    parsedJsonBody: JsonValue | undefined,
    contextValues: ContextValues | undefined,
  ] | [
    parsedJsonBody: JsonValue | undefined,
    contextValues: ContextValues | undefined,
  ],
): UniversalServerRequest {
  const nodeResponse: NodeServerResponse | undefined = rest.length === 3 ? rest[0] : undefined;
  const parsedJsonBody: JsonValue | undefined = (rest.length === 3 ? rest[1] : rest[0]);
  const contextValues: ContextValues | undefined = (rest.length === 3 ? rest[2] : rest[1]);
  // ...
  (nodeResponse ?? nodeRequest).once("close", onH1Close);
```

And then drop the deprecated part in v2.

timostamm (approved):

Issue #1158, opened 2024-07-26
Headers in NodeTransportOptions are not being passed to the request when using createConnectTransport()


dgroux:
**Describe the bug**

Headers set in the `nodeOptions` property when calling `createConnectTransport()` are not being passed to subsequent requests.

**To Reproduce**

Services for my local development environment are routed using (`.localhost`) subdomains, using a
reverse proxy (Traefik), i.e., api.myapp.localhost

Chrome and curl are able to reach these subdomains as they understand that URLs that end in `.localhost` must be changed to 127.0.0.1 and the "Host" header must be set to the domain name.

Unfortunately, the NodeJS http library does not have this functionality.

I can check if the URL host ends in `.localhost` and change it to 127.0.0.1, but I must be able to
change the "Host" header in order for the reverse proxy load balancer to route the request to the correct service.

Using `createConnectTransport()`, which accepts `nodeOptions` and allows me to pass custom headers, it seems as if connect-es ignores, either intentionally, or accidentally overrides these headers, thus I'm not able to route my service:

```
const transport = createConnectTransport({
        baseUrl: "http://127.0.0.1",
        httpVersion: "1.1",
        nodeOptions:{
            headers:{
                Host: "api.myapp.localhost"
            },
        }
    });
```

**Additional context**

The headers are being set on line `140` in ```packages/connect-node/src/node-universal-client.ts``` (`main`)

Which should be changed to:

```
headers: {
    ...httpOptions?.headers,
    ...webHeaderToNodeHeaders(req.header)
},
```

I would make a pull request but I'm having trouble getting this up and running on my machine.

PR #1219, opened 2024-09-09
Respect headers from transport option `nodeOptions.headers`

timostamm:
Fixes https://github.com/connectrpc/connect-es/issues/1158

### DIFF
--- a/packages/connect-node/src/node-universal-header.ts
+++ b/packages/connect-node/src/node-universal-header.ts
@@ -53,45 +53,66 @@ export function nodeHeaderToWebHeader(
 
 /**
  * Convert a fetch API Headers object to a Node.js headers object.
+ *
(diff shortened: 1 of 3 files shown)


packages/connect-node/src/node-universal-header.ts
```diff
@@ -53,45 +53,66 @@ export function nodeHeaderToWebHeader(
...
+    } else {
+      for (const [key, value] of Object.entries<string>(headersInit)) {
+        appendWebHeader(o, key, value);
+      }
     }
   }
   return o;
 }
+
+function appendWebHeader(
```
timostamm, line 104: This logic needed to change a bit, now that we're _joining_ two Node.js headers objects. We append headers, and - for example - need to convert an existing `number` header to a `string` to add a value, since the headers object allows singular numbers, but only string arrays.

srikrsna-buf (approved):

Issue #958, opened 2023-12-13
baseUrl without // leads to HTTP 404 response


timostamm:
**Describe the bug**

Using a `baseUrl` like `https:localhost` (note the omission of `//` after `https:`), leads to an unexpected error `[unimplemented] HTTP 404`.

**To Reproduce**

In the [example](https://github.com/connectrpc/connect-es/tree/main/packages/example), modify `src/client.ts` on line 36 to read:

```ts
  baseUrl: "https:localhost:8443",
```

Run `npm start` for the server and, separately, `npm run client`. The request will fail with the error message `[unimplemented] HTTP 404`.

**Environment (please complete the following information):**
- @connectrpc/connect-node version: 1.2.0
- Node.js version: v20.9.0

**Additional context**
It's unclear why this leads to a HTTP 404. Perhaps the URL is constructed in an unexpected way. It seems that the transport should rather raise an error on the invalid URL instead of issuing an invalid request.

PR #1220, opened 2024-09-09
Fix baseUrl without // mangling request URL on Node.js with HTTP/2

timostamm:
Fixes https://github.com/connectrpc/connect-es/issues/958

### DIFF
--- a/packages/connect-node/src/node-universal-client.ts
+++ b/packages/connect-node/src/node-universal-client.ts
@@ -280,7 +280,7 @@ function h2Request(
   options: Omit<http2.ClientSessionRequestOptions, "signal">,
   onStream: (stream: http2.ClientHttp2Stream) => void,
 ): void {
-  const requestUrl = new URL(url, sm.authority);
+  const requestUrl = new URL(url);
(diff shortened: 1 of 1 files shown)


packages/connect-node/src/node-universal-client.ts
```diff
@@ -280,7 +280,7 @@ function h2Request(
   options: Omit<http2.ClientSessionRequestOptions, "signal">,
   onStream: (stream: http2.ClientHttp2Stream) => void,
 ): void {
-  const requestUrl = new URL(url, sm.authority);
+  const requestUrl = new URL(url);
```
timostamm, line 283: The URL constructor parses `https:localhost:8443`, but not when the `base` argument is given.

The result here used to be `https://localhost:8443/localhost:8443/connectrpc.eliza.v1.ElizaService/Introduce`, and is now the expected `https://localhost:8443/connectrpc.eliza.v1.ElizaService/Introduce`.

IIRC, the authority was only given as a base to potentially support relative URLs, but I don't actually see us supporting them in a meaningful way with this HTTP client.

srikrsna-buf (approved):

PR #1230, opened 2024-09-13
V2: Use `Stream{Request|Response}` types in interceptors for all streaming rpcs

srikrsna-buf:
In v1 the server used more exact types in interceptors, for example `UnaryRequest` for server streaming rpcs. The client always used streaming variants. This is inconsistent and was unintended. 
 
This change unifies the behavior, now all streaming rpcs will use the `StreamRequest` and `StreamResponse` on the server as well. This is breaking change.

### DIFF
--- a/packages/connect/src/interceptor.ts
+++ b/packages/connect/src/interceptor.ts
@@ -140,6 +140,9 @@ export interface StreamResponse<
   readonly message: AsyncIterable<MessageShape<O>>;
 }
 
+/**
+ * @private
+ */
 export interface RequestCommon<I extends DescMessage, O extends DescMessage> {
   /**
    * Metadata related to the service that is being called.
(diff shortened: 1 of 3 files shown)


packages/connect/src/interceptor.ts
```diff
@@ -140,7 +140,7 @@ export interface StreamResponse<
   readonly message: AsyncIterable<MessageShape<O>>;
 }
 
-interface RequestCommon<I extends DescMessage, O extends DescMessage> {
+export interface RequestCommon<I extends DescMessage, O extends DescMessage> {
```
timostamm, line 143: Nice clean up.

Can you mark the two new exports as `@private`? They are not exported in the package, but this makes it unmistakably clear that they shouldn't be.

timostamm (approved):

Issue #810, opened 2023-09-01
Move away from the karma test runner for browser tests


timostamm:
We test web clients on browsers via [browserstack](https://www.browserstack.com/) with the jasmine test framework and karma, which has an integration with browserstack. (See [here](https://github.com/connectrpc/connect-es/tree/main/packages/connect-web-test) for details.)

[Karma is deprecated since April 2023](https://github.com/karma-runner/karma#karma-is-deprecated-and-is-not-accepting-new-features-or-general-bug-fixes), and we should move to an alternative mechanism to ensure compatibility with old browser.

timostamm:
We started using `webdriverio` for conformance tests. It [appears to support browserstack](https://webdriver.io/docs/browserstack-service/), which means that we're potentially able to run the conformance tests against older web browsers. This would be desirable because it gives us better coverage, and removes the need for handwritten tests specific for browserstack.

timostamm:
This ran successfully for me:

```ts
import {remote} from "webdriverio";

const browserstackUsername = process.env.BROWSERSTACK_USERNAME ?? "";
const browserstackAccessKey = process.env.BROWSERSTACK_ACCESS_KEY ?? "";
if (browserstackUsername.length == 0 || browserstackAccessKey.length == 0) {
  throw new Error(
    "The environment variables BROWSERSTACK_USERNAME and BROWSERSTACK_ACCESS_KEY are required to run tests on browserstack.",
  );
}

const browser = await remote({
  user: browserstackUsername,
  key: browserstackAccessKey,
  services: [
    ["browserstack", {
      // browserstackLocal: true
    }]
  ],
  capabilities: [
    {
      browserName: 'Chrome',
      'bstack:options': {
        browserVersion: '120.0',
        os: 'Windows',
        osVersion: '10'
      }
    },
  ],
  // Directory to store all test runner log files (including reporter logs and wdio logs).
  // If not set, all logs are streamed to stdout, which conflicts with the conformance runner I/O.
  // outputDir: new URL("logs", import.meta.url).pathname,
});

await browser.url("https://connectrpc.com");

await browser.deleteSession();
```

timostamm:
> We started using `webdriverio` for conformance tests. It [appears to support browserstack](https://webdriver.io/docs/browserstack-service/), which means that we're potentially able to run the conformance tests against older web browsers. This would be desirable because it gives us better coverage, and removes the need for handwritten tests specific for browserstack.

[...]

PR #1232, opened 2024-09-13
Switch Browserstack tests to Webdriver

smaye81:
This switches our usage of Karma's test runner to Webdriver since Karma's runner is now [deprecated](https://github.com/karma-runner/karma?tab=readme-ov-file#karma-is-deprecated-and-is-not-accepting-new-features-or-general-bug-fixes).

Closes https://github.com/connectrpc/connect-es/issues/810


.github/workflows/browserstack.yaml
```diff
@@ -36,4 +36,4 @@ jobs:
         env:
           BROWSERSTACK_USERNAME: ${{ secrets.BROWSERSTACK_USERNAME }}
           BROWSERSTACK_ACCESS_KEY: ${{ secrets.BROWSERSTACK_ACCESS_KEY }}
-        run: npx turbo run test-browserstack --output-logs new-only --log-order stream
+        run: npx turbo run test-browserstack
```
timostamm, line 39: We should keep the flags. For long-running tasks such as this, it can be very useful to have the unbuffered logs.

timostamm (commented):
Awesome! Just one comment, see below.

timostamm (approved):

PR #1234, opened 2024-09-16
V2: Don't trigger handler signal on success

srikrsna-buf:
Don't trigger handler signal on success. Implements the behavior highlighted in https://github.com/connectrpc/connect-es/issues/1117#issuecomment-2352384065.

This is a breaking change. In v1 the signal will always trigger on completion.

### DIFF
--- a/packages/connect/src/protocol-connect/handler-factory.ts
+++ b/packages/connect/src/protocol-connect/handler-factory.ts
@@ -258,6 +258,7 @@ function createUnaryHandler<I extends DescMessage, O extends DescMessage>(
       );
       body = serialization.getO(type.binary).serialize(output);
     } catch (e) {
+      context.abort(e);
       let error: ConnectError | undefined;
       if (e instanceof ConnectError) {
         error = e;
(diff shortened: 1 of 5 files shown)


packages/connect/src/protocol-connect/handler-factory.ts
```diff
@@ -462,7 +462,7 @@ function createStreamHandler<I extends DescMessage, O extends DescMessage>(
       },
       transformSerializeEnvelope(serialization.getO(type.binary)),
       transformCatchFinally<EnvelopedMessage>((e) => {
-        context.abort();
+        context.abort(e);
```
timostamm, line 465: The connect handler has a separate code path for unary. For consistency, we have to call `context.abort(e)` in the catch block on line 261.

timostamm (commented):
Can we update the JSDoc on HandlerContext? E.g.

```ts
  /**
   * An AbortSignal that triggers when the deadline is reached, or when an
   * error occurs that aborts processing of the request.
   *
   * The signal can be used to automatically cancel downstream calls.
   */
  readonly signal: AbortSignal;
```

timostamm (approved):

PR #1235, opened 2024-09-17
Deprecate `createPromiseClient` in favor of `createClient`

srikrsna-buf:
Deprecate `createPromiseClient` in favor of `createClient`. Promises are more widely adopted than ever before, this change will mark the promise variant as the default.

This is effectively just a rename, but to keep backward compatibility we keep the existing signature and mark it as deprecated. `createPromiseClient` will be removed in v2.


timostamm (commented):
Looks good to me. Can we add a migration so that `npx @connectrpc/connect-migrate` will update imports and calls to use the renamed function and type?

timostamm (approved):

PR #1244, opened 2024-09-20
Update migration versions

timostamm:


packages/connect-migrate/src/migrations/v0.13.1.ts
```diff
@@ -26,8 +26,8 @@ import { migrateSourceFiles } from "../lib/migrate-source-files";
 import { migratePackages } from "../lib/migrate-packages";
 import { migrateLockFiles } from "../lib/migrate-lock-files";
 
-export const targetVersionConnectEs = "1.1.2";
-export const targetVersionConnectQuery = "0.5.1";
+export const targetVersionConnectEs = "1.6.0";
+export const targetVersionConnectQuery = "1.4.2";
```
timostamm, line 30: Not strictly necessary to bump the connect-es version, but it doesn't hurt.
We should definitely update to the stable version of connect-query-es.

srikrsna-buf (approved):

PR #1252, opened 2024-09-28
add details about underlying message types on input message protocol errors

tmm1:
in general the protocol errors are all missing a lot of context

### DIFF
--- a/packages/connect/src/protocol/invoke-implementation.ts
+++ b/packages/connect/src/protocol/invoke-implementation.ts
@@ -151,6 +151,14 @@ export function transformInvokeImplementation<
           throw new ConnectError(
             "protocol error: received extra input message for unary method",
             Code.Unimplemented,
+            {
+              methodName: spec.method.name,
+              serviceTypeName: spec.service.typeName,
+              inputTypeName: spec.method.I.typeName,
+              outputTypeName: spec.method.O.typeName,
+              url: context.url,
+              requestMethod: context.requestMethod,
+            }
           );
         }
       };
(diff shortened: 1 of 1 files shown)


srikrsna-buf:
Hey @tmm1 Can you tell us a bit about how often you encountered protocol errors and what the backend was? Because we don't expect users to ever hit these errors.

tmm1:
We are seeing large numbers of these proto errors and are trying to figure out why. Adding these details here would help.

packages/connect/src/protocol/invoke-implementation.ts
```diff
@@ -149,8 +149,16 @@ export function transformInvokeImplementation<
...
+            `protocol error: received extra input message ${spec.method.I.typeName} for unary method ${spec.method.name}`,
             Code.Unimplemented,
+            {
+              methodName: spec.method.name,
+              serviceTypeName: spec.service.typeName,
+              inputTypeName: spec.method.I.typeName,
+              outputTypeName: spec.method.O.typeName,
+              url: context.url,
+              requestMethod: context.requestMethod,
+            }
```
srikrsna-buf, line 161: Metadata is used for errors returned as part of the protocol, fully qualified method name should be sufficient to pin point the rpc

packages/connect/src/protocol/invoke-implementation.ts
```diff
@@ -149,8 +149,16 @@ export function transformInvokeImplementation<
         const input2 = await inputIt.next();
         if (input2.done !== true) {
           throw new ConnectError(
-            "protocol error: received extra input message for unary method",
+            `protocol error: received extra input message ${spec.method.I.typeName} for unary method ${spec.method.name}`,
```
srikrsna-buf, line 152: Fully qualified service name is unique and will be sufficient to pinpoint:
```suggestion
            `protocol error: received extra input message for unary method ${spec.service.typeName}.${spec.method.name}`,
```

srikrsna-buf:
Hey @tmm1 Can you tell us what is the client in use?

packages/connect/src/protocol/invoke-implementation.ts
tmm1: If I understand correctly, you are suggesting I remove this as adding metadata would interfere with other parts of the stack?

I can do that. I agree that the rpc method is the main piece of information missing.

tmm1:
We're using typescript on both sides.

tmm1:
Thanks for the guidance. I believe this may be an application specific issue, so I will close for now.

PR #1255, opened 2024-10-03
V2: Replace MethodInfo with Descriptor Types

srikrsna-buf:
Replace MethodInfo with Descriptor Types. We plan on adding these types to `@bufbuild/protobuf`. We can remove the types once that is done.

This also makes the types stricter, transport will now only accept unary descriptors for `unary` and streaming descriptors for `stream` methods. Likewise the interceptor types are also stricter.

### DIFF
--- a/packages/connect/src/protocol/invoke-implementation.ts
+++ b/packages/connect/src/protocol/invoke-implementation.ts
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import type { DescMessage, MessageShape } from "@bufbuild/protobuf";
+import type { DescMessage, DescMethod, MessageShape } from "@bufbuild/protobuf";
 import { ConnectError } from "../connect-error.js";
 import { Code } from "../code.js";
 import type { HandlerContext, MethodImplSpec } from "../implementation.js";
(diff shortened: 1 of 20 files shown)


packages/connect/src/protocol/invoke-implementation.ts
```diff
@@ -111,6 +114,7 @@ export function transformInvokeImplementation<
             return {
               stream: true,
               message: output,
+              method: spec.method as DescMethodStreaming<I, O>,
```
timostamm, line 117: Defining the `DescMethod*` types as an intersection (e.g. `DescMethod & { methodKind: "unary" }`) doesn't limit the union of `methodKind` in all contexts:

```ts
function f(x: DescMethodUnary<any, any>) {
  x.methodKind; // "unary" | "server_streaming" | "client_streaming" | "bidi_streaming"
}
```

I think we can replace the property instead (`Omit<DescMethod, "methodKind"> & {methodKind: "unary"}`). We shouldn't need the cast here anymore, and users won't need it either in similar contexts.

Not sure if this also applies to `input` and `output`.

timostamm (approved):
LGTM!

What do you think about calling this change "Replace MethodInfo with Descriptor Types"? Would probably provide a bit more context for users reading through the changelog.

PR #1266, opened 2024-10-10
V2: Fix Connect-Query v2.0.0-beta.1 in connect-migrate

timostamm:
Missed some TODOs in https://github.com/connectrpc/connect-es/pull/1264.

### DIFF
--- a/packages/connect-migrate/src/migrations/v2.0.0.ts
+++ b/packages/connect-migrate/src/migrations/v2.0.0.ts
@@ -83,16 +83,15 @@ const dependencyMigrations: DependencyMigration[] = [
     remove: { name: "@connectrpc/protoc-gen-connect-es", range: "^1.0.0" },
   },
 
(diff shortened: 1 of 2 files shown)


packages/connect-migrate/src/migrations/v2.0.0.ts
```diff
@@ -119,7 +118,7 @@ const bufGenYamlMigrations: BufGenYamlMigration[] = [
     updatePlugin: {
       remote: "buf.build/connectrpc/query-es",
       from: "^1.0.0",
-      to: "1.4.2", // TODO
+      to: targetVersionConnectQuery,
```
timostamm, line 121: This will update the remote plugin to the beta version, which will not be available as a pre-release (only stable versions are). I think that's acceptable, since the migration is also beta.

smaye81 (approved):

PR #1268, opened 2024-10-10
V2: Fix transform for `createPromiseClient` -> `createClient` in connect-migrate

paul-sachs:
Existing migration missed a corner case where not all the imports used were to be renamed.


timostamm (approved):
So the bug is that connect-migrate updates imports for `createPromiseClient` or `PromiseClient`, but only if the import statement doesn't import anything else.

Thanks for the fix!

Issue #1271, opened 2024-10-11
Allow user-provided User-Agent request header


polRk:
https://github.com/connectrpc/connect-es/blob/3d1019187838069b849b1ec216d7fd795d0f878d/packages/connect/src/protocol-grpc/request-header.ts#L43

User header must have priority and override useragent header defined buy library

timostamm:
The user-agent set by connect-es is important to use because it helps users and us to diagnose issues. 

Can you provide some details on why it is important to you to override the header?

polRk:
> The user-agent set by connect-es is important to use because it helps users and us to diagnose issues.
> 
> Can you provide some details on why it is important to you to override the header?


It is very important to add information not only about connect-es, but also about runtime and even about the product

timostamm:
Okay, it's reasonable to _not_ set the User-Agent request header if the user already provided one via `CallOptions.headers`. 

This applies to the Connect, gRPC, and gRPC-Web transports, but not to the ones from `@connectrpc/connect-web`, since we do not set the header there (https://github.com/connectrpc/connect-es/pull/912). 

Until this change is made, you can use an interceptor to set a custom user agent:

```ts
const transport = createConnectTransport({
  baseUrl: "https://demo.connectrpc.com",
  interceptors: [
    (next) => async (req) => {
      req.header.set("User-Agent", "FOO");
      return await next(req);
    }
  ]
});
```

polRk:
The problem is: Using interceptors introduces unnecessary overhead, especially when using AsyncLocalStorage.

timostamm:
Closed by https://github.com/connectrpc/connect-es/pull/1272.

It would be convenient to have a transport-level option, but we try to keep the API lean. Interceptors (as suggested in https://github.com/connectrpc/connect-es/issues/1271#issuecomment-2411048990), shouldn't have a perceivable impact on performance.

PR #1272, opened 2024-10-11
Allow user-provided User-Agent request header

polRk:
#1271

### DIFF
--- a/packages/connect/src/protocol-grpc/request-header.spec.ts
+++ b/packages/connect/src/protocol-grpc/request-header.spec.ts
@@ -48,6 +48,17 @@ describe("requestHeader", () => {
     expect(headers.get("Grpc-Timeout")).toBe("10m");
   });
 
+  it("should create request headers with user provided user-agent", () => {
+    const headers = requestHeader(true, 10, { "User-Agent": "grpc-es/0.0.0" });
+    expect(listHeaderKeys(headers)).toEqual([
+      "content-type",
+      "grpc-timeout",
+      "te",
+      "user-agent",
+    ]);
+    expect(headers.get("User-Agent")).toBe("grpc-es/0.0.0");
+  });
+
   it("should create request headers with compression", () => {
     const compressionMock: Compression = {
       name: "gzip",
--- a/packages/connect/src/protocol-grpc/request-header.ts
+++ b/packages/connect/src/protocol-grpc/request-header.ts
@@ -32,11 +32,18 @@ export function requestHeader(
   timeoutMs: number | undefined,
   userProvidedHeaders: HeadersInit | undefined,
 ): Headers {
-  const result = new Headers(userProvidedHeaders);
+  const result = new Headers(userProvidedHeaders ?? {});
   result.set(
     headerContentType,
     useBinaryFormat ? contentTypeProto : contentTypeJson,
   );
+
+  if (!result.has(headerUserAgent)) {
+    // Note that we do not strictly comply with gRPC user agents.
+    // We use "connect-es/1.2.3" where gRPC would use "grpc-es/1.2.3".
+    // See https://github.com/grpc/grpc/blob/c462bb8d485fc1434ecfae438823ca8d14cf3154/doc/PROTOCOL-HTTP2.md#user-agents
+    result.set(headerUserAgent, "CONNECT_ES_USER_AGENT");
+  }
 
   if (timeoutMs !== undefined) {
     result.set(headerTimeout, `${timeoutMs}m`);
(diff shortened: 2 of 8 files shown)


packages/connect/src/protocol-grpc/request-header.ts
```diff
@@ -32,18 +32,16 @@ export function requestHeader(
...
 ): Headers {
-  const result = new Headers(userProvidedHeaders ?? {});
+  const result = new Headers(userProvidedHeaders);
   result.set(
     headerContentType,
     useBinaryFormat ? contentTypeProto : contentTypeJson,
   );
-  // Note that we do not strictly comply with gRPC user agents.
-  // We use "connect-es/1.2.3" where gRPC would use "grpc-es/1.2.3".
-  // See https://github.com/grpc/grpc/blob/c462bb8d485fc1434ecfae438823ca8d14cf3154/doc/PROTOCOL-HTTP2.md#user-agents
```
timostamm, line 42: If you follow the link in the comment, you'll see that the gRPC spec recommends that clients provide a user-agent.

polRk: Yes, and I want to pass my user-agent: to understand who my consumers are, add the runtime version, the library version

polRk: @timostamm I've updated PR

timostamm (changes requested):
Thanks for the PR. If you can update it to only make the change to not set the header if the user already provided one, we can merge it.

I understand the idea, but we do not want to make a breaking change to user agent strings. Please avoid unnecessary changes to whitespace (`npx turbo run format` runs prettier).

polRk:
> Thanks for the PR. If you can update it to only make the change to not set the header if the user already provided one, we can merge it.

So, are you saying that the logic for generating the grpc-js-* string should be removed, and we should just pass the header instead?

timostamm:
> So, are you saying that the logic for generating the grpc-js-* string should be removed

Yes.

> and we should just pass the header instead?

Currently, we're always setting the `User-Agent` header, regardless of whether the user has already provided one. For example:

```ts
await client.say(
  { sentence: "hello" },
  {
    headers: {
      "User-Agent": "foo",
    }
  }
);
```

The `User-Agent: foo` is overwritten.

Following https://github.com/connectrpc/connect-es/issues/1271#issuecomment-2411048990, we should change the behavior, and not overwrite the user-provided value, but continue to set our value if the user didn't provide one.

The important part is that we do it consistently, not just in the gRPC transport, but also for gRPC-Web and the Connect transport.

Adding a transport option for a default user-agent is a nice idea, but it raises the question whether we shouldn't have an option for default headers instead. For now, the best action is to not add this option, and consider this change separately.

polRk:
> Currently, we're always setting the `User-Agent` header, regardless of whether the user has already provided one. For example:

The problem is:

1. Passing headers would require changing the code throughout the entire codebase. In my previous project, I had thousands of gRPC requests, and modifying all of them would be difficult.

2. Using interceptors introduces unnecessary overhead, especially when using AsyncLocalStorage.

timostamm:
To fix #1271, we can allow to pass a custom user agent via `CallOptions.headers`, which is a reasonable incremental change.

We can discuss other features separately.

polRk:
@timostamm

packages/connect/src/protocol-grpc/request-header.spec.ts
```diff
@@ -34,7 +34,7 @@ describe("requestHeader", () => {
       "user-agent",
     ]);
     expect(headers.get("Content-Type")).toBe("application/grpc+proto");
-    expect(headers.get("User-Agent")).toMatch(/^connect-es\/\d+\.\d+\.\d+/);
+    expect(headers.get("User-Agent")).toMatch(/^connect-es\/.*/);
```
timostamm, line 37: Let's not modify this test. It makes sure the default user agent doesn't regress.

packages/connect/src/protocol-grpc/request-header.spec.ts
```diff
@@ -48,6 +48,17 @@ describe("requestHeader", () => {
     expect(headers.get("Grpc-Timeout")).toBe("10m");
   });
 
+  it("should create request headers with userAgent", () => {
```
timostamm, line 51: Thanks for the test!

packages/connect/src/protocol-grpc/request-header.ts
```diff
@@ -37,13 +37,18 @@ export function requestHeader(
...
-  result.set(headerUserAgent, "CONNECT_ES_USER_AGENT");
+
+  if (!result.has(headerUserAgent)) {
```
timostamm, line 41: Nice and simple 👍

timostamm (commented):
Looks great. 

Can you apply the same change to the Connect and gRPC-Web transports for consistency? It would be very confusing if this only works with gRPC. The transports from `@connectrpc/connect-web` don't set a user-agent, no need to update them.

polRk:
@timostamm Greetings! Will you be able to watch it?

PR #1278, opened 2024-10-15
V2: Fix CI

srikrsna-buf:
Cloudflare run: https://github.com/connectrpc/connect-es/actions/runs/11347159865/job/31557879997


packages/connect-migrate/src/migrations/v2.0.0.spec.ts
```diff
@@ -396,7 +396,7 @@ plugins:
             `
 version: v2
 plugins:
-  - remote: remote: buf.build/bufbuild/connect-es:v0.13.0
+  - remote: buf.build/bufbuild/connect-es:v0.13.0
```
srikrsna-buf, line 399: No clue how it worked before

packages/connect-node/src/node-readme.spec.ts
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+/* eslint-disable @typescript-eslint/unbound-method */
```
timostamm, line 15: It looks like we might be destructuring the router in examples, which means that users are likely to also run into this lint rule. 

We can change the signatures:

```ts
  service: <T extends DescService>(
    service: T,
    implementation: Partial<ServiceImpl<T>>,
    options?: Partial<UniversalHandlerOptions>,
  ) => this;
  rpc: <M extends DescMethod>(
    method: M,
    impl: MethodImpl<M>,
    options?: Partial<UniversalHandlerOptions>,
  ) => this;
```

This makes them function properties instead of methods.

I don't see how this change could cause any problems, but it solves the linter issue for us and for users. WDYT?

timostamm (approved):

PR #1279, opened 2024-10-15
V2: Require HTTP/2 for the gRPC transport

### LINKED ISSUE #1258: httpVersion check for `createRouterTransport`
When Connect upgrade to v1.5, it starts to check the gRPC code and gRPC message from trailer or header by default. We encountered an issue when using`createRouterTransport` with `httpVersion: 1.1`, the test didn't pick up the breaking change for us.

I am wondering if possible to check the transport option when using `createRouterTransport`, ensure it's similar to testing against with real service.

Please specify whether the request is for Connect for Web or Connect for Node.js.
Connect for Node.js

I am not sure of the best way to handle this, but perhaps we could verify the transport option like enforce checking the `httpVersion` starting from v1.5.

**Additional context**
https://github.com/connectrpc/connect-es/pull/1205/files#diff-92e5a379d00b073cbd57f841a9669260a8e9603c2015e09aeeb49daf6fe788b8R40

timostamm:
gRPC implementations typically require HTTP/2, but the protocol itself does not. connect-es can run gRPC over HTTP 1.1, but that requires support for HTTP trailers end to end. This can be difficult to achieve and guarantee, and most other gRPC implementations do not support it, so we are disabling support for gRPC over HTTP 1.1 in version 2.

Closes #1258.


packages/connect-node/src/grpc-transport.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+  it("should take just baseUrl", function () {
+    const t = createGrpcTransport({
+      baseUrl: "https://example.com",
+    });
+    expect(t).toBeDefined();
+  });
+  it("should raise type error for httpVersion: 2", function () {
+    const t = createGrpcTransport({
+      // @ts-expect-error TS2353: Object literal may only specify known properties, and httpVersion does not exist in type GrpcTransportOptions
+      httpVersion: "2",
```
timostamm, line 27: `httpVersion` cannot be specified for createGrpcTransport anymore. It always uses HTTP/2.

srikrsna-buf (approved):

PR #1319, opened 2024-11-05
Update dependencies

srikrsna-buf:
Group and update several dependabot updates.


packages/example/package.json
```diff
@@ -23,7 +23,7 @@
   "devDependencies": {
     "@bufbuild/buf": "^1.46.0",
     "@bufbuild/protoc-gen-es": "^2.1.0",
-    "@types/express": "^4.17.18",
+    "@types/express": "^5.0.0",
```
timostamm, line 26: It looks like we aren't using this at all. The example uses vanilla Node modules. Can you remove it?

timostamm (approved):
Please see https://github.com/connectrpc/connect-es/pull/1319#discussion_r1829197603

PR #1340, opened 2024-11-25
Fix toBinary example in MIGRATING.md

HermanBilous:
toBinary is declared as 
```typescript
export declare function toBinary<Desc extends DescMessage>(schema: Desc, message: MessageShape<Desc>, options?: Partial<BinaryWriteOptions>): Uint8Array;
```

### DIFF
--- a/MIGRATING.md
+++ b/MIGRATING.md
@@ -203,7 +203,7 @@ import { toBinary } from "@bufbuild/protobuf";
 import { SayRequestSchema } from "./gen/eliza_pb";
 
 - sayRequest.toBinary();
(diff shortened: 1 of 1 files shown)


timostamm:
Thanks for the PR, Herman.

I think we already fixed this typo in https://github.com/connectrpc/connect-es/pull/1338.

Do we have a link somewhere that points to MIGRATING.md on the v2.0.0 tag? Please let us know if we do, and we'll fix it to point to the main branch instead, where the typo is fixed.

timostamm (approved):
Oh, thanks for the fix Herman!

PR #1360, opened 2024-12-02
Update dependencies

srikrsna-buf:
Barring TS and eslint handles all the dependabot updates. TS update is causing an import error while importing json files. Better to handle that in a dedicated PR.

### DIFF
--- a/packages/typescript-compat/v4.9.x/package.json
+++ b/packages/typescript-compat/v4.9.x/package.json
@@ -6,7 +6,7 @@
   "dependencies": {
     "@connectrpc/connect-conformance": "*",
     "@connectrpc/example": "*",
-    "@types/node": "22.10.1",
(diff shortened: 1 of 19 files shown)


packages/typescript-compat/v4.9.x/package.json
```diff
@@ -6,7 +6,7 @@
   "dependencies": {
     "@connectrpc/connect-conformance": "*",
     "@connectrpc/example": "*",
-    "@types/node": "22.9.0",
+    "@types/node": "22.10.1",
```
timostamm, line 9: The `@types/node` package has tags for the TypeScript versions it's compatible with. Looks like we can only update to 22.9.3 here. See https://www.npmjs.com/package/@types/node?activeTab=versions

timostamm (approved):

PR #1364, opened 2024-12-05
Add additional options for generating imports

smaye81:
Including imports is not enabled by default in the Buf CLI. As a result, users have been getting tripped up on generating code when they have dependencies on protovalidate.

This adds some additional guidance in our generating code docs to make sure they include imports since these should essentially always be included when generating with Protobuf-ES.

### DIFF
--- a/MIGRATING.md
+++ b/MIGRATING.md
@@ -162,7 +162,7 @@ way your project is configured. For example, `npx buf generate` or `npm run gene
 >
 > - `include_imports: true`
 
-See the [Gotchas](#missing-imports) section for an explanation.
+See the [Gotchas](#the-new-plugins-generates-missing-imports) section for an explanation.
 
 ## Update your application code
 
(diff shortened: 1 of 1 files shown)


smaye81:
@timostamm I added `include_imports` to all snippets of `buf.gen.yaml` just to reinforce that it should be there. But if you'd rather it be listed only in`Regenerating code`, lmk.

MIGRATING.md
```diff
@@ -709,7 +722,7 @@ Make sure to update your imports to the new package name and file names:
 
 ## Gotchas
 
-### The new plugins generates missing imports
+### The new plugins generates missing imports <a id="missing-imports"></a>
```
timostamm, line 725: There's already a standard anchor for all headlines. It's used by the GH outline feature. Best to stick to the standard anchor.

```suggestion
### The new plugins generates missing imports
```

smaye81: Sounds good. I was just worried that if the wording changed for that heading, the link to it would be broke.

timostamm (approved):

Issue #669, opened 2023-06-06
Support Client Streaming for fetch API


oott123:
**Is your feature request related to a problem? Please describe.**
We cannot use client stream on web.

**Describe the solution you'd like**
Support client stream requests with the fetch API by default.

**Describe alternatives you've considered**
I can write use own transport function.

**Additional context**
[Streaming requests with the fetch API](https://developer.chrome.com/articles/fetch-streaming-requests/) by Chrome Developers

oott123:
Actually, Firefox already supported this feature, at least here on my machine, and yes I completely agree that we should add tests to make sure that it works correctly.

timostamm:
> Actually, Firefox already supported this feature, at least here on my machine

Could you share which version you are running? We did not see the feature detection recommended by the Chrome Developers Blog working with the [developer edition](https://www.mozilla.org/en-CA/firefox/developer/).

oott123:
@timostamm  I'm so sorry, I have re-checked this on my Firefox (114.0.1) browser and it do does NOT work. Sorry for my mistake, I may confused between browsers.

smaye81:
If you want a pure Connect solution, the best option for large files would probably be to chunk up the file on the client and then send individual unary requests to the backend and then reassembling the chunks there.

In addition, depending on the size of the file, you could potentially send it all in one request rather than having to chunk it.

StarpTech:
@smaye81 came to the same conclusion but since client-side streaming is not supported on the browser what's the workaround there?

smaye81:
Since client-streaming is not yet supported from browsers, those are probably your options. You could use regular HTTP and multipart form data, also. But if you wanted a pure Connect approach from a browser, then your options are limited unfortunately.

Some gRPC-Web implementations such as Improbable's gRPC-Web client facilitate client-streaming via websockets, but Connect does not support that at the moment.

StarpTech:
This is sad but understandable. It would be nice if this workaround is somehow managed by Connect protocol. As a Dev, I need to plumb all together again, which I wanted to prevent by using Connect in the first place.

smaye81:
Yeah, that's fair. We're keeping an eye on support for sending `ReadableStream` and evaluating possible alternatives, so hopefully, we have an answer for this in the near future.

hydrz:
#[b5e5c96](https://github.com/opensraph/connect-es/commit/b5e5c96b909905f1de3c0ef2966a48f221315ab0)

[...]

PR #1375, opened 2024-12-30
@connectrpc/connect-web unlock client and bidi streaming in supported browsers

davidfiala:
As discussed in https://github.com/connectrpc/connect-es/issues/669#issuecomment-2564797946 this PR brings client-side streaming (client or bidi) to supported browsers in `@connectrpc/connect-web`. The behavior in Google's grpc-web appears to allow client-side streaming in supported browsers and has quirky behavior in unsupported ones (see linked comment for specifics). However, this PR aims to simplify the developer experience by immediately throwing if a client-side streaming RPC is started on an unsupported browser. I believe that this PR does not introduce any harm or backwards compatibility errors because the preexisting case was that all client-side streaming calls would throw. We only unlock the capability when possible, and effectively do not change the existing behavior for unsupported browsers.

Support browsers are known to include Chromium-based browsers. Unsupported browsers are known to include Firefox and Safari.

Previous error message: `The fetch API does not support streaming request bodies`
Revised to: `The fetch API does not support streaming request bodies in this browser`

Note that the logic to detect compatibility is following the Chrome Developer docs which explains the nuance around how/why it works: https://developer.chrome.com/docs/capabilities/web-apis/fetch-streaming-requests#feature_detection

There's an important caveat: Only half duplex support is [available](https://github.com/whatwg/fetch/issues/1254). So while bidi streams do work, you will not see server responses until after the client finishes sending its stream. Once the client finishes its iterator, all prior server responses are immediately sent. From that point on, if the server's iterator is still producing more responses, those will continue to stream in as per normal.

### DIFF
--- a/packages/connect-web/src/connect-transport.ts
+++ b/packages/connect-web/src/connect-transport.ts
@@ -129,6 +129,24 @@ const fetchOptions: RequestInit = {
   redirect: "error",
 };
 
+const supportsRequestStreams = (() => {
+  let duplexAccessed = false;
+
+  // explicit casting. RequestInit does not standardly support `duplex` yet
+  const hasContentType = new Request("https://example.com/", {
+    body: new ReadableStream(),
+    method: "POST",
+    get duplex() {
+      duplexAccessed = true;
+      return "half";
+    },
+  } as RequestInit).headers.has("Content-Type");
+
+  // eslint is unaware that `get duplex()` may be called
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+  return duplexAccessed && !hasContentType;
+})();
+
 /**
  * Create a Transport for the Connect protocol, which makes unary and
  * server-streaming methods available to web browsers. It uses the fetch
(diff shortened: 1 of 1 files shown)


timostamm:
Thanks for opening the PR, David.

I'm afraid we prefer not to enable this feature until all major browsers support it. 

Please feel free to publish a transport that does support it!

guest271314:
> So while bidi streams do work, you will not see server responses until after the client finishes sending its stream. 

FYI Notwithstanding WHATWG Fetch #1254, in Chromium-based browsers it is possible to full-duplex stream over WHATWG Fetch implementation, where internally Mojo is used between a `ServiceWorker` and a `WindowClient` or `Client`. `fetch()` can be intercepted in `fetch` event and a `ReadableStream` can be passed to `respondWith()`, which in the `WindowClient` will be fulfilled before the upload stream is completely read. See this plnkr for an example https://embed.plnkr.co/plunk/2PQK21kZTZvZ2oVi.

PR #1395, opened 2025-01-06
Update dependencies

srikrsna-buf:


packages/connect-express/package.json
```diff
@@ -41,7 +41,7 @@
   },
   "peerDependencies": {
     "express": "^4.18.2 || ^5.0.1",
-    "@bufbuild/protobuf": "^2.2.0",
+    "@bufbuild/protobuf": "^2.2.3",
```
timostamm, line 44: Does `@bufbuild/protobuf` v2.2.3 include any changes we need for `@connectrpc/connect-express`? If not, wouldn't it be better to keep the lowest compatible version at v2.2.0?

Same for the other packages with constraints relevant for end-users.

packages/typescript-compat/v4.9.x/package.json
```diff
@@ -7,7 +7,7 @@
   "dependencies": {
     "@connectrpc/connect-conformance": "*",
     "@connectrpc/example": "*",
-    "@types/node": "22.9.3",
+    "@types/node": "22.10.5",
```
timostamm, line 10: Do we really want to bump this version for TS 4.9.x? The package has dist tags for every TS version.
Same for the other compat packages.

timostamm (approved):

PR #1399, opened 2025-01-07
Add support for import aliases

smaye81:
This adds support for converting import aliases when running the migrate tool. The same alias scheme is applied to the `Schema` imports added by the migration tool.

For example:

```js
import { Foo as MyFoo } from "./foo_pb.js";` --> `import { FooSchema as MyFooSchema } from "./foo_pb.js";
```


timostamm (approved):
Nice, let's get it merged.

Issue #1428, opened 2025-02-17
Otel looses Trace Context while streaming


jonathangraf:
As stated [here](https://github.com/connectrpc/connect-es/issues/988) Otel support should implement yourself via Interceptors. But when doing so, it seems that while streaming, Otel seems to lose its Context and doesn't link the traces correctly. 
Here is my implementation of it:

```typescript
import { context, propagation, Span, SpanKind, trace } from '@opentelemetry/api';

import { Interceptor } from '@connectrpc/connect';

export type Carrier = { 
    traceparent?: string;
    tracestate?: string;
}

const tracer = trace.getTracer('connect-rpc');

type InterceptorResponse = Awaited<ReturnType<ReturnType<Interceptor>>>;

const handleSpanEnd = (response: InterceptorResponse, span: Span) => {
    if (!response.stream) {
        span.end();

        return response;
    }

    async function* stream() {
        if (!response.stream) {
            throw new Error('Stream is not defined');
        }

        const activeContext = context.active();

        // eslint-disable-next-line fp/no-loops
        for await (const message of response.message) {
            yield context.with(activeContext, () => message);
        }

        span.end();
    }
    
    return {
        ...response,
        message: stream(),
    };
};

export const serverInterceptor: Interceptor[] = [
    (next) => async (request) => {
        const parentContext = propagation.extract(
            context.active(),
            Object.fromEntries(request.header.entries()),
        );
        const span = tracer.startSpan(
            `${request.service.typeName}/${request.method.name}`,
            {
                kind: SpanKind.SERVER,
            },
            parentContext,
        );

        return context.with(trace.setSpan(parentContext, span), async () => {
            const activeContext = context.active();
            context.bind(activeContext, request);
            context.bind(activeContext, next);
            const response = await next(request);
            context.bind(activeContext, response);
            
            return handleSpanEnd(response, span);
        });
    },
];

export const clientInterceptor: Interceptor[] = [ 
    (next) => async (request) => {
        const span = tracer.startSpan(
            `${request.service.typeName}/${request.method.name}`, {
                kind: SpanKind.CLIENT,
            });
        const parentContext = context.active();

        return context.with(trace.setSpan(parentContext, span), async () => {
            const activeContext = context.active();
            context.bind(activeContext, request);
            const output: Carrier = {};
            propagation.inject(activeContext, output);
            Object.entries(output).forEach(([key, value]) => {
                key && value && request.header.set(key, value);
            });

            context.bind(activeContext, next);
[...]

PR #1429, opened 2025-02-18
Fix server interceptor not seeing the request message in server streaming rpcs

srikrsna-buf:
Closes https://github.com/connectrpc/connect-es/issues/1428. Because of the conversion between handler types interceptor, we missed using the message that we get from the interceptor chain. I also added similar assertions for other rpc types.

### DIFF
--- a/packages/connect/src/protocol/invoke-implementation.ts
+++ b/packages/connect/src/protocol/invoke-implementation.ts
@@ -106,7 +106,7 @@ export function transformInvokeImplementation<
             const output = normalizeIterable(
               spec.method.output,
               spec.impl(
-                await ensureSingle(input, "server-streaming"),
+                await ensureSingle(req.message, "server-streaming"),
                 mergeRequest(context, req),
               ),
             );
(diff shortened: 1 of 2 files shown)


packages/connect/src/protocol/invoke-implementation.ts
```diff
@@ -106,7 +106,7 @@ export function transformInvokeImplementation<
             const output = normalizeIterable(
               spec.method.output,
               spec.impl(
-                await ensureSingle(input, "server-streaming"),
+                await ensureSingle(req.message, "server-streaming"),
```
timostamm, line 109: Makes sense, but we should double check with https://github.com/connectrpc/examples-es/pull/2452 before cutting a release.

timostamm (approved):

packages/connect/src/protocol/invoke-implementation.ts
srikrsna-buf: Checked against the example by overriding the connect package: https://github.com/connectrpc/examples-es/compare/Add-example-for-interceptors...sk/interceptor-check?expand=1

timostamm: Thank you for confirming!

PR #1454, opened 2025-03-23
feat: throw DeadlineExceeded caused by AbortSignal.timeout

polRk:


timostamm (commented):
Thanks! We'll have to use `Code.Canceled`, see comments below.

Can you add test below https://github.com/connectrpc/connect-es/blob/v2.0.2/packages/connect/src/connect-error.spec.ts#L157?

```ts
    it("wraps AbortError with code canceled", () => {
      // abort() on AbortSignal aborts with a AbortError
      const error: unknown = new DOMException("foo", "AbortError");
      const got = ConnectError.from(error);
      expect(got.code).toBe(Code.Canceled);
      expect(got.rawMessage).toBe("foo");
    });
    it("wraps TimeoutError with code canceled", () => {
      // AbortSignal.timeout() aborts with a TimeoutError
      const error: unknown = new DOMException("foo", "TimeoutError");
      const got = ConnectError.from(error);
      expect(got.code).toBe(Code.Canceled);
      expect(got.rawMessage).toBe("foo");
    });
```

Please run `npx turbo run format bundle-size` to update formatting and bundle-size benchmarks.

PR #1464, opened 2025-04-01
Update `routes` option within adapter builders to allow multiple service functions

d3rpp:
This is a fairly simple PR that adds the ability to include more than one function in the `routes` option of the various adapters.

I've been able to test it successfully on a small Bun project (internally at my company unfortunately).

This change is backwards compatible as you're still able to pass a single function.

If this feature requires its own conformance tests, I'd be happy to look into it, though none of the adapter/middleware/etc. files seem to have tests as is.

> Apologies for all of the force pushes, I'm not used to signing off commits and the commits were using the wrong email.


d3rpp:
Apologies for the force push, my company gets weirdly annoyed when they see a company email in a GH pr, so I rewrote the commit emails after realising (and added sign-offs)

timostamm:
Apologies for the late reply. 

While the `router` option might not be perfect for every use case, it's simple and flexible. For example, you could delegate to other router functions:

```ts
const adapter = createNodeAdapter({
  routes(router) {
    ElizaRouter(router);
    HelloWorldRouter(router);
  }
});
```

Or:

```ts
const adapter = createNodeAdapter({
  routes: (router) => [ElizaRouter, HelloWorldRouter].forEach(fn => fn(router)),
});
```

For now, we want to keep it simple, but thanks for putting up a PR!

PR #1485, opened 2025-05-01
Replace prettier and eslint with biome

timostamm:
Same as https://github.com/bufbuild/protobuf-es/pull/1104.


packages/connect-migrate/src/migrations/v1.6.0-transform.ts
```diff
@@ -138,12 +144,14 @@ function replace(
   qualifier?: string,
 ) {
   if (qualifier === undefined) {
-    root
-      .find(j.Identifier, { name: from })
-      .forEach((path) => (path.value.name = to));
+    // biome-ignore lint/complexity/noForEach: not alternative to forEach available
+    root.find(j.Identifier, { name: from }).forEach((path) => {
```
paul-sachs, line 148: I guess we could also do something like:

```tsx
for (const path of root.find(j.Identifier, { name: from }).paths()) {
  ...
```

Which would avoid the forEach. Not really sure it's worth the effort just to satisfy this lint rule 🤷

packages/connect/src/protocol-grpc/trailer-status.spec.ts
```diff
@@ -17,36 +17,39 @@ import { ConnectError } from "../connect-error.js";
...
 import { Int32ValueSchema } from "@bufbuild/protobuf/wkt";
 
-describe("setTrailerStatus()", function () {
-  it("should set grpc-status when called without error", function () {
+describe("setTrailerStatus()", () => {
+  it("should set grpc-status when called without error", () => {
     const t = new Headers();
     setTrailerStatus(t, undefined);
     let count = 0;
+    // biome-ignore lint/complexity/noForEach: Headers is not iterable, and we don't have access to entries()
```
paul-sachs, line 25: I guess we could also add `DOM.Iterable` to `tsconfig.base.json` since they are baseline at this point. Though again, maybe not worth it for now.

paul-sachs (approved):

srikrsna-buf (approved):

packages/connect-migrate/src/migrations/v1.6.0-transform.ts
timostamm: This is already a very large diff, and who knows which linter we'll be using in a couple of months.

PR #1491, opened 2025-05-12
Explicitly set Content-Length for empty request bodies to improve compatibility with http proxies

tmm1:
I have been investigating an issue where certain types of connectrpc calls fail when using an http proxy (specifically Zscaler in this instance).

Investigation revealed the affected calls were all unary requests where the input message has zero fields.

On the wire, the nodejs request in both h1 and h2 modes fail via this proxy, because they have the equivalent of `Transfer-Encoding: chunked` set but send zero bytes in the payload. Signaling this case explicitly with `Content-Length: 0` avoids the buggy behavior in these proxies.


tmm1:
I can reproduce this issue very easily over http1 using the node https module directly.

This works:

```js
const req = https.request(...)
req.end()
```

Both of these fail:

```js
const req = https.request(...)
req.flushHeaders()
req.end()
```

```js
const req = https.request(...)
req.write(Buffer.from(''))
req.end()
```

When the data flows through the Zscaler proxy, there is some bug where it gets confused expecting a streaming request body but not seeing one.

Setting Content-Length explicitly fixes the problem.

Unfortunately the problem is very specific to corporate http middleware so it is hard to reproduce without that kind of environment.

PR #1529, opened 2025-07-17
Add proto syntax highlighting to snippet in README

stefanvanburen:


timostamm (commented):
Thanks! Do you mind adding the same to https://github.com/connectrpc/connect-es/blob/main/packages/connect/README.md?

Issue #1333, opened 2024-11-20
async-iterable append() implementation is slow


tmm1:
![Screenshot 2024-10-18 at 2 04 27 PM](https://github.com/user-attachments/assets/2b2f23d6-bf1a-48a3-afea-356ecad4560b)

https://github.com/connectrpc/connect-es/blob/36af3f296f6e70a59f59135c1658de664581f6bf/packages/connect/src/protocol/async-iterable.ts#L1008-L1017

this causes repeated memcpy on the cpu every time a new chunk comes in.

imagine a long streaming response where a 50kb response is streaming in 10 bytes at a time. towards the end of the stream, a huge amount of data is copied over and over every time 10 bytes are added.

tmm1:
the implementation here should mimic realloc, which generally grows the underlying buffer exponentially to avoid repeated copies.

one option is the uint8arraylist node module. i had some success integrating it, but it only supports ESM whereas connect needs to support CJS as well.

https://github.com/achingbrain/uint8arraylist
https://medium.com/@lagerdata/a-resizable-typedarray-f5b9a861958

timostamm:
Thanks for the issue! It's certainly possible to avoid a lot of allocations and copies.
- Uint8ArrayList keeps a reference of chunks and defers concatenation.
- GrowableUint8Array allocates legroom in a buffer.

The downside of both approaches is that both implementations aren't TypedArrays, and can't be used for views (e.g. Uint8Array, DataView). This complicates using them efficiently for reading envelope sizes.

The proper solution are [resizable ArrayBuffers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer/resize). The API allows growing on demand, makes it easy to allocate for an envelope, and is compatible with views. Unfortunately, it's relatively new, and we have to hold back on using it to support older browsers and Node.js < v20.

Besides `transformSplitEnvelope`, `readAllBytes` from the same file could also make use of resizable ArrayBuffers, as well as some other places in the code base, and several places in Protobuf-ES. So I'm not sure that it's worth putting much work into optimizing `transformSplitEnvelope`, when the more general optimization with native support is in sight.

tmm1:
Thanks for the resizable ArrayBuffers reference. I will see if I can wire that up.

Here's what I am using currently:

```diff
diff --git a/packages/connect/src/protocol/async-iterable.ts b/packages/connect/src/protocol/async-iterable.ts
[...]

PR #1530, opened 2025-07-22
Optimize envelope decoding

timostamm:
This adds the internal type `EnvelopeDecoder`. It decodes a stream of raw byte chunks into enveloped messages. 

It allocates a single ArrayBuffer for envelope headers (always 5 bytes), and allocates an ArrayBuffer with the size of the enveloped message as soon as the size is known. Incoming chunks are buffered in a regular array, and copied to the allocated ArrayBuffer once complete.

This new type is used in `transformSplitEnvelope()` (Node.js) and `createEnvelopeReadableStream()` (Web), consolidating the logic.

Compared to the current approach, this avoids allocating a new buffer each time a chunk comes in, noticeably increasing envelope decoding performance.

Closes https://github.com/connectrpc/connect-es/issues/1333

### DIFF
--- a/packages/connect/src/protocol/async-iterable.ts
+++ b/packages/connect/src/protocol/async-iterable.ts
@@ -14,7 +14,7 @@
 
 import { Code } from "../code.js";
 import { ConnectError } from "../connect-error.js";
-import type { EnvelopedMessage } from "./envelope.js";
+import { createEnvelopeDecoder, type EnvelopedMessage } from "./envelope.js";
 import {
   encodeEnvelope,
   envelopeCompress,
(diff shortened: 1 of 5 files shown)


srikrsna-buf (approved):

tmm1 (approved):

packages/connect/src/protocol/async-iterable.ts
```diff
@@ -14,7 +14,7 @@
 
 import { Code } from "../code.js";
 import { ConnectError } from "../connect-error.js";
-import type { EnvelopedMessage } from "./envelope.js";
+import { createEnvelopeDecoder, type EnvelopedMessage } from "./envelope.js";
 import {
```
tmm1, line 18:
```suggestion
import type { EnvelopedMessage } from "./envelope.js";
import {
  createEnvelopeDecoder,
```

PR #1559, opened 2025-08-13
Drop support for Node.js 18

timostamm:
This PR removes Node.js 18 from CI. 

Node.js 18 has reached its end of life earlier this year, and does not receive security updates anymore (see the [release schedule](https://nodejs.org/en/about/previous-releases)). 

If you are running Connect on Node.js 18, this doesn't mean that it will stop working right away, but we stop testing on  this version of Node.js, and will no longer consider failures a bug. However, we strongly recommend updating to a more recent version (see this [blog post](https://nodejs.org/en/blog/announcements/node-18-eol-support) for more information).

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,8 +62,8 @@ to spin up a service in Node.js, and call it from the web, and from a gRPC clien
 in your terminal.
 
 You can serve your Connect RPCs with vanilla Node.js, or use our [server plugins](https://connectrpc.com/docs/node/server-plugins)
-for **Fastify**, **Next.js**, and **Express**. We support Node.js v20 with the
-builtin `http` and `http2` modules.
+for **Fastify**, **Next.js**, and **Express**. We support Node.js v20 and later with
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -62,8 +62,8 @@ to spin up a service in Node.js, and call it from the web, and from a gRPC clien
 in your terminal.
 
 You can serve your Connect RPCs with vanilla Node.js, or use our [server plugins](https://connectrpc.com/docs/node/server-plugins)
-for **Fastify**, **Next.js**, and **Express**. We support Node.js v18.14.1 and later with
-the builtin `http` and `http2` modules.
+for **Fastify**, **Next.js**, and **Express**. We support Node.js v20 with the
```
smaye81, line 65: Worth rephrasing to `v20 and later...` just to avoid any confusion?

smaye81 (approved):

PR #1651, opened 2026-03-18
Replace undici with headers-polyfill in connect-node

danwsong:
Undici v5 is affected by CVE-2026-1526 and CVE-2026-2229 (WebSocket permessage-deflate vulnerabilities). Since connect-node only uses undici for the Headers class polyfill on Node < 18, this replaces it with the lightweight headers-polyfill package. No engine requirement changes — Node >= 16 continues to work.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2856,6 +2856,15 @@
         "@types/send": "*"
       }
     },
+    "node_modules/@types/set-cookie-parser": {
+      "version": "2.4.10",
+      "resolved": "https://registry.npmjs.org/@types/set-cookie-parser/-/set-cookie-parser-2.4.10.tgz",
+      "integrity": "sha512-GGmQVGpQWUe5qglJozEjZV/5dyxbOOZ0LHe/lqyWssB88Y4svNfst0uqBVscdDeIKl5Jy5+aPSvy7mI9tYRguw==",
+      "license": "MIT",
+      "dependencies": {
+        "@types/node": "*"
+      }
+    },
     "node_modules/@types/sinonjs__fake-timers": {
       "version": "8.1.5",
       "resolved": "https://registry.npmjs.org/@types/sinonjs__fake-timers/-/sinonjs__fake-timers-8.1.5.tgz",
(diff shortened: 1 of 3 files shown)


danwsong:
Hi @timostamm, would you be open to backporting CVE fixes to v1 like this?

timostamm:
Hey @danwsong, thanks for the PR.

The CVEs are about the WebSocket client in undici. There is no usage of the WebSocket client in connect-es. 

I assume the problem is that the dependency is flagged regardless?

danwsong:
@timostamm Yes, precisely. It's true that this library doesn't use undici for the WebSocket client, but it's still getting flagged. I looked at a few options before landing on this:
- Upgrading undici to version 6
  - Version 6 drops support for Node.js versions < 18, but those versions are the ones that required the Headers polyfill
- Dropping support for Node.js versions < 18
  - This is technically a breaking change, and probably wouldn't be feasible for a backport
- Removing undici as a dependency
  - This would require manually writing a polyfill, which led me ultimately to:
- Replacing undici with headers-polyfill
  - The package is actively maintained by the MSW team and is significantly lighter at 113 KB vs. undici at 1.6 MB

danwsong:
@timostamm Are any of these checks of concern?

danwsong:
@timostamm I'd like to follow up on this, would you have any concerns with this change in principle?

AlbinoGeek:
Checking in on status here. A few of our projects are working around the undici pin in connect-node and would benefit from this landing. We're watching both this PR and #1729 and are curious which approach is preferred going forward — dropping undici entirely seems like the cleaner long-term fix. Any update on timing or what's needed to move it forward?

ajeetdsouza:
Hi @danwsong, the CVE you mentioned doesn't affect any version of connect-es. If the audit report is an issue, this is something that is fixed on connect-es v2 - I would highly recommend upgrading. It should be fairly straightforward, and we also provide a tool for migration: https://github.com/connectrpc/connect-es/blob/main/MIGRATING.md.

If that is not workable, you can remove the offending undici version from the dependency graph with an [npm override](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#overrides).

PR #1688, opened 2026-05-04
Add TypeScript 5.9 compat package

ajeetdsouza:
Add a compatibility package to check that `connect-es` works with TypeScript 5.9's default `tsconfig.json`.


packages/connect-conformance/src/protocol.ts
```diff
@@ -76,12 +76,12 @@ export function getSingleRequestMessage<T extends DescMessage>(
   compatRequest: ClientCompatRequest,
   desc: T,
 ): MessageShape<T> {
-  if (compatRequest.requestMessages.length !== 1) {
+  const any = compatRequest.requestMessages[0];
+  if (any === undefined || compatRequest.requestMessages.length > 1) {
```
srikrsna-buf, line 80: I am unsure why these changes are needed, the repo itself has been on 5.9 and these were built using that so far. Is it because of the default options?

If they are because of the default options maybe we should explore changing them in the root tsconfig first.

ajeetdsouza: Yeap, the default options have changed.

Using these options in the compat package only checks that the public API follows the required conventions. I did try adding the flags to the root `tsconfig.base.json` and it flagged type errors everywhere, and IMO they made the code harder to read in a lot of places.

I propose we limit the new flags to the public API, we can always add the flag to the root tsconfig later if we feel like that's an upgrade.

timostamm (approved):

PR #1689, opened 2026-05-04
Upgrade to TypeScript 6.0

ajeetdsouza:
- Upgrade TypeScript dependency to 6.0.
- Add a compatibility package to check that `connect-es` works with TypeScript 6.0's default `tsconfig.json`.


tsconfig.base.json
```diff
@@ -29,6 +39,8 @@
     "moduleResolution": "Node16",
     "module": "Node16",
     "verbatimModuleSyntax": true,
-    "skipLibCheck": false
-  }
+    "skipLibCheck": false,
+    // Silence TS 6.0 deprecation of moduleResolution=node10, which some builds still rely on.
+    "ignoreDeprecations": "6.0",
```
srikrsna-buf, line 44: Typescript 7.0 will be the next release and it will remove this [flag](https://devblogs.microsoft.com/typescript/announcing-typescript-6-0/). We should fix this in version 6.0, so that we will be ready for 7.0 where potentially other issues might come up because of the native port. It will make it easier to report as it is supposed to work just like 6.0 without this flag.

packages/typescript-compat/v6.0.x/src/uint8array-arraybuffer.ts
srikrsna-buf, line 1: I don't think we need this file.

ajeetdsouza: Added the rest of the compat package, I had missed it earlier.

tsconfig.base.json
ajeetdsouza: Done - I moved our module resolution to `bundler`.

timostamm (approved):
Assuming this is equivalent to https://github.com/bufbuild/protobuf-es/pull/1398, let's get this merged.

PR #1697, opened 2026-05-11
Ignore typescript-compat in Dependabot

ajeetdsouza:


.github/dependabot.yaml
```diff
@@ -43,3 +43,23 @@ updates:
           - "@wdio/*"
           - "webdriver"
           - "webdriverio"
+  # The typescript-compat packages deliberately pin older versions of
+  # typescript and @types/node to test compatibility. Only allow patch
+  # updates for those two dependencies in these directories.
+  - package-ecosystem: "npm"
+    directories:
+      - "/packages/typescript-compat/*"
```
timostamm, line 51: I expect dependabot to look for a package.json and package-lock.json pair at the given directory. We're already setting this up on line 10, and this config will interleave.

Is this documented to work with a npm monorepo?

ajeetdsouza:
This doesn't seem possible with Dependabot.

PR #1717, opened 2026-06-10
Disallow propagation of protocol headers

### LINKED ISSUE #1582: ConnectError thrown through expressConnectMiddleware has invalid code&message
**Describe the bug**
Hello, I am using connect as a proxy to my server. I am experiencing that ConnectErrors thrown from server gets transformed into something else. Not sure if this is intentional or not.

For instance, I have this error on server: 

```js
// server throws this error:

ConnectError {
  name: 'ConnectError',
  rawMessage: 
    'Message validation failed: customer_account_id value does not match regex pattern `^([A-Z]{3}):([A-Za-z]+):([0-9ÆØÅæøåA-Za-z_\-]+)$`',
  code: 3,
  metadata: Headers {},
  details: [ ...],
  cause: undefined,
  message: 
    '[invalid_argument] Message validation failed: customer_account_id value does not match regex pattern `^([A-Z]{3}):([A-Za-z]+):([0-9ÆØÅæøåA-Za-z_\-]+)$`'
}

// client gets this error: 
ConnectError {
  name: 'ConnectError',
  rawMessage: 'HTTP 400',
  code: 13,
  metadata: Headers {},
  details: [],
  cause: undefined,
  message: '[internal] HTTP 400',
  line: 77,
  column: 10,
  sourceURL: 'http://localhost:3000/node_modules/.vite/deps/chunk-NKPO2XQB.js',
  stack:  ...
}

```

It seems like the error gets translated into a 400 bad request. I can find the error details in the `error.metadata`, but I find it strange that it gets transformed. Is it intentional? 

**To Reproduce**
Throw a ConnectError som server. Use expressConnectMiddleware. 

- @connectrpc/connect-express: "2.0.4",
- @connectrpc/connect-node: "2.0.4",
- Node.js version: v22.13.1

### LINKED ISSUE #1604: Forwarding upstream connectrpc client errors results in incorrect content-type in server response
**Describe the bug**

When an upstream grpc-go service is called (using connectrpc) within a connectrpc server implementation, and an error is returned, the content-type returned back to the browser is `application/grpc+proto`, which prevents the browser client from parsing the connectrpc error payload properly.

**To Reproduce**

Setup:

- upstream go-grpc service
- connectrpc nodejs service impl that calls upstream go-grpc service
  - served via express middleware

Bug:

1. When the upstream service call throws (not catching, letting connectrpc forward it to frontend)
2. connectrpc returns the error w/ content-type `application/grpc+proto` instead of `application/json`
3. client in browser parses this as `ConnectError: [unknown] HTTP 500` due to the content-type

If we create a _new_ ConnectError, the content-type is properly set. The issue seems to be rooted around reusing the connect error.

- @connectrpc/connect-web version: `2.1.0`
- @connectrpc/connect-node version: `2.1.0`
- Frontend framework and version: `18.3.1`
- Node.js version: `20.19.2`
- Browser and version: Chromium version `141.0.7390.123`

ajeetdsouza:
Fixes https://github.com/connectrpc/connect-es/issues/1582
Fixes https://github.com/connectrpc/connect-es/issues/1604

- Wire error metadata should not be propagated to the final ConnectError
- Protocol headers in general should never be propagated to the final ConnectError.


packages/connect/src/protocol/protocol-headers.ts
```diff
@@ -0,0 +1,122 @@
...
+ * Returns true if the given header name is a protocol header, i.e. it carries
+ * protocol-level information rather than user metadata. Protocol headers are
+ * reserved by the Connect, gRPC, and gRPC-Web protocol implementations.
+ *
+ * Mirrors isProtocolHeader in connect-go. Any changes to this method must be
+ * made there first.
+ *
+ * @private Internal code, does not follow semantic versioning.
+ */
+export function isProtocolHeader(name: string): boolean {
```
timostamm: Is this function only used in `mergeNonProtocolHeaders`? Then we don't export it (and don't need the `@private` tag because it's inaccessible anyways).

packages/connect/src/protocol/protocol-headers.ts
```diff
@@ -0,0 +1,122 @@
...
+/**
+ * The set of HTTP header name prefixes that carry protocol-level information.
+ *
+ * Mirrors protocolHeaderPrefixes in connect-go. Any changes to this constant must be
+ * made there first.
+ *
+ * @private Internal code, does not follow semantic versioning.
+ */
+const protocolHeaderPrefixes: ReadonlySet<string> = new Set(
+  [headerConnectUnaryTrailerPrefix].map((prefix) => prefix.toLowerCase()),
```
timostamm: Ah, I think there's a mix-up. Fields with the "Trailer-" prefix aren't carrying protocol-level information. It's quite the opposite, they carry user-provided fields. The prefix disambiguates trailer fields from true header fields on the wire.

packages/connect/src/protocol/protocol-headers.ts
ajeetdsouza: Removed the function.

packages/connect/src/protocol/protocol-headers.ts
ajeetdsouza: Fixed.

timostamm (approved):

PR #1721, opened 2026-06-12
Migrate dual-build packages to tshy

ajeetdsouza:
tshy eases a lot of the pain of building CJS + ESM modules. This PR migrates all dual-build packages to build with tshy.

### DIFF
--- a/packages/connect/package.json
+++ b/packages/connect/package.json
@@ -11,10 +11,7 @@
   "scripts": {
     "generate": "buf generate",
     "postgenerate": "license-header src/protocol-grpc/gen",
-    "prebuild": "rm -rf ./dist/*",
-    "build": "npm run build:cjs && npm run build:esm",
-    "build:cjs": "tsc --project tsconfig.json --module commonjs --verbatimModuleSyntax false --moduleResolution node10 --outDir ./dist/cjs && echo >./dist/cjs/package.json '{\"type\":\"commonjs\"}'",
-    "build:esm": "tsc --project tsconfig.json --outDir ./dist/esm",
+    "build": "tshy",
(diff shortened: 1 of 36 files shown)


packages/connect/package.json
```diff
@@ -24,44 +21,101 @@
...
+    "@bufbuild/buf": "^1.66.1",
+    "@bufbuild/protoc-gen-es": "^2.10.0",
+    "@types/jasmine": "^5.1.13",
+    "jasmine": "^5.13.0",
+    "tshy": "^4.1.3"
+  },
+  "files": [
+    "dist/**",
+    "!dist/**/*.spec.js",
+    "!dist/**/*.spec.d.ts"
```
timostamm, line 47: The package.json `files` property has bitten us before, see https://github.com/bufbuild/protobuf-es/pull/416

I'm okay with using it to exclude scripts, e.g. `scripts/update-user-agent.mjs` here, but I really want to avoid it becoming our go-to solution for excluding test files. I suggest to rely on tshy's `exclude` filter here.

timostamm (commented):
"Migrate connect to tshy" is a misleading PR title, see my comment on https://github.com/bufbuild/protobuf-es/pull/1435#pullrequestreview-4416936961.

I'm noticing that you update `tsconfig.json` to disable `verbatimModuleSyntax` in the package. Would it not make more sense to remove the config option from `tsconfig.base.json` instead? Otherwise we end up enabling it, only to disable it for every package individually.

Commented on "files" with some extra context. I think it may be easier to make this migration in a single PR. It's a bigger diff, but helps with consistency, and removes the changelog problem too.

ajeetdsouza:
Closed in favor of https://github.com/connectrpc/connect-es/pull/1744

PR #1730, opened 2026-06-24
Migrate jasmine to node:test

ajeetdsouza:
Jasmine is dated, increases our dependency surface, and does not run directly on source files, requiring `files` exclusions in `package.json`. This PR migrates all our pure-Jasmine tests to node:test.


packages/connect/src/protocol-connect/request-header.spec.ts
```diff
@@ -33,25 +35,27 @@ function listHeaderKeys(header: Headers): string[] {
...
       "user-agent",
     ]);
-    expect(headers.get("Content-Type")).toBe("application/proto");
-    expect(headers.get("Connect-Protocol-Version")).toBe("1");
-    expect(headers.get("User-Agent")).toMatch(/^connect-es\/\d+\.\d+\.\d+/);
+    assert.strictEqual(headers.get("Content-Type"), "application/proto");
+    assert.strictEqual(headers.get("Connect-Protocol-Version"), "1");
+    // The user-agent placeholder is replaced with the actual version at build
+    // time (see scripts/update-user-agent.mjs); these tests run against source.
+    assert.strictEqual(headers.get("User-Agent"), "CONNECT_ES_USER_AGENT");
```
ajeetdsouza, line 47: Flagging this: we now run tests on the source rather than on the built JS files, so the `postbuild` task hasn't yet replaced the user agent strings. This does make this test less useful, but I kept it in case we want to enforce that this string remains unchanged.

timostamm (approved):

Issue #1561, opened 2025-08-18
[connect-node] Transport does not recover from server restarts - connections remain in zombie state


ludusrusso:
**Describe the bug**

When using `@connectrpc/connect-node` with persistent gRPC connections, the transport fails to recover after the server restarts. The HTTP/2 connections remain in a "zombie" state, causing all subsequent requests to hang or fail, even though the server is back online and healthy.

Not sure if I've lost something or it is not expected behaviour.

**To Reproduce**

1. Create a gRPC transport with keep-alive configuration
2. Make successful requests to the server
3. Server restarts (e.g., deployment, update) during active call
4. Client connections become stale but are not detected as dead
5. All subsequent requests hang indefinitely or fail with timeout errors
6. The transport never recovers automatically, even with aggressive ping settings

NOTE: this happens only then a grpc call is in progress during restarts. I'm not able to reproduce it if there is not call coming.

**Environment (please complete the following information):**

- "@connectrpc/connect-node": "^2.0.4",
- Node.js version: 24.4

### Context

This happens inside a kubernetes cluster.

PR #1746, opened 2026-07-22
Fix Http2SessionManager.verify() never settling when the connection closes during verification

jackyzha0:
# Background

`verify()` settles its promise in two ways: the PING callback succeeds, or the connection emits `error`. There is a third case: the connection closes gracefully while the PING is in flight. This happens in practice when the idle timeout elapses during a verification - the idle timer keeps running while a verify is pending, and `onIdleTimeout` calls `conn.close()`.

A graceful close emits `close` without `error`. The PING callback receives `ERR_HTTP2_PING_CANCEL` and returns without resolving (deferring to the `error` listener that never fires), and the PING timeout timer that would have destroyed the connection is cleared by the ready state's `cleanup()` on close. So none of the settle paths run, and the promise stays pending forever.

Because the manager memoizes the promise in `this.verifying` and only clears it once it settles, a single occurrence poisons the manager permanently: every future request that hits `requiresVerify()` awaits the pinned promise until its own deadline. New connections still get dialed once the state reaches `closed`, but each one only serves the request that dialed it - after the next quiet period, requests funnel back into the stuck verification. `abort()` doesn't recover either, it never touches `verifying`.

We hit this in production: a process's transport stopped serving requests for ~50 minutes (every RPC timed out on its deadline, sockets healthy at the kernel level, no PING frames on the wire) until the process was restarted. Possibly the same root cause as the zombie state described in #1561.

# Summary

Resolve `verify()` on `close` as well, and detach the listeners on every settle path.

The new test reproduces the hang with a small TCP proxy that drops the answer to the verification PING while the idle timeout closes the connection gracefully. Without the fix, the request never settles; with it, the verification fails and the request proceeds on a new connection.

### DIFF
--- a/packages/connect-node/src/http2-session-manager.spec.ts
+++ b/packages/connect-node/src/http2-session-manager.spec.ts
@@ -274,21 +274,21 @@ describe("Http2SessionManager", () => {
     it("should open a new connection if the existing one closes gracefully during verification", async (t) => {
       // proxy the server, so that the answer to a verification PING can be dropped
       let dropServerToClient = false;
-      const proxy = net.createServer((clientSide) => {
-        const serverSide = net.connect(
+      const proxy = net.createServer((clientSocket) => {
+        const serverSocket = net.connect(
           Number(new URL(server.getUrl()).port),
           "localhost",
         );
-        clientSide.pipe(serverSide);
-        serverSide.on("data", (chunk) => {
+        clientSocket.pipe(serverSocket);
+        serverSocket.on("data", (chunk) => {
           if (!dropServerToClient) {
-            clientSide.write(chunk);
+            clientSocket.write(chunk);
           }
         });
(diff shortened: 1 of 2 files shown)


jackyzha0:
hello @ajeetdsouza is this project still active? we're currently using connect-es in production and wanted to contribute back upstream some fixes we've made locally via patching that have helped us

ajeetdsouza:
Hey, thanks for the PRs! Yes, this project is actively maintained, and we welcome contributions.

I'll take a look at your PRs soon. Thanks for your patience, and I appreciate you taking the time to contribute!

jackyzha0:
@ajeetdsouza sorry for the bump, wanted to ask again here! would love to get rid of our in-house patch if possible

ajeetdsouza (approved):

ajeetdsouza:
Merged, thank you for the fix!

PR #1774, opened 2026-08-25
Enable gRPC in Cloudflare conformance tests

mar-cf:
Cloudflare Workers can't send gRPC directly so the client sends gRPC-Web and opts into Cloudflare's conversion via `cf: { grpcWeb: "convert" }` request property or the `auto_grpc_convert` compatibility flag

If not already, you may need to enable gRPC on the zone serving `CLOUDFLARE_WORKERS_SERVER_HOST`

### DIFF
--- a/packages/connect-cloudflare/conformance/conformance-cloudflare-server.yaml
+++ b/packages/connect-cloudflare/conformance/conformance-cloudflare-server.yaml
@@ -4,7 +4,7 @@ features:
     - HTTP_VERSION_2
   protocols:
     - PROTOCOL_CONNECT
-    # - PROTOCOL_GRPC
+    - PROTOCOL_GRPC
     - PROTOCOL_GRPC_WEB
   codecs:
     - CODEC_PROTO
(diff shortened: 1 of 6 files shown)


packages/connect-cloudflare/conformance/conformance-cloudflare-server.yaml
```diff
@@ -24,3 +24,7 @@ excludeCases:
     streamType: STREAM_TYPE_FULL_DUPLEX_BIDI_STREAM
   - protocol: PROTOCOL_GRPC_WEB
     streamType: STREAM_TYPE_FULL_DUPLEX_BIDI_STREAM
+  - protocol: PROTOCOL_GRPC
```
ajeetdsouza, line 28: Why are we disabling cases with compression? Might be good to document the reason in a comment.

mar-cf: Cloudflare's inbound gRPC conversion does not currently handle compressed messages correctly, so those conformance cases are excluded while we investigate a fix.

Added a comment.

ajeetdsouza (approved):

packages/connect-cloudflare/conformance/known-failing-client.txt
```diff
@@ -1,2 +1,3 @@
 # On receiving 504, cloudflare is returning a static response back instead of the actual body.
 Errors/*/Protocol:PROTOCOL_CONNECT/*/*/TLS:true/unary/deadline-exceeded
+gRPC Unexpected Responses/HTTPVersion:2/TLS:true/trailers-only/ignore-header-if-body-present
```
ajeetdsouza, line 3: This appears to be an edge case of the gRPC spec that should be addressed separately.

PR #1777, opened 2026-08-26
Drop support for Node.js 20

ajeetdsouza:
Removes support for Node 20, as per our [compatibility policy](https://github.com/connectrpc/connect-es#compatibility). A bunch of our dependencies also no longer support Node 20, as it is EOL.

This PR also adds Node 26 to CI.


timostamm (approved):
Please update the PR title to be more clear on what this is, see https://github.com/connectrpc/connect-es/pull/1559 for example.

PR #1782, opened 2026-08-28
Simplify tests with type-narrowing assertions

ajeetdsouza:
`assert.notStrictEquals` does not do any type narrowing, so we had to add if-conditions to make our tests type-check. This should simplify the tests - the decision tree I'm proposing is:
- use `assert.strictEqual()` for all equality checks.
- use`assert.notStrictEqual()` for all inequality checks, unless type narrowing is required, in which case, use `assert.ok()`.
- use `assert.ok()` for asserting booleans.

### DIFF
--- a/packages/connect-migrate/src/lib/bufgenyaml.spec.ts
+++ b/packages/connect-migrate/src/lib/bufgenyaml.spec.ts
@@ -54,12 +54,12 @@ plugins:
 describe("parseBufGenYaml()", () => {
   it("should parse golden v1", () => {
     const y: BufGenYaml = parseBufGenYaml(goldenV1);
-    assert.notStrictEqual(y, undefined);
+    assert.ok(y !== undefined);
     assert.strictEqual(y.get("version"), "v1");
   });
   it("should parse golden v2", () => {
     const y: BufGenYaml = parseBufGenYaml(goldenV2);
-    assert.notStrictEqual(y, undefined);
+    assert.ok(y !== undefined);
     assert.strictEqual(y.get("version"), "v2");
   });
   it("raises error for unknown version", () => {
--- a/packages/connect-migrate/src/lib/bufgenyaml.spec.ts
+++ b/packages/connect-migrate/src/lib/bufgenyaml.spec.ts
@@ -54,12 +54,12 @@ plugins:
 describe("parseBufGenYaml()", () => {
   it("should parse golden v1", () => {
     const y: BufGenYaml = parseBufGenYaml(goldenV1);
-    assert.ok(y !== undefined);
+    assert.notStrictEqual(y, undefined);
     assert.strictEqual(y.get("version"), "v1");
   });
   it("should parse golden v2", () => {
     const y: BufGenYaml = parseBufGenYaml(goldenV2);
-    assert.ok(y !== undefined);
+    assert.notStrictEqual(y, undefined);
     assert.strictEqual(y.get("version"), "v2");
   });
   it("raises error for unknown version", () => {
--- a/packages/connect-node/src/node-universal-handler.spec.ts
+++ b/packages/connect-node/src/node-universal-handler.spec.ts
@@ -70,8 +70,8 @@ describe("universalRequestFromNodeResponse()", () => {
     it("should abort request signal with ConnectError and Code.Canceled for NO_ERROR", async () => {
       await request(http2.constants.NGHTTP2_NO_ERROR);
       assert.ok(serverRequest?.signal instanceof AbortSignal);
-      assert.strictEqual(serverRequest?.signal.aborted, true);
-      assert.ok(serverRequest?.signal.reason instanceof Error);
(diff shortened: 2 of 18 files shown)


packages/connect-node/src/node-universal-handler.spec.ts
```diff
@@ -72,86 +72,73 @@ describe("universalRequestFromNodeResponse()", () => {
...
       assert.ok(serverRequest?.signal.reason instanceof Error);
-      if (serverRequest?.signal.reason instanceof Error) {
-        assert.strictEqual(serverRequest.signal.reason.name, "AbortError");
-        assert.strictEqual(
-          serverRequest.signal.reason.message,
-          "This operation was aborted",
-        );
-      }
+      assert.strictEqual(serverRequest.signal.reason.name, "AbortError");
+      assert.strictEqual(
```
timostamm, line 76: Looks like a win in readability to me. 👍

packages/connect-node/src/node-universal-handler.spec.ts
```diff
@@ -72,86 +72,73 @@ describe("universalRequestFromNodeResponse()", () => {
...
       await request(http2.constants.NGHTTP2_NO_ERROR);
-      assert.notStrictEqual(serverRequest, undefined);
-      if (serverRequest !== undefined) {
-        assertByteStreamRequest(serverRequest);
-        const iterator = serverRequest.body[Symbol.asyncIterator]();
-        const r = await iterator.next();
-        assert.ok(r.done);
-      }
+      assert.ok(serverRequest !== undefined);
+      assertByteStreamRequest(serverRequest);
```
timostamm, line 84: Also looks like a win in readability to me - and worth switching to assert.ok to drop the if statement  👍

packages/connect-node/src/node-universal-handler.spec.ts
```diff
@@ -72,86 +72,73 @@ describe("universalRequestFromNodeResponse()", () => {
...
     });
     it("should abort request signal with ConnectError and Code.Canceled for CANCEL", async () => {
       await request(http2.constants.NGHTTP2_CANCEL);
       assert.ok(serverRequest?.signal instanceof AbortSignal);
       assert.strictEqual(serverRequest?.signal.aborted, true);
       assert.ok(serverRequest?.signal.reason instanceof ConnectError);
-      const ce = ConnectError.from(serverRequest?.signal.reason);
       assert.strictEqual(
-        ce.message,
+        serverRequest.signal.reason.message,
```
timostamm, line 95: Also a win, `ConnectError.from` didn't add anything.

packages/connect/src/protocol-connect/handler-factory.spec.ts
```diff
@@ -532,7 +515,7 @@ describe("createHandlerFactory()", () => {
         });
         ac.abort("test-reason");
         await resPromise;
-        assert.notStrictEqual(handlerContextSignal, undefined);
+        assert.ok(handlerContextSignal !== undefined);
```
timostamm, line 518: This change allows the lines below to drop the optional chaining. Why don't they? Same on line 556. Maybe elsewhere?

packages/connect-migrate/src/lib/bufgenyaml.spec.ts
```diff
@@ -54,12 +54,12 @@ plugins:
 describe("parseBufGenYaml()", () => {
   it("should parse golden v1", () => {
     const y: BufGenYaml = parseBufGenYaml(goldenV1);
-    assert.notStrictEqual(y, undefined);
+    assert.ok(y !== undefined);
```
timostamm: The downside of assert.ok is that its error message isn't as descriptive as notStrictEqual. We're not gaining anything from the narrowed type here, so this is not a clear win from my POV.

timostamm (commented):
Thanks for looking into this. The tests were ported from Jasmine, which never did any narrowing. It's unfortunate that the negations can't narrow, but at least we can drop a large number of conditional tests.

> `assert.notStrictEquals` does not do any type narrowing, so we had to add if-conditions to make our tests type-check. This should simplify the tests - the decision tree I'm proposing is:
> 
> - `x === y`: freely use `strictEqual` / `deepStrictEqual` (performs narrowing)
> - `x instanceof y`, `x !== undefined` or `x !== null`: use `ok(...)` (performs narrowing)
> - `x !== y`: use `notStrictEqual` (we don't need narrowing)
> - checking a boolean `x`: `ok(x)`

Isn't the second bullet a bit to narrow (pardon the pun)? I mean that it's also possible to narrow a string literal union, for example.

One thing to weigh: assert.ok can't show expectation and actual value in the test output. So from my point of view, notStrictEqual and notDeepStrictEqual still provide value, even though I'd be comfortable to replace with assert.ok where it _actually_ simplifies the tests. WDYT?

timostamm (approved):
Great cleanup, thanks!

PR #1790, opened 2026-09-01
Upgrade to Biome v2

ajeetdsouza:
This PR gets rid of all the package-specific configs we had set up for formatting, in favor of a unified config file spanning the entire workspace. This is also the approach favored by Biome v2.

This removes package-level `lint` and `format` tasks in favor of single root tasks. This is much simpler in terms of configuration as well as workflow. If a package-level task is needed in the future, `biome format` and `biome lint` can be run on individual subdirectories.

I checked our other repositories too, there are no task dependencies which rely on a package-level `lint` / `format`.


timostamm:
As discussed in person, biome's favored approach doesn't necessarily match turborepo's favored approach, so the decisions to make isn't clearly cut.

Getting rid of boilerplate config is obviously a win, but no longer being able to use `turbo` as the single interface for build tasks including package scoping is a loss.

It's a trade-off, and it would be very useful to be able to understand the trade-off and rationale from the PR. If the decision is to keep all formatting and linting in root tasks, please make sure to update CONTRIBUTING.md to reflect that.

Please check how this approach translates to other repositories using biome. For example, if a package level task depends on `lint` or - more likely - `format`, I'm not sure running a root level `format` instead is a fix.

ajeetdsouza:
Updated the description.

PR #1792, opened 2026-09-02
Stop delivering stream messages after cancellation

ajeetdsouza:
Clients read bytes from the server in _chunks_. If the chunk happens to contain more than one envelope, the envelope reader queues them all up. This means that if a client reads a single message and aborts, the remaining queued messages from that chunk would keep getting delivered.

This change makes `runStreamingCall` check the abort signal before delivering each message. This fixes the client behavior for bidi / server-side streaming, in both `connect-web` and `connect-node`. Servers are unaffected.

Fixes the conformance flake found in: https://github.com/connectrpc/connect-es/actions/runs/33062690944/job/98485105428?pr=1780

### DIFF
--- a/packages/connect-web-bench/README.md
+++ b/packages/connect-web-bench/README.md
@@ -15,10 +15,10 @@ usually do. We repeat this for an increasing number of RPCs.
 
 | code generator | RPCs | bundle size |  minified | compressed |
 | -------------- | ---: | ----------: | --------: | ---------: |
-| Connect-ES     |    1 |   285,189 b | 180,619 b |   36,513 b |
-| Connect-ES     |    4 |   289,441 b | 183,721 b |   37,280 b |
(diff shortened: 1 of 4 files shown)


timostamm (approved):
Good catch!

That's a user-facing fix, right? Consider:

```ts
const abortController = new AbortController();
const requestMessage = {}
const callOptions = {
  signal: abortController.signal,
}
for await (const responseMessage of client.serverStreaming(requestMessage, callOptions)) {
  abortController.abort();
}
```

The body of the loop aborts at the first response message, but the body can actually run multiple times, depending on how the response is chunked on the wire.

With this fix, the body runs exactly once before the loop throws.

That seems worthy to add to the PR description, along with some details on what's affected - e.g. clients or servers as well? Only clients from `@connectrpc/connect-web`, or also `@connectrpc/connect-node`? Only server-streaming, or also bidi-streaming RPCs?

PR #1798, opened 2026-09-15
Remove "stay tuned for Connect for Node" from error message

haines:
There is an assertion in `connect-web` to check for the Fetch API. All supported versions of Node.js (22+) have the Fetch API so `connect-web` should work fine, so this PR drops the part of the message that states that Node.js is not supported, and also the part that says to "stay tuned for Connect for Node" (the error predates the existence of `connect-node`).

To be honest, I reckon it would be fine to remove the assertion altogether now, given the wide support for the Fetch API. What do you think?


timostamm (approved):
Yes, all maintained Node.js versions ship the API today, but I think it's still possible to end up without it, for example in Jest. Needs a closer look before we remove it. 

Thanks for fixing the outdated message!

